PR #19, opened 2021-01-28
VHensorskyi/authorization

Hensorskyy:


Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+WORKDIR /OoS-Backend
+RUN dotnet build -c $Configuration -o /app
+
+FROM builder AS publish
+ARG Configuration=Release
+RUN dotnet publish -c $Configuration -o /app
+
+FROM base AS final
+WORKDIR /app
```
DmyMi, line 36: `workdir app` is already specified in the base image. Don't need to add another layer here

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+COPY ./OutOfSchool/*.sln ./
+COPY ./OutOfSchool/OutOfSchool.Tests/*.csproj ./OutOfSchool.Tests/
+COPY ./OutOfSchool/OutOfSchool.WebApi/*.csproj ./OutOfSchool.WebApi/
+COPY ./OutOfSchool/IdentityServer/*.csproj ./IdentityServer/
+
+RUN dotnet restore
+
+COPY ./OutOfSchool/ ./
+
+WORKDIR /OoS-Backend
```
DmyMi, line 28: `WORKDIR /OoS-Backend` is already set on line 18, don't need to repeat

DmyMi (changes requested):

DmyMi (approved):

PR #26, opened 2021-02-04
Hotfix

Bogdan-Hasanov:


OutOfSchool/OutOfSchool.WebApi/Controllers/OrganizationController.cs
```diff
@@ -10,15 +13,20 @@ namespace OutOfSchool.WebApi.Controllers
     public class OrganizationController : ControllerBase
     {
         private readonly ILogger<OrganizationController> _logger;
+        private readonly OutOfSchoolDbContext _context;
 
-        public OrganizationController(ILogger<OrganizationController> logger)
+        public OrganizationController(ILogger<OrganizationController> logger, OutOfSchoolDbContext context)
```
dmitrykiev, line 18: shouldn't interface be passed here instead of concrete class?

dmitrykiev:
There are failing tests.

PR #48, opened 2021-02-26
Dzhus/tests for organization controller and service

VyacheslavDzhus:


OutOfSchool/OutOfSchool.DataAccess/Models/Organization.cs
```diff
@@ -40,12 +40,11 @@ public class Organization
             ErrorMessage = "INPP code should contain 12, 10 or 9 digits")]
         public string INPP { get; set; }
         
-        [DataType(DataType.ImageUrl)]
-        public byte[]? Image { get; set; }
+        //[DataType(DataType.ImageUrl)]
+        //public byte[]? Image { get; set; }
```
dmitrykiev, line 44: is this code needed?

OutOfSchool/OutOfSchool.DataAccess/Repository/EntityRepository.cs
```diff
@@ -37,10 +37,11 @@ public async Task<T> Create(T entity)
         /// <inheritdoc/>
         public async Task Delete(T entity)
         {
-            this.dbSet.Remove(entity);
+           
+            this.dbSet.Remove(entity);
```
dmitrykiev, line 41: are these changes relevant?

OutOfSchool/OutOfSchool.Tests/OrganizationControllerTest.cs
```diff
@@ -1,33 +1,280 @@
...
+                },
+            }.ToList();
+
+            organizationService.Setup( it => it.GetAll()).Returns(Task.FromResult(organizations));
+
+            // Act
+            var response = await organizationController.GetOrganizations();
+
+            // Assert
+            Assert.IsInstanceOf<OkObjectResult>(response);
```
dmitrykiev, line 63: it's nice to test not only the response type but content as well

OutOfSchool/OutOfSchool.Tests/OrganizationControllerTest.cs
```diff
@@ -1,33 +1,280 @@
...
+            var response = await organizationController.GetOrganizationById(organization.Id);
+
+            // Assert
+            Assert.IsInstanceOf<OkObjectResult>(response);
         }
-        [SetUp]
-        public void Setup()
+
+        [Test]
+        public async Task Get_OrganizationById_ReturnsBadRequestObjectResult()
```
dmitrykiev, line 108: it would be nice to understand why this test returns bad request
for example from the test name like GetOrganizationById_NonExistentId_ReturnsBadRequestObjectResult

OutOfSchool/OutOfSchool.Tests/OrganizationControllerTest.cs
VyacheslavDzhus: You said that we have to test only status codes in controller, didn't you ?

OutOfSchool/OutOfSchool.Tests/OrganizationControllerTest.cs
```diff
@@ -1,33 +1,280 @@
...
+            // Assert
+            Assert.IsInstanceOf<BadRequestObjectResult>(response);
+        }
+        [Test]
+        public async Task Update_OrganizationThatNull_ReturnsBadRequestObjectResult()
+        {
+            // Arrange
+            OrganizationDTO organization = null;
+            
+            organizationService.Setup(it => it.Update(organization)).Returns(Task.FromResult(organization));
```
dmitrykiev, line 185: no need to setup service in this test

OutOfSchool/OutOfSchool.Tests/OrganizationControllerTest.cs
```diff
@@ -1,33 +1,280 @@
...
+
+            // Assert
+            Assert.IsInstanceOf<OkObjectResult>(result);
         }
+
+        [Test]
+        public async Task Update_OrganizationNotValidModel_ReturnsBadRequestObjectResult()
+        {
+            // Arrange
+            var organization = new OrganizationDTO
```
dmitrykiev, line 154: this can be declared as class field

OutOfSchool/OutOfSchool.Tests/OrganizationServiceTest.cs
```diff
@@ -0,0 +1,336 @@
...
+                UserId = "123"
+            };
+
+            // Act and Assert
+            Assert.ThrowsAsync<ArgumentException>(
+                   async () => await _organizationService.Update(organization).ConfigureAwait(false));
+        }
+
+        [Test, Order(14)]
+        public async Task Delete_Organization_DeletedFromDatabase()
```
dmitrykiev, line 312: please add a test for the case when id is not found in the db

OutOfSchool/OutOfSchool.Tests/OrganizationControllerTest.cs
```diff
@@ -1,33 +1,280 @@
...
+            // Act
+            var responce = await organizationController.Delete(organizationId);
+
+            // Assert
+            Assert.IsInstanceOf<BadRequestObjectResult>(responce);
+        }
+
+
+        [Test]
+        public async Task Create_Organization_ReturnsCreatedAtActionResult()
```
dmitrykiev, line 254: please add a test for unsuccessful case

OutOfSchool/OutOfSchool.Tests/OrganizationServiceTest.cs
```diff
@@ -0,0 +1,336 @@
...
+            // Act and Assert
+            Assert.ThrowsAsync<ArgumentException>(
+                async () => await _organizationService.Create(newOrganization).ConfigureAwait(false));           
+        }
+
+        [Test, Order(5)]
+        public async Task GetAll_Organizations_ReturnsSameAmountOrganizations()
+        {
+            // Arrange
+            var allOrganizations = await _organizationService.GetAll();
```
dmitrykiev, line 117: service method is called, so this looks more like the Act step, not the Arrange step

OutOfSchool/OutOfSchool.Tests/OrganizationControllerTest.cs
```diff
@@ -1,33 +1,280 @@
...
+
+            // Act
+            var responce = await organizationController.Update(organization);
+
+            // Assert
+            Assert.IsInstanceOf<BadRequestObjectResult>(responce);
+        }
+
+        [Test]
+        public async Task Update_OrganizationExceptionOccurred_ReturnsBadRequestObjectResult()
```
dmitrykiev, line 195: in this case the request is valid but user will get BadRequestObjectResult
this might be confusing

consider making changes to return result that describes what happened

OutOfSchool/OutOfSchool.Tests/OrganizationControllerTest.cs
dmitrykiev: you are right, status code is enough here

PR #65, opened 2021-03-09
Ostapchuk/parent controller

YehorOstapchuk:


OutOfSchool/OutOfSchool.DataAccess/Models/Parent.cs
```diff
@@ -6,21 +6,31 @@ namespace OutOfSchool.Services.Models
     public class Parent
     {
         public long Id { get; set; }
-        
-        [DataType(DataType.Text)]
+
         [Required(ErrorMessage = "First name is required")]
+        [DataType(DataType.Text)]
+        [MaxLength(30)]
+        [RegularExpression(@"^([^0-9]*)$", ErrorMessage = "First name cannot contains digits")]
```
mmmpolishchuk, line 13: Personal info cannot contains digits, as well as special symbols. 
Try this pattern: ^[a-zA-Z]*$

dmitrykiev: but it can contain hyphens, for example "Дмитрий Мамин-Сибиряк"

PR #67, opened 2021-03-10
exceptions handling & logging

mmmpolishchuk:


OutOfSchool/OutOfSchool.Tests/Services/WorkshopServiceTests.cs
```diff
@@ -0,0 +1,317 @@
...
+using OutOfSchool.WebApi.Models;
+using OutOfSchool.WebApi.Models.ResultModel;
+using OutOfSchool.WebApi.Services;
+
+namespace OutOfSchool.Tests.Services
+{
+    [TestFixture]
+    public class WorkshopServiceTests
+    {
+        // private OutOfSchoolDbContext context;
```
dmitrykiev, line 21: please remove commented out code

OutOfSchool/OutOfSchool.WebApi/Controllers/ChildController.cs
```diff
@@ -31,8 +34,9 @@ public ChildController(IChildService childService)
         /// Get all children from the database.
         /// </summary>
         /// <returns>List of all children.</returns>
+        [ProducesResponseType(StatusCodes.Status200OK)]
         [HttpGet]
-        public async Task<ActionResult<IEnumerable<Child>>> GetChildren()
+        public async Task<ActionResult> GetChildren()
```
dmitrykiev, line 39: please make in consistent, return IActionResult

OutOfSchool/OutOfSchool.WebApi/Controllers/OrganizationController.cs
```diff
@@ -33,8 +35,9 @@ public OrganizationController(ILogger<OrganizationController> logger, IOrganizat
         /// Get all organization from the database.
         /// </summary>
         /// <returns>List of all organizations.</returns>
+        [ProducesResponseType(StatusCodes.Status200OK)]
```
dmitrykiev, line 38: could it produce responses other than 200?

OutOfSchool/OutOfSchool.WebApi/Controllers/TeacherController.cs
```diff
@@ -101,12 +113,15 @@ public async Task<ActionResult> Update(TeacherDTO teacherDto)
...
         /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
         [Authorize(Roles = "organization,admin")]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
         [HttpDelete("{id}")]
-        public async Task<ActionResult> Delete(long id)
+        public async Task<IActionResult> DeleteTeacher(long id)
         {
-            if (id == 0)
+            if (id < 1 || teacherService.GetAll().Result.AsQueryable().Count() < id)
```
dmitrykiev, line 121: teacherService.GetAll()
we retrieve all data from the db just to check the count
this is very unproductive, please change

OutOfSchool/OutOfSchool.WebApi/Controllers/WorkshopController.cs
```diff
@@ -99,14 +108,18 @@ public async Task<ActionResult> Update(WorkshopDTO workshopDto)
...
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
         [HttpDelete("{id}")]
-        public async Task<ActionResult> Delete(long id)
+        public async Task<IActionResult> Delete(long id)
         {
-            if (id == 0)
+            if (id < 1)
             {
-                return BadRequest("Id cannot be 0.");
+                throw new ArgumentOutOfRangeException(
```
dmitrykiev, line 118: we need to pass parameter name to the ArgumentOutOfRangeException constructor
id.ToString() is not the parameter name

OutOfSchool/OutOfSchool.WebApi/Extensions/ExceptionMiddlewareExtension.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        private Task HandleExceptionAsync(HttpContext context, Exception exception)
+        {
+            context.Response.ContentType = "application/json";
+            context.Response.StatusCode = 500;
+
+            return context.Response.WriteAsync(JsonSerializer.Serialize(new ErrorDetails()
+            {
+                StatusCode = context.Response.StatusCode,
```
dmitrykiev, line 50: do we need to specify StatusCode here? we have specified it for context.Response.StatusCode already

OutOfSchool/OutOfSchool.WebApi/Extensions/MappingExtensions.cs
```diff
@@ -64,23 +60,30 @@ public static CategoryDTO ToModel(this Category category)
         public static Workshop ToDomain(this WorkshopDTO workshopDto)
         {
             var workshop =
```
dmitrykiev, line 62: no need to introduce workshop variable
we can just write 'return'

OutOfSchool/OutOfSchool.WebApi/Program.cs
```diff
@@ -1,24 +1,52 @@
...
+namespace OutOfSchool.WebApi
 {
     public class Program
     {
         public static void Main(string[] args)
         {
-            CreateHostBuilder(args).Build().Run();
+            // Read Configuration from appSettings
+            var config = new ConfigurationBuilder()
+                .AddJsonFile("appsettings.Development.json")
```
dmitrykiev, line 16: we should not hardcode development config here

OutOfSchool/OutOfSchool.WebApi/Program.cs
```diff
@@ -1,24 +1,52 @@
...
+            finally
+            {
+                Log.CloseAndFlush();
+            }
         }
-
+        
         public static IHostBuilder CreateHostBuilder(string[] args) =>
             Host.CreateDefaultBuilder(args)
+                .UseSerilog() // Uses Serilog instead of default .NET Logger
```
dmitrykiev, line 49: this comment is quite obvious
we usually comment other cases, complex logic etc.

OutOfSchool/OutOfSchool.WebApi/Services/ChildService.cs
```diff
@@ -15,124 +18,101 @@ namespace OutOfSchool.WebApi.Services
...
         {
+            logger.Information("Process of getting Child by id started.");
+
             var child = await repository.GetById(id).ConfigureAwait(false);
 
             if (child == null)
             {
-                throw new ArgumentException("Incorrect Id!", nameof(id));
+                throw new ArgumentOutOfRangeException(id.ToString(),
+                    "The id cannot be greater than number of table entities.");
```
dmitrykiev, line 71: maybe child is null because of some other reason, the message "The id cannot be greater than number of table entities." might be confusing

OutOfSchool/OutOfSchool.WebApi/Services/ChildService.cs
```diff
@@ -15,124 +18,101 @@ namespace OutOfSchool.WebApi.Services
...
-            }
-            catch (Exception ex)
-            {
-                throw new Exception($"{nameof(dto)} could not be saved: {ex.Message}");
-            }
+            return newChild.ToModel();
         }
 
         /// <inheritdoc/>
         public async Task<IEnumerable<ChildDTO>> GetAll()
```
dmitrykiev, line 48: probably we also need Get methods with some filters like ParentId or SectionId

OutOfSchool/OutOfSchool.WebApi/Services/OrganizationService.cs
```diff
@@ -15,30 +16,29 @@ namespace OutOfSchool.WebApi.Services
...
-                throw new ArgumentNullException(nameof(dto), "Organization was null.");
-            }
+            logger.Information("Organization creating was started.");
 
             var organization = dto.ToDomain();
 
             if (!repository.IsUnique(organization))
             {
-                throw new ArgumentException(nameof(organization), "There is already an organizationDto with such data");
+                throw new ArgumentException(nameof(organization), "There is already an organization with a such data");
```
dmitrykiev, line 41: it was correct - "with such data"

OutOfSchool/OutOfSchool.WebApi/Services/OrganizationService.cs
```diff
@@ -49,80 +49,68 @@ public async Task<OrganizationDTO> Create(OrganizationDTO dto)
...
         public async Task Delete(long id)
         {
+            var dtoToDelete = new Organization() { Id = id };
+
             try
             {
-                await repository
-                    .Delete(await repository.GetById(id).ConfigureAwait(false))
-                    .ConfigureAwait(false);
+                await repository.Delete(dtoToDelete).ConfigureAwait(false);
```
dmitrykiev, line 106: small note
dtoToDelete could be named entity or organization
code will read better then: repository.Delete(entity)

OutOfSchool/OutOfSchool.WebApi/Services/OrganizationService.cs
```diff
@@ -49,80 +49,68 @@ public async Task<OrganizationDTO> Create(OrganizationDTO dto)
...
+
+                logger.Information("Deleting successfully finished.");
```
dmitrykiev, line 108: small note
could be 'Organization successfully deleted.'

OutOfSchool/OutOfSchool.WebApi/Controllers/WorkshopController.cs
mmmpolishchuk: image
yes, it should be the parameter name but casted to the string

OutOfSchool/OutOfSchool.WebApi/Services/ChildService.cs
mmmpolishchuk: I cannot find another reason that could cause an exception here. 
The first check is located in the controller - it checks param if it is not less than 1. 
Then another one is in the service - if there is no entity found by this id, for me it just means that id is greater than entities count, and we get null. 
Can’t be null in the database?

dmitrykiev: if we retrieve a child that was previously deleted, we'll get null but error message won't be relevant

OutOfSchool/OutOfSchool.WebApi/Controllers/OrganizationController.cs
mmmpolishchuk: Do you mean 204 No Content?

dmitrykiev: could it produce 400 or 500 response?

OutOfSchool/OutOfSchool.WebApi/Extensions/ExceptionMiddlewareExtension.cs
mmmpolishchuk: Firstly, in the line 46, I manually assigned a 500 status code to the context.Response.StatusCode. Then it needs to be returned it as Json.
We can put only an object type inside the Serialize method (in our case it's ErrorDetails. This class is like a plug. It has no meaning at all, only returns exception details).  And during creating new ErrorDetails, I just assigned a value from line 46 to the Status code. 
In my opinion, it's a good approach. 
If you have suggestion to do it in better way and I missed something - please write.

OutOfSchool/OutOfSchool.WebApi/Services/ChildService.cs
mmmpolishchuk: Do you mean to get a list of children of a specific workshop, by workshop's id? And the same for parent - get all his/her children?

dmitrykiev: yes

PR #74, opened 2021-03-18
Dzhus/address controller

VyacheslavDzhus:


OutOfSchool/OutOfSchool.WebApi/Models/AddressDto.cs
```diff
@@ -21,7 +21,7 @@ public class AddressDTO
 
         [Required(ErrorMessage = "Building number is required")]
         [MaxLength(15)]
-        public string Building { get; set; }
+        public string BuildingNumb { get; set; }
```
dmitrykiev, line 24: BuildingNumb -> BuildingNumber

OutOfSchool/OutOfSchool.WebApi/Controllers/AddressController.cs
```diff
@@ -0,0 +1,142 @@
...
+            }
+
+            try
+            {
+                AddressDto address = await addressService.Create(addressDto).ConfigureAwait(false);
+                return CreatedAtAction(
+                    nameof(GetAddressById),
+                    new
+                    {
+                        id = address.Id,
```
dmitrykiev, line 91: why do we pass address.Id and address? 
address contains address.Id already

OutOfSchool/OutOfSchool.WebApi/Controllers/AddressController.cs
```diff
@@ -0,0 +1,142 @@
...
+                    }, address);
+            }
+            catch (ArgumentException ex)
+            {
+                return BadRequest(ex.Message);
```
dmitrykiev, line 96: don't forget to add logging and save full exception details

OutOfSchool/OutOfSchool.WebApi/Services/AddressService.cs
```diff
@@ -0,0 +1,100 @@
...
+        public Task<AddressDto> Create(AddressDto dto)
+        {
+            if (dto == null)
+            {
+                throw new ArgumentNullException(nameof(dto), "Address is null.");
+            }
+
+            var address = dto.ToDomain();
+
+            return CreateInternal(address);
```
dmitrykiev, line 39: wouldn't it be shorter 
return CreateInternal(dto.ToDomain());
?

OutOfSchool/OutOfSchool.WebApi/Services/AddressService.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        /// <inheritdoc/>
+        public async Task<AddressDto> GetById(long id)
+        {
+            var address = await repository.GetById(id).ConfigureAwait(false);
+
+            if (address == null)
+            {
+                throw new ArgumentException("Incorrect Id!", nameof(id));
```
dmitrykiev, line 57: we don't usually put exclamation marks in error messages

OutOfSchool/OutOfSchool.WebApi/Services/AddressService.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <inheritdoc/>
+        public async Task<AddressDto> Update(AddressDto dto)
+        {
+            try
+            {
+                var address = await repository.Update(dto.ToDomain()).ConfigureAwait(false);
+
+                return address.ToModel();
+            }
+            catch (DbUpdateConcurrencyException)
```
dmitrykiev, line 72: what is this for?

OutOfSchool/OutOfSchool.WebApi/Controllers/AddressController.cs
mmmpolishchuk: > why do we pass address.Id and address?
> address contains address.Id already

otherwise, if we pass only address.id, it will not be working. I have tested it. Both id and address should be passed.

YehorOstapchuk (approved):

dmitrykiev (approved):

PR #77, opened 2021-03-18
Ostapchuk/parent controller

YehorOstapchuk:


OutOfSchool/OutOfSchool.WebApi/Extensions/MappingExtensions.cs
```diff
@@ -45,6 +45,13 @@ public static CategoryDTO ToModel(this Category category)
             return categoryDto;
         }
 
+        public static ParentDTO ToModel(this Parent parent)
+        {
+            var parentDto =
+                Mapper<Parent, ParentDTO>(parent, cfg => { cfg.CreateMap<Parent, ParentDTO>(); });
+            return parentDto;
```
mmmpolishchuk, line 52: can be just `return Mapper<Parent, ParentDTO>(parent, cfg => { cfg.CreateMap<Parent, ParentDTO>(); });`

OutOfSchool/OutOfSchool.WebApi/Services/ParentService.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        /// <inheritdoc/>
+        public async Task<ParentDTO> Update(ParentDTO parent)
+        {
+            await UpdateValidation(parent).ConfigureAwait(false);
+            Parent res = await repository.Update(parent.ToDomain()).ConfigureAwait(false);
+            return res.ToModel();
+        }
+
+        public static void CreateValidation(ParentDTO parent)
```
mmmpolishchuk, line 75: do we need this validation? Regex are responsible for these validations. Even data isn't valid, it will not get to this method.

mmmpolishchuk (commented):
Please, could you reformat code? In order to remove trailing spaces etc. 
Ctrl + K + D command.

PR #83, opened 2021-03-21
Feature/#76 new registration layout

Art-mimiron:


.idea/OoS-Backend.iml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dmitryy90, line 1: i believe settings files should be added to git ignore

dmitryy90 (approved):

.idea/OoS-Backend.iml
Art-mimiron: file was removed

DmyMi (approved):

PR #85, opened 2021-03-22
Ivanytskay/ChildService&ChildController tests

Elizabeth129:


DmyMi (approved):

OutOfSchool/OutOfSchool.DataAccess/Models/BirthCertificate.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class BirthCertificate
+    {
+        [Key]
+        [ForeignKey("Child")]
+        public long Id { get; set; }
+
+        /// <summary>
+        /// Gets or sets series of birth certificate.
+        /// </summary>
+        public string SvidSer { get; set; }
```
dmitrykiev, line 18: Is it better to name it 'Series'?

then when you access it it will look like certificate.Series

OutOfSchool/OutOfSchool.DataAccess/Models/Child.cs
```diff
@@ -9,27 +9,35 @@ public class Child
         public long Id { get; set; }
 
         [DataType(DataType.Text)]
-        [Required(ErrorMessage = "First name is required")]
+        [MaxLength(40)]
+        [RegularExpression(@"[\W\-\']*", ErrorMessage = "First name cannot contains digits")]
```
dmitrykiev, line 13: "First name cannot contains digits"
what about special symbols?

dmitrykiev (approved):

PR #147, opened 2021-06-07
Dzhus/workshop fix

VyacheslavDzhus:


OutOfSchool/OutOfSchool.WebApi/Controllers/WorkshopController.cs
```diff
@@ -16,7 +16,6 @@ namespace OutOfSchool.WebApi.Controllers
     /// </summary>
     [ApiController]
     [Route("[controller]/[action]")]
-    [Authorize(AuthenticationSchemes = "Bearer")]
```
DmyMi, line 19: it's probably better to leave `[Authorize(AuthenticationSchemes = "Bearer")]` on class level and add `[AllowAnonymous]` on methods that should be accessible to anyone. It will make sure we won't forget to protect some endpoints that should be only accessible to authorised users.

DmyMi (changes requested):

DmyMi (approved):

v-ivanchuk (approved):

PR #148, opened 2021-06-07
Krasnova/statistic

P0linux:


v-ivanchuk (approved):

OutOfSchool/OutOfSchool.WebApi/Controllers/StatisticController.cs
```diff
@@ -0,0 +1,94 @@
...
+            this.localizer = localizer;
+        }
+
+        /// <summary>
+        /// Get popular categories.
+        /// </summary>
+        /// <param name="number">The number of entries.</param>
+        /// <returns>List of popular categories.</returns>
+        [HttpGet("{number}")]
+        [ProducesResponseType(StatusCodes.Status200OK)]
```
DmyMi, line 42: For better swagger documentation when you use `Task<IActionResult>` you need to specify the object type in this attribute like `[ProducesResponseType(StatusCodes.Status200OK, Type = typeof(YOUR_DTO_HERE))]`

DmyMi (changes requested):

OlhaHoliak (approved):

OutOfSchool/OutOfSchool.WebApi/Services/StatisticService.cs
```diff
@@ -0,0 +1,129 @@
...
+            this.workshopRepository = workshopRepository;
+            this.categoryRepository = categoryRepository;
+            this.logger = logger;
+        }
+
+        public async Task<IEnumerable<CategoryStatistic>> GetPopularCategories(int number)
+        {
+            logger.Information("Getting popular categories started.");
+
+            var categories = await categoryRepository.GetAll().ConfigureAwait(false);
```
DmyMi, line 48: Check this example [enumerable vs queryable](https://dotnetcorecentral.com/blog/ienumerable-vs-iqueryable/). tl;dr version: our get all method returns every record in the database, so you load a complete list of categories from database every time and then start filtering/processing.
As you get alot of different entities from DB and do some processing with them - try to see if it's possible to make less requests to DB and load less data if you switch to iqueryable (modify repo to include iqueryable methods if needed) so that your filtering, ordering and counting happens in the database. The database queries are printed in console so you can see what's going on

DmyMi (changes requested):

DmyMi (approved):

Bogdan-Hasanov (approved):

Issue #153, opened 2021-06-09
Refactor relationship between parent and child


DmyMi:
Refactor relationship between parent and child so that you can get a list only of your own children

v-ivanchuk:
Done under -> Ivanchuk/Child refactor (#165)

PR #165, opened 2021-06-18
Ivanchuk/Child refactor

v-ivanchuk:
Task #153

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ChildControllerTests.cs
+++ b/OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ChildControllerTests.cs
@@ -93,6 +93,46 @@ namespace OutOfSchool.WebApi.Tests.Controllers
         }
 
         [Test]
+        [TestCase(1)]
+        public async Task GetByParentId_WhenIdIsValid_ShouldReturnOkResultObject(long id)
+        {
+            // Arrange
+            service.Setup(x => x.GetAllByParent(id)).ReturnsAsync(children.Where(p => p.ParentId == id));
+
+            // Act
+            var result = await controller.GetByParentId(id).ConfigureAwait(false) as OkObjectResult;
+
+            // Assert
+            Assert.That(result, Is.Not.Null);
+            Assert.AreEqual(200, result.StatusCode);
+        }
+
+        [Test]
+        [TestCase(0)]
+        public void GetByParentId_WhenIdIsNotValid_ShouldThrowArgumentOutOfRangeException(long id)
+        {
+            // Assert
+            Assert.That(
+                async () => await controller.GetByParentId(id),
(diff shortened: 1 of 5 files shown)


OutOfSchool/OutOfSchool.WebApi/Controllers/ChildController.cs
```diff
@@ -67,6 +68,29 @@ public async Task<IActionResult> GetById(long id)
...
+        /// <returns>List of all children.</returns>
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        [HttpGet("{id}")]
+        public async Task<IActionResult> GetByParentId(long id)
+        {
+            this.ValidateId(id, localizer);
+
+            var children = await service.GetAllByParent(id).ConfigureAwait(false);
```
OlhaHoliak, line 84: I think we need to implement checkout whether ParentId belongs to User, who called method.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ChildControllerTests.cs
```diff
@@ -92,6 +92,46 @@ public async Task GetChildById_WhenIdIsNotValid_ShouldReturnNull(long id)
...
+        {
+            // Arrange
+            service.Setup(x => x.GetAllByParent(id)).ReturnsAsync(children.Where(p => p.ParentId == id));
+
+            // Act
+            var result = await controller.GetByParentId(id).ConfigureAwait(false) as OkObjectResult;
+
+            // Assert
+            Assert.That(result, Is.Not.Null);
+            Assert.AreEqual(200, result.StatusCode);
```
yfedo, line 107: would be nice to use constants for 200 and 204 status code))
cold use this  Assert.IsNotNull(result); for isnot null check))

yfedo (approved):

OutOfSchool/OutOfSchool.WebApi/Controllers/ChildController.cs
v-ivanchuk: Dear @OlhaHoliak,
Thank you a lot for your advice. I think it is a good idea, so I have implemented it.
Please review.
Thank you a lot.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ChildControllerTests.cs
v-ivanchuk: Dear @yfedo,
Thank you a lot for your advice. I have implemented them in the current PR.

OlhaHoliak (approved):

PR #179, opened 2021-06-28
Dzhus/parent user fix

VyacheslavDzhus:


DmyMi (approved):

OutOfSchool/OutOfSchool.IdentityServer/Controllers/AuthController.cs
```diff
@@ -170,11 +175,21 @@ public async Task<IActionResult> Register(RegisterViewModel model)
                 IdentityResult roleAssignResult = IdentityResult.Failed();
 
                 roleAssignResult = await userManager.AddToRoleAsync(user, user.Role);
-              
+
                 if (roleAssignResult.Succeeded)
                 {
                     await signInManager.SignInAsync(user, false);
 
+                    if (user.Role == Role.Parent.ToString().ToLower())
```
DmyMi, line 183: i approved the request, but maybe it is worth to do it in a transaction so that if the parent was not created - we won't have a user without parent somewhere in the system

DmyMi (approved):

OlhaHoliak (approved):

v-ivanchuk (approved):

PR #200, opened 2021-07-07
Ivanchuk/Add keywords to workshop

v-ivanchuk:


yfedo (approved):

OutOfSchool/OutOfSchool.DataAccess/Models/Workshop.cs
```diff
@@ -79,6 +79,8 @@ public class Workshop
         [MaxLength(60)]
         public string ProviderTitle { get; set; } = string.Empty;
 
+        public string Keywords { get; set; } = string.Empty;
```
SergeyNovitsky, line 82: It would be nice to limit string field

v-ivanchuk: Dear @SergeyNovitsky,
Thank you for your advice. This was done. Please check.

SergeyNovitsky (approved):

PR #246, opened 2021-07-26
Dzhus/photo coverage

VyacheslavDzhus:


OutOfSchool/OutOfSchool.WebApi/Services/PhotoStorage/PhotoStorage.cs
```diff
@@ -0,0 +1,350 @@
...
+                }
+
+                if (fileName is null)
+                {
+                    throw new ArgumentNullException(localizer["File name can not be null!."]);
+                }
+
+                var dirPath = Path.GetDirectoryName(FilePath);
+
+                if (!Directory.Exists(dirPath))
```
Bogdan-Hasanov, line 57: I think we don't need to check if it exists, since it will be done in CreateDirectory method https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.createdirectory?redirectedfrom=MSDN&view=net-5.0#System_IO_Directory_CreateDirectory_System_String_

OutOfSchool/OutOfSchool.WebApi/Services/PhotoStorage/PhotoStorage.cs
```diff
@@ -0,0 +1,350 @@
...
+
+                photo.Path = filePath;
+
+                var photoInfo = await repository.Create(photo.ToDomain()).ConfigureAwait(false);
+
+                logger.Information($"Photo with Id = {photoInfo?.Id} created successfully.");
+
+                return photoInfo.ToModel();
+            }
+            catch (ArgumentException ex)
```
Bogdan-Hasanov, line 77: Since the handling logic is the same, we can avoid duplication by using Exception filters.

```suggestion
            catch (Exception ex) when (ex is ArgumentException || ex is IOException)
```

OutOfSchool/OutOfSchool.WebApi/Services/PhotoStorage/PhotoStorage.cs
```diff
@@ -0,0 +1,350 @@
...
+            }
+            catch (IOException ex)
+            {
+                logger.Error($"Process of getting photos failed. {ex}");
+                throw;
+            }
+        }
+
+        /// <inheritdoc/>
+        public async Task<List<string>> GetFilesPaths(long entityId, EntityType entityType)
```
Bogdan-Hasanov, line 262: GetFilesPaths and GetFilePath methods have some code duplications, mb we can extract some code into method that will be reused?

Bogdan-Hasanov (approved):

OutOfSchool/OutOfSchool.WebApi/Models/PhotoDto.cs
```diff
@@ -0,0 +1,23 @@
...
+        [Required]
+        public long EntityId { get; set; }
+
+        [Required]
+        public EntityType EntityType { get; set; }
+
+        [Required]
+        public PhotoExtension PhotoExtension { get; set; }
+
+        public byte[] Photo { get; set; }
```
DmyMi, line 21: Byte array image in JSON will look like an array with **millions** of entries :) 
image
if we stick to JSON then use a **Base64** string. But better use [multipart/form-data](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads?view=aspnetcore-3.0#upload-large-files-with-streaming) as IFormFile has some additional info for easy validation.

yfedo (approved):

DmyMi (changes requested):

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Services/PhotoStorage/Scale.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        /// <summary>
+        /// Get the width of the scaled image according to its given height.
+        /// </summary>
+        /// <param name="size">The source image size.</param>
+        /// <param name="height">The desired image height.</param>
+        /// <returns>Size result of the scaling calculation.</returns>
+        private Size ScaleByHeight(Size size, float height)
+        {
+            var ratio = (float)size.Width / (float)size.Height;
```
h4wk13, line 85: 1. can we have Width and Height properties as float to avoid redundant convert operation ?
2. what if size.Height will be 0 ?

OutOfSchool/OutOfSchool.WebApi/Services/PhotoStorage/Scale.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <param name="targetSize">Target size.</param>
+        public Scale(Size imgSize, Size targetSize)
+        {
+            ImgSize = imgSize;
+            Origin = new Point(0, 0);
+
+            TargetSize
+                = targetSize.Height == 0 && targetSize.Width > 0 ? ScaleByWidth(imgSize, targetSize.Width)
+                : targetSize.Width == 0 && targetSize.Height > 0 ? ScaleByHeight(imgSize, targetSize.Height)
+                : AutoScale(imgSize, targetSize);
```
h4wk13, line 23: 1. Looks like a method
2. Does it make any sense to upscale image with 0 width or height ?

OutOfSchool/OutOfSchool.WebApi/Services/PhotoStorage/Scale.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// Get new sizes of the image to resize,
+        /// The scale calculation will fit the new size till both width and height are contianed,
+        /// So the final image is not cropped and completely fits in the new size.
+        /// </summary>
+        /// <param name="imgSize">Image size.</param>
+        /// <param name="targetSize">Target size.</param>
+        /// <returns>Size result of the scaling calculation.</returns>
+        private Size AutoScale(Size imgSize, Size targetSize)
+        {
+            var orgRatio = (float)imgSize.Width / (float)imgSize.Height;
```
h4wk13, line 99: What if (float)imgSize.Height == 0 ?

OutOfSchool/OutOfSchool.WebApi/appsettings.Development.json
```diff
@@ -12,5 +12,11 @@
     "Url": "http://localhost/",
     "User": "user",
     "Password": "pass"
+  },
+
+  "PhotoSettings": {
```
h4wk13, line 17: Can we have some docs here to explain limitations ?

OutOfSchool/OutOfSchool.WebApi/appsettings.Development.json
```diff
@@ -12,5 +12,11 @@
...
+    "BasePath": "OoS-FileStorage\\Photos",
+    "MaxSize": 10485760
```
h4wk13, line 19: Please add measure units here, it is not clear what does this size means

h4wk13 (changes requested):

OutOfSchool/OutOfSchool.DataAccess/Repository/PhotoRepository.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// Create photo.
+        /// </summary>
+        /// <param name="photo">Photo as byte array.</param>
+        /// <param name="fileName">Photo name.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+        public async Task CreateUpdatePhoto(byte[] photo, string fileName)
+        {
+            var filePath = GenerateFilePath(fileName);
+
+            Directory.CreateDirectory(Path.GetDirectoryName(filePath));
```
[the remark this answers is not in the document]
VyacheslavDzhus: According to the information from the client, we do not  know yet where the files will be stored. (Probably, it will be NAS / NFS). When the information will be in more detail, we can easily add another implementation to the interface.

OutOfSchool/OutOfSchool.DataAccess/Repository/PhotoRepository.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace OutOfSchool.Services.Repository
+{
+    public class PhotoRepository : IPhotoRepository
+    {
+        private readonly string basePhotoPath;
+
+        public PhotoRepository(IConfiguration config)
+        {
+            basePhotoPath = config.GetSection("PhotoSettings:BasePath").Value;
```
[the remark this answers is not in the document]
VyacheslavDzhus: We do not  know yet where it will be stored and for development it is possible to dynamically change it in app settings.

OutOfSchool/OutOfSchool.WebApi/Models/PhotoDto.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace OutOfSchool.WebApi.Models
+{
+    public class PhotoDto
+    {
+        public long Id { get; set; }
+
+        public string FileName { get; set; }
+
+        [Required]
```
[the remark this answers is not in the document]
VyacheslavDzhus: What do you mean by pure ?

OutOfSchool/OutOfSchool.WebApi/Services/PhotoStorage/PhotoStorage.cs
```diff
@@ -0,0 +1,332 @@
...
+                    throw new ArgumentNullException(localizer["File name can not be null!"]);
+                }
+
+                var photos = await GetFilesByName(fileName).ConfigureAwait(false);
+
+                var photo = photos.FirstOrDefault();
+
+                repository.DeletePhoto(Path.Combine(photo.EntityType.ToString(), photo.FileName));
+
+                await repositoryDB.Delete(photo).ConfigureAwait(false);
```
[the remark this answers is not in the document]
VyacheslavDzhus: There are 2 repositories, one is responsible for the files that are stored in the directory, and the second for information about the files in the database.

OutOfSchool/OutOfSchool.DataAccess/Repository/PhotoRepository.cs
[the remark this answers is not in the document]
h4wk13: unfortunately there is no possibility to mark part of the string to comment, or I cannot find it.
It was about "PhotoSettings:BasePath"

OutOfSchool/OutOfSchool.DataAccess/Repository/PhotoRepository.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <param name="fileName">Photo name.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+        public async Task CreateUpdatePhoto(byte[] photo, string fileName)
+        {
+            var filePath = GenerateFilePath(fileName);
+
+            Directory.CreateDirectory(Path.GetDirectoryName(filePath));
+
+            await File.WriteAllBytesAsync(filePath, photo).ConfigureAwait(false);
+        }
```
yfedo, line 29: public Task CreateUpdatePhoto(byte[] photo, string fileName)
        {
            var filePath = GenerateFilePath(fileName);

            Directory.CreateDirectory(Path.GetDirectoryName(filePath));

            return File.WriteAllBytesAsync(filePath, photo);
        }
you can avoid unnecessary await here and just return task that will be awaited somewhere higher))

OutOfSchool/OutOfSchool.DataAccess/Repository/PhotoRepository.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        /// <summary>
+        /// Returns photo by path.
+        /// </summary>
+        /// <param name="fileName">Photo name.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+        public async Task<byte[]> GetPhoto(string fileName)
+        {
+            return await File.ReadAllBytesAsync(GenerateFilePath(fileName)).ConfigureAwait(false);
+        }
```
yfedo, line 48: public Task<byte[]> GetPhoto(string fileName) => File.ReadAllBytesAsync(GenerateFilePath(fileName));
 looks like await here can also be omitted here))

yfedo (approved):

OutOfSchool/OutOfSchool.DataAccess/Repository/PhotoRepository.cs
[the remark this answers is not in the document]
VadymLevkovskyi: Backup, moving data to other server, access rights, ... - for case of separate folder has higher maintainability cost.

OutOfSchool/OutOfSchool.WebApi/Controllers/PhotoController.cs
```diff
@@ -0,0 +1,478 @@
...
+            this.maxSize = config.GetValue<long>("PhotoSettings:MaxSizeBytes");
+        }
+
+        /// <summary>
+        /// Get file by it's name.
+        /// </summary>
+        /// <param name="fileName">Name of the file.</param>
+        /// <returns>Photo.</returns>
+        [HttpGet]
+        [Authorize(Roles = "parent")]
```
VadymLevkovskyi, line 49: From [Role-based authorization in Asp.Net Core](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/roles?view=aspnetcore-3.0)

> If you apply multiple attributes then an accessing user must be a member of all the roles specified

is it really expected for user to have all specified roles?

OutOfSchool/OutOfSchool.WebApi/Controllers/PhotoController.cs
```diff
@@ -0,0 +1,478 @@
...
+            }
+            catch (Exception ex) when (ex is ArgumentException || ex is IOException || ex is InvalidOperationException)
+            {
+                return BadRequest("An error occurred while updating the photo.");
+            }
+        }
+
+        private bool IsSizeValid(IFormFile photo)
+        {
+            if (photo.Length > maxSize)
```
VadymLevkovskyi, line 435: `if` might be simplified

OutOfSchool/OutOfSchool.WebApi/Controllers/PhotoController.cs
```diff
@@ -0,0 +1,478 @@
...
+            {
+                return true;
+            }
+        }
+
+        private bool IsExtensionValid(IFormFile photo)
+        {
+            var contentType = GetFormat(photo.FileName);
+
+            if (string.IsNullOrEmpty(contentType))
```
VadymLevkovskyi, line 449: `if` might be simplified

OutOfSchool/OutOfSchool.WebApi/Controllers/PhotoController.cs
```diff
@@ -0,0 +1,478 @@
...
+        private string GetFormat(string fileName)
+        {
+            string contentType;
+
+            if (!contentTypeProvider.TryGetContentType(fileName, out contentType))
+            {
+                contentType = string.Empty;
+            }
+
+            if (contentType is ImageTypeNames.Jpeg || contentType is ImageTypeNames.Png)
```
VadymLevkovskyi, line 468: `if` might be simplified. String comparison using `is` operator looks weird

OutOfSchool/OutOfSchool.WebApi/Controllers/PhotoController.cs
```diff
@@ -0,0 +1,478 @@
...
+            }
+        }
+
+        private string GetFormat(string fileName)
+        {
+            string contentType;
+
+            if (!contentTypeProvider.TryGetContentType(fileName, out contentType))
+            {
+                contentType = string.Empty;
```
VadymLevkovskyi, line 465: Immediate `return string.Empty;` might be done here

OutOfSchool/OutOfSchool.WebApi/Controllers/PhotoController.cs
DmyMi: Adding especially for **this** endpoint - it should be anonymous as photos should be visible even for unregistered users.

OutOfSchool/OutOfSchool.WebApi/Controllers/PhotoController.cs
```diff
@@ -0,0 +1,478 @@
...
+        /// <summary>
+        /// Get file by it's name.
+        /// </summary>
+        /// <param name="fileName">Name of the file.</param>
+        /// <returns>Photo.</returns>
+        [HttpGet]
+        [Authorize(Roles = "parent")]
+        [Authorize(Roles = "provider")]
+        [Authorize(Roles = "admin")]
+        [ProducesResponseType(StatusCodes.Status200OK)]
```
DmyMi, line 52: For this and other endpoints if using `IActionResult` - please specify response types in attributes so swagger documents generate correct data.

DmyMi (approved):

VadymLevkovskyi:
Shouldn't this PR be marked as Draft or closed at all?

PR #278, opened 2021-08-19
Krasnova/statistic get workshops update

P0linux:


DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi.Tests/Services/StatisticServiceTest.cs
```diff
@@ -18,68 +18,183 @@ namespace OutOfSchool.WebApi.Tests.Services
...
         }
 
         [Test]
         public async Task GetPopularWorkshops_ShouldReturnWorkshops()
         {
-            var directions = context.Directions;
-
             // Arrange
-            var expected = new WorkshopDTO
+            var expected = new List<WorkshopCard>
```
h4wk13, line 57: Please move creating of random data into separate methods

OutOfSchool/OutOfSchool.WebApi.Tests/Services/StatisticServiceTest.cs
```diff
@@ -18,68 +18,183 @@ namespace OutOfSchool.WebApi.Tests.Services
...
+                It.IsAny<bool>()))
+                .Returns(workshopsMock.Object);
+
             // Act
             var result = await service.GetPopularWorkshops(2).ConfigureAwait(false);
 
             // Assert
             result.Should().HaveCount(2);
-            result.Should().ContainEquivalentOf(expected);
+            result.Should().BeEquivalentTo(expected);
```
h4wk13, line 87: You can use [CollectionAssert.AreEqual](https://docs.nunit.org/articles/nunit/writing-tests/assertions/classic-assertions/Collection-Assert.html)
It makes sure that both collections contains same items in the same order.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/StatisticServiceTest.cs
```diff
@@ -18,68 +18,183 @@ namespace OutOfSchool.WebApi.Tests.Services
...
                 WorkshopsCount = 2,
                 ApplicationsCount = 2,
             };
 
+            var workshopsMock = workshops.AsQueryable().BuildMock();
+            var applicationsMock = applications.AsQueryable().BuildMock();
+            var directionsMock = directions.AsQueryable().BuildMock();
+
+            workshopRepository.Setup(w => w.Get<It.IsAnyType>(
+                It.IsAny<int>(),
```
h4wk13, line 106: I've seen exact code in the previous test.
Can we move all generic setups in the separate method ?

OutOfSchool/OutOfSchool.WebApi/Services/StatisticService.cs
```diff
@@ -113,11 +113,11 @@ public async Task<IEnumerable<DirectionStatistic>> GetPopularDirections(int limi
         }
 
         /// <inheritdoc/>
-        public async Task<IEnumerable<WorkshopDTO>> GetPopularWorkshops(int limit)
+        public async Task<IEnumerable<WorkshopCard>> GetPopularWorkshops(int limit)
         {
             logger.Information("Getting popular workshops started.");
 
-            var workshops = workshopRepository.Get<int>();
+            var workshops = workshopRepository.Get<int>(includeProperties: "Address,Direction");
```
h4wk13, line 120: can we use something like a $"{nameof(Address)},{nameof(Direction)}" instead of string literal ?

h4wk13 (approved):

OutOfSchool/OutOfSchool.WebApi.Tests/Services/StatisticServiceTest.cs
P0linux: Could I use `Should().BeEquivalentTo()` with option [WithStrictOrdering] (https://fluentassertions.com/objectgraphs/#ordering) instead? 
It is supposed to do the same: comparing lists members considering ordering.

PR #286, opened 2021-08-24
updated create repository flow.

h4wk13:
Removed redundant properties.


OutOfSchool/OutOfSchool.WebApi/Models/ProviderDto.cs
```diff
@@ -65,26 +68,26 @@ public class ProviderDto
 
         [Required]
         [MaxLength(30)]
+        // TODO: looks like it should be changed to person.
         public string Founder { get; set; } = string.Empty;
 
+        // TODO: it will be always filled because of enum type
```
VadymLevkovskyi, line 74: This required is more hint for model binder than c# code. That's why we still need this `RequiredAttribute`

VadymLevkovskyi (approved):

PR #299, opened 2021-08-29
updated Provider entity: 'Actual' address now optional and can be null

h4wk13:
updated Provider entity: 'Actual' address now optional and can be null, added migration.
Updated unit tests

### DIFF
--- a/OutOfSchool/OutOfSchool.DataAccess/Repository/ProviderRepository.cs
+++ b/OutOfSchool/OutOfSchool.DataAccess/Repository/ProviderRepository.cs
@@ -58,14 +58,13 @@ namespace OutOfSchool.Services.Repository
         {
             db.Entry(entity).State = EntityState.Deleted;
 
-            if (entity.LegalAddressId == entity.ActualAddressId)
+            // TODO: Q: why we are using new entity instead of Provider.LegalAddress here and belowe ?
+            db.Entry(new Address { Id = entity.LegalAddressId }).State = EntityState.Deleted;
+
+            if (entity.ActualAddressId.HasValue
+                && entity.ActualAddressId.Value == entity.LegalAddressId)
             {
-                db.Entry(new Address { Id = entity.LegalAddressId }).State = EntityState.Deleted;
-            }
-            else
-            {
-                db.Entry(new Address { Id = entity.LegalAddressId }).State = EntityState.Deleted;
-                db.Entry(new Address { Id = entity.ActualAddressId }).State = EntityState.Deleted;
+                db.Entry(new Address { Id = entity.ActualAddressId.Value }).State = EntityState.Deleted;
             }
 
             await db.SaveChangesAsync();
--- a/OutOfSchool/OutOfSchool.DataAccess/Repository/ProviderRepository.cs
+++ b/OutOfSchool/OutOfSchool.DataAccess/Repository/ProviderRepository.cs
@@ -62,7 +62,7 @@ namespace OutOfSchool.Services.Repository
             db.Entry(new Address { Id = entity.LegalAddressId }).State = EntityState.Deleted;
 
             if (entity.ActualAddressId.HasValue
-                && entity.ActualAddressId.Value == entity.LegalAddressId)
+                && entity.ActualAddressId.Value != entity.LegalAddressId)
             {
                 db.Entry(new Address { Id = entity.ActualAddressId.Value }).State = EntityState.Deleted;
             }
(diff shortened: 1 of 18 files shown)


OutOfSchool/OutOfSchool.IdentityServer/Data/Migrations/OutOfSchoolMigrations/20210828133338_ActualAddressCanBeNull.cs
```diff
@@ -0,0 +1,55 @@
...
+                oldClrType: typeof(long),
+                oldType: "bigint");
+
+            migrationBuilder.AddForeignKey(
+                name: "FK_Providers_Addresses_ActualAddressId",
+                table: "Providers",
+                column: "ActualAddressId",
+                principalTable: "Addresses",
+                principalColumn: "Id",
+                onDelete: ReferentialAction.Restrict);
```
VadymLevkovskyi, line 27: Why not `ReferentialAction.Cascade` ?

h4wk13: because our foreign key now nullable.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ProviderServiceTests.cs
```diff
@@ -1,611 +1,374 @@
...
+        private List<Provider> fakeProviders = ProvidersGenerator.Generate(10);
+        private User fakeUser;
+
         [SetUp]
         public void SetUp()
         {
-            var builder =
-                new DbContextOptionsBuilder<OutOfSchoolDbContext>().UseInMemoryDatabase(
-                    databaseName: "OutOfSchoolTestDB").ConfigureWarnings(w => w.Ignore(InMemoryEventId.TransactionIgnoredWarning));
+            //fakeProviders = CreateFakeProviders();
```
VadymLevkovskyi, line 45: Pls remove comment

DmyMi (approved):

OutOfSchool/OutOfSchool.IdentityServer/Data/Migrations/OutOfSchoolMigrations/20210828133338_ActualAddressCanBeNull.Designer.cs
```diff
@@ -0,0 +1,1286 @@
...
+                        .IsRequired();
+
+                    b.Navigation("User");
+                });
+
+            modelBuilder.Entity("OutOfSchool.Services.Models.Provider", b =>
+                {
+                    b.HasOne("OutOfSchool.Services.Models.Address", "ActualAddress")
+                        .WithMany()
+                        .HasForeignKey("ActualAddressId");
```
OlhaHoliak, line 1153: I think we need to add ` .OnDelete(DeleteBehavior.Cascade)`
Now when we delete Provider it does not delete the Actual address from DB.

h4wk13: I will update provider model creation in dbContext later so EF will be able to prepare its migration better

OutOfSchool/OutOfSchool.IdentityServer/Data/Migrations/OutOfSchoolMigrations/20210828133338_ActualAddressCanBeNull.cs
OlhaHoliak: When Cascade is set, so if we delete Actual address we will delete Provider.
May be it's not bad to set Restrict.
But consider that if we set DeleteBehavior of EF Core like Cascade and delete Address(with tracking Provider in the navigation property) it will delete Provider anyway.

OutOfSchool/OutOfSchool.DataAccess/Repository/ProviderRepository.cs
```diff
@@ -58,14 +58,13 @@ public new async Task Delete(Provider entity)
...
-            if (entity.LegalAddressId == entity.ActualAddressId)
-            {
-                db.Entry(new Address { Id = entity.LegalAddressId }).State = EntityState.Deleted;
-            }
-            else
+            // TODO: Q: why we are using new entity instead of Provider.LegalAddress here and belowe ?
+            db.Entry(new Address { Id = entity.LegalAddressId }).State = EntityState.Deleted;
+
+            if (entity.ActualAddressId.HasValue
+                && entity.ActualAddressId.Value == entity.LegalAddressId)
```
OlhaHoliak, line 65: Change `==` to `!=`
As we need to delete ActualAddress if it is not equal to Legal

OutOfSchool/OutOfSchool.IdentityServer/Data/Migrations/OutOfSchoolMigrations/20210828133338_ActualAddressCanBeNull.cs
h4wk13: migration recreated

OlhaHoliak (approved):

PR #306, opened 2021-08-30
Local deployment on par with future prod

DmyMi:
# Disclaimer
Lots of changes, but more than half of them are due to moving files from one namespace to another.

This is a first set of changes. Rest will mostly include changes to compose files & nginx config.

# Contents
- Add api versioning (with backward compatibility so frontend can transition seamlessly) to easily support breaking changes to production later. As per [MS Recommendations](https://docs.microsoft.com/en-us/azure/architecture/best-practices/api-design#versioning-a-restful-web-api) & [example on the internet](https://dev.to/moesmp/what-every-asp-net-core-web-api-project-needs-part-2-api-versioning-and-swagger-3nfm)
- Add reverse proxy configuration with optional base path as per [MS Documentation](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-3.1). Depending on final deployment we will either have a single host & basepath or two virtual hosts & no basepath. This change allows to work with both approaches without changing any part of application except `appsettings` & some env variables.
- Refactor delay in [Program.cs](https://github.com/ita-social-projects/OoS-Backend/blob/350bfde3ebd4f45a6b597f3fa77e017ce31ff574/OutOfSchool/OutOfSchool.IdentityServer/Program.cs#L33) and increased it to 5 seconds so that Identity won't spam SQL with requests on startup.
- Add first iteration of Nginx config for local deployment in compose
- Add Compose environment settings for local compose deployments.
- Add local automated dev environment to simulate production. It works using a single set of tools & one command on MacOS (tested on Big Sur 11.5), Linux (tested on Ubuntu 20.04.3) & Windows 10 (but on Windows `work` means it doesn't ~50% of time, depending on Win10 build :) )
- Add supplementary scripts (powershell & bash) to support local dev environment
- Add first iteration of Readme & with aforementioned environment launch docs


docker-compose.yml
```diff
@@ -1,23 +1,93 @@
 ﻿version: '3.4'
```
h4wk13, line 1: mistype ?

docker-compose.local.yml
```diff
@@ -0,0 +1,54 @@
+﻿version: '3.4'
```
h4wk13, line 1: mistype ?

h4wk13 (approved):

docker-compose.local.yml
DmyMi: fixed

docker-compose.yml
DmyMi: fixed

docker-compose.m1.yml
```diff
@@ -0,0 +1,4 @@
+﻿version: '3.4'
```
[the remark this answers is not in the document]
DmyMi: fixed

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AddressController.cs
```diff
@@ -8,9 +8,11 @@
 using OutOfSchool.WebApi.Models;
 using OutOfSchool.WebApi.Services;
 
-namespace OutOfSchool.WebApi.Controllers
+namespace OutOfSchool.WebApi.Controllers.V1
 {
     [ApiController]
+    [ApiVersion("1.0")]
+    [Route("api/v{version:apiVersion}/[controller]/[action]")]
```
VadymLevkovskyi, line 15: Instead of changing each and every Controller couldn't routing be configured via [Conventional routing](https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/routing?view=aspnetcore-5.0#conventional-routing) ? (In `Startup.Configure`)

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopController.cs
```diff
@@ -107,7 +109,7 @@ public async Task<IActionResult> GetByFilter([FromQuery] WorkshopFilter filter)
         {
             if (string.IsNullOrWhiteSpace(filter.City))
             {
-                filter.City = "���";
+                filter.City = "Київ";
```
VadymLevkovskyi, line 112: "Київ" - should be `Default_City` constant introduced somewhere, but not in `WorkshopController`

OutOfSchool/OutOfSchool.IdentityServer/appsettings.Compose.json
```diff
@@ -0,0 +1,36 @@
+﻿{
+  "ConnectionStrings": {
+    "DefaultConnection": "Server=sql-server; Database=Master;User Id=SA;Password=Oos-password1"
```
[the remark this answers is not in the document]
DmyMi: will update this in next PR.

VyacheslavDzhus (approved):

VadymLevkovskyi (approved):

PR #309, opened 2021-08-31
Changed the result of the  endpoint GetByProviderId from lists of Wor…

VyacheslavDzhus:
…kshops to list of WorkshopCards.

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopControllerTests.cs
+++ b/OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopControllerTests.cs
@@ -125,7 +125,7 @@ namespace OutOfSchool.WebApi.Tests.Controllers
         public void GetByProviderId_WhenIdIsInvalid_ShouldThrowArgumentOutOfRangeException(long id)
         {
             // Arrange
-            workshopServiceMoq.Setup(x => x.GetByProviderId(id)).ReturnsAsync(workshopCards.Where(x => x.ProviderId == id));
+            workshopServiceMoq.Setup(x => x.GetByProviderId(id)).ReturnsAsync(workshopCards.Where(x => x.ProviderId == id).ToList());
 
             // Assert
             Assert.That(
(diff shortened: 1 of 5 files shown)


OutOfSchool/OutOfSchool.WebApi/Services/WorkshopServicesCombiner.cs
```diff
@@ -119,20 +119,23 @@ public async Task<SearchResult<WorkshopCard>> GetByFilter(WorkshopFilter filter)
         }
 
         /// <inheritdoc/>
-        public async Task<IEnumerable<WorkshopDTO>> GetByProviderId(long id)
+        public async Task<IEnumerable<WorkshopCard>> GetByProviderId(long id)
```
VadymLevkovskyi, line 122: I feel some code smell in naming - `GetById()` returns `WorkshopDto`, while `GetByProviderId()` returns `IEnumerable<WorkshopCard>`. Results are different - names are (almost) similar. I see that we have such situation before this PR, but I'd like to clarify shouldn't it be improved

OutOfSchool/OutOfSchool.WebApi/Services/WorkshopServicesCombiner.cs
```diff
@@ -119,20 +119,23 @@ public async Task<SearchResult<WorkshopCard>> GetByFilter(WorkshopFilter filter)
...
+
+            return workshopCards;
         }
 
         private List<WorkshopCard> DtoModelsToWorkshopCards(IEnumerable<WorkshopDTO> source)
         {
             List<WorkshopCard> workshopCards = new List<WorkshopCard>();
-            foreach (var item in source)
+
+            Parallel.ForEach(source, currentElement =>
```
VadymLevkovskyi, line 135: `Parallel.ForEach` seems to be bit excessive here

VadymLevkovskyi: You could use instead simple 
`var workshopCards = source.Select(currentElement => currentElement.ToCardDto()).ToList()`

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopControllerTests.cs
```diff
@@ -130,23 +130,23 @@ public void GetByProviderId_WhenIdIsInvalid_ShouldThrowArgumentOutOfRangeExcepti
...
         {
             // Arrange
-            workshopServiceMoq.Setup(x => x.GetByProviderId(id)).ReturnsAsync(workshops.Where(x => x.ProviderId == id));
+            workshopServiceMoq.Setup(x => x.GetByProviderId(id)).ReturnsAsync(workshopCards.Where(x => x.ProviderId == id));
 
             // Act
             var result = await controller.GetByProviderId(id).ConfigureAwait(false) as OkObjectResult;
 
             // Assert
             Assert.That(result, Is.Not.Null);
```
VadymLevkovskyi, line 139: I don't insist, just recommendation - not to mix 2 styles of `Assert`s: `.That` and `.Are...`. I know it's not scope of this PR, just discussion point

OlhaHoliak (approved):

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopControllerTests.cs
```diff
@@ -117,7 +117,7 @@ public async Task GetWorkshopById_WhenThereIsNoWorkshopWithId_ShouldReturnNoCont
         public void GetByProviderId_WhenIdIsInvalid_ShouldThrowArgumentOutOfRangeException(long id)
         {
```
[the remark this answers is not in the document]
h4wk13: `[Test]` attribute is not valid here since this test have input argument and which cannot be provided by the `[Test]` attribute .

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopControllerTests.cs
```diff
@@ -130,23 +130,23 @@ public void GetByProviderId_WhenIdIsInvalid_ShouldThrowArgumentOutOfRangeExcepti
...
             var result = await controller.GetByProviderId(id).ConfigureAwait(false) as OkObjectResult;
 
             // Assert
             Assert.That(result, Is.Not.Null);
             Assert.AreEqual(Ok, result.StatusCode);
-            Assert.AreEqual(2, (result.Value as IEnumerable<WorkshopDTO>).Count());
+            Assert.AreEqual(2, (result.Value as IEnumerable<WorkshopCard>).Count());
         }
 
         [Test]
```
h4wk13, line 144: Redundant `[Test]` attribute

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopControllerTests.cs
```diff
@@ -130,23 +130,23 @@ public void GetByProviderId_WhenIdIsInvalid_ShouldThrowArgumentOutOfRangeExcepti
...
         [TestCase(3)]
```
h4wk13, line 145: Why 3 ?
I think it will be better to configure mock with nonempty collection and get id value from outside of the existing ids. like
`workshops.Select(w => w.Id).Max() + 1`  
`workshops.Select(w => w.Id).Max() + randomNumber`
You can randomize your data with [RandomAttribute](https://docs.nunit.org/articles/nunit/writing-tests/attributes/random.html)

OutOfSchool/OutOfSchool.WebApi/Services/IWorkshopServicesCombiner.cs
```diff
@@ -18,12 +18,12 @@ public interface IWorkshopServicesCombiner : ICRUDService<WorkshopDTO>
...
-        /// Get all workshops with the specified provider's Id.
+        /// Get all workshop cards with the specified provider's Id.
         /// </summary>
         /// <param name="id">Provider's key.</param>
-        /// <returns>A <see cref="Task{WorkshopDTO}"/> representing the result of the asynchronous operation.
-        /// The task result contains a <see cref="IEnumerable{WorkshopDTO}"/> that contains elements from the input sequence.</returns>
-        Task<IEnumerable<WorkshopDTO>> GetByProviderId(long id);
+        /// <returns>A <see cref="Task{WorkshopCard}"/> representing the result of the asynchronous operation.
+        /// The task result contains a <see cref="IEnumerable{WorkshopCard}"/> that contains elements from the input sequence.</returns>
+        Task<IEnumerable<WorkshopCard>> GetByProviderId(long id);
```
h4wk13, line 26: Why we are returning `IEnumerable` instead of full type which we are already have with all it metadata ?

OutOfSchool/OutOfSchool.WebApi/Services/WorkshopServicesCombiner.cs
```diff
@@ -119,20 +119,23 @@ public async Task<SearchResult<WorkshopCard>> GetByFilter(WorkshopFilter filter)
...
-        public async Task<IEnumerable<WorkshopDTO>> GetByProviderId(long id)
+        public async Task<IEnumerable<WorkshopCard>> GetByProviderId(long id)
         {
-            var workshop = await databaseService.GetByProviderId(id).ConfigureAwait(false);
+            var workshops = await databaseService.GetByProviderId(id).ConfigureAwait(false);
 
-            return workshop;
+            var workshopCards = DtoModelsToWorkshopCards(workshops);
+
+            return workshopCards;
```
h4wk13, line 128: redundant temporary variable ?

OutOfSchool/OutOfSchool.WebApi/Services/WorkshopServicesCombiner.cs
h4wk13: 1. Writing to List{T} [is not thread safe](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1?view=net-5.0#thread-safety)
2. Be aware that parallelization here will change items order inside the collection so it will be hard to add a pagination functionality in the future.
If you still want to use parallelization here please use thread safe collection or try something like 
`=> source.AsParallel()` ... `code from @VadymLevkovskyi  comment`

h4wk13 (approved):

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopControllerTests.cs
VyacheslavDzhus: Thank you for the advice

OutOfSchool/OutOfSchool.WebApi/Services/WorkshopServicesCombiner.cs
VadymLevkovskyi: Well, my "vision" is that method name should reflect it's action and result, otherwise we have to check method implementation, that breaks abstraction and encapsulation principles. 

As it's only about naming - I don't insist, but my inner perfectionist is sad )

OutOfSchool/OutOfSchool.WebApi/Services/WorkshopServicesCombiner.cs
VadymLevkovskyi: @h4wk13 please be aware, that I'm not suggesting to use `ToList()` inside `Parallel.ForEach()`. I was talking about total replacing of `Parallel.ForEach()` by simple `Select()`.

VadymLevkovskyi: Discussion seems to be outdated, and nothing is touched, so I set this thread to resolved

PR #316, opened 2021-09-06
Olhaholiak/Remove messageDto creation to service

OlhaHoliak:
1. Creation of new message DTO is removed from Hub/Controller method to the ChatMessageWorkshopService
2. Changed some methods' userRole parameter from string to enum
3. Deleted some Controller's methods, because there are no business requirements for them yet.


Bogdan-Hasanov (approved):

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ValidationServiceTests.cs
```diff
@@ -72,15 +72,18 @@ public async Task UserIsProviderOwnerAsync_WhenFalse_ReturnsFalse()
         }
 
         [Test]
-        public void UserIsProviderOwnerAsync_WhenEntityWasNotFaund_ThrowsException()
+        public async Task UserIsProviderOwnerAsync_WhenEntityWasNotFaund_ReturnsFalse()
```
VadymLevkovskyi, line 75: mistype `Faund` - should be `Found`. I see that it's not your change, but would be good to fix it as well )

PR #318, opened 2021-09-07
chat controller methods are splited by user role

OlhaHoliak:


Bogdan-Hasanov (approved):

OutOfSchool/OutOfSchool.WebApi/Controllers/ChatWorkshopController.cs
```diff
@@ -56,30 +56,179 @@ public class ChatWorkshopController : ControllerBase
...
+        [HttpGet("parent/chatrooms/{id}")]
+        [Authorize(Roles = "parent")]
         [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(ChatRoomWorkshopDto))]
         [ProducesResponseType(StatusCodes.Status204NoContent)]
         [ProducesResponseType(StatusCodes.Status401Unauthorized)]
         [ProducesResponseType(StatusCodes.Status403Forbidden)]
         [ProducesResponseType(StatusCodes.Status500InternalServerError)]
-        public async Task<IActionResult> GetRoomByIdAsync([Range(1, long.MaxValue)] long id)
+        public Task<IActionResult> GetRoomForParentByRoomIdAsync([Range(1, long.MaxValue)] long id)
+            => this.GetRoomByIdAsync(id, chatRoom => this.ParentHasRigtsForChatRoomAsync(chatRoom.ParentId));
```
VadymLevkovskyi, line 71: 1. This lambda `chatRoom => this.ParentHasRigtsForChatRoomAsync(chatRoom.ParentId)` and it's twin is heavily duplicated. Why not introduce helper method that can be reused?
2. `ParentHasRigtsForChatRoomAsync` - please, fix the mistype - it should be `...Rights..`

OlhaHoliak: thank you very much!
Done.
I changed the original method to take not Id but ChatRoom.

VadymLevkovskyi: Cool, this is even better

VadymLevkovskyi (approved):

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Controllers/ChatWorkshopController.cs
```diff
@@ -56,30 +56,179 @@ public class ChatWorkshopController : ControllerBase
...
 
         /// <summary>
-        /// Get a chat room with information about Parent and Workshop.
+        /// Get parent's chat room with information about Parent and Workshop.
         /// </summary>
         /// <param name="id">ChatRoom's Id.</param>
         /// <returns>User's chat room that was found.</returns>
-        [HttpGet("chatrooms/{id}")]
+        [HttpGet("parent/chatrooms/{id}")]
+        [Authorize(Roles = "parent")]
```
h4wk13, line 64: nameof(Roles.parent) ?

OutOfSchool/OutOfSchool.WebApi/Controllers/ChatWorkshopController.cs
```diff
@@ -56,30 +56,179 @@ public class ChatWorkshopController : ControllerBase
...
+        }
+
+        private Task<bool> IsProviderAChatRoomParticipantAsync(ChatRoomWorkshopDto chatRoom)
+        {
+            var userId = this.GetUserId();
+
+            return validationService.UserIsWorkshopOwnerAsync(userId, chatRoom.WorkshopId);
+        }
+
+        private string GetUserId()
```
h4wk13, line 194: `=> HttpContext.User.GetUserPropertyByClaimType(IdentityResourceClaimsTypes.Sub)
                ?? throw new AuthenticationException($"Can not get user's claim {nameof(IdentityResourceClaimsTypes.Sub)} from HttpContext.");`
?
But we will throw that exception instead of return an HttpError, is it Ok ?

OutOfSchool/OutOfSchool.WebApi/Controllers/ChatWorkshopController.cs
```diff
@@ -56,30 +56,179 @@ public class ChatWorkshopController : ControllerBase
...
+                ?? throw new AuthenticationException($"Can not get user's claim {nameof(IdentityResourceClaimsTypes.Sub)} from HttpContext.");
+
+            Role userRole = (Role)Enum.Parse(typeof(Role), userRoleName, true);
+
+            return userRole;
+        }
+
+        private void LogWarningAboutUsersTryingToGetNotOwnChatRoomInformation(long chatRoomId)
+        {
+            var messageToLog = $"{this.GetUserRole()} with UserId:{this.GetUserId()} is trying to get not his own chat room: {nameof(chatRoomId)}={chatRoomId}.";
```
h4wk13, line 214: What if `GetUserRole` will throw an exception ?

OutOfSchool/OutOfSchool.WebApi/Controllers/ChatWorkshopController.cs
```diff
@@ -56,30 +56,179 @@ public class ChatWorkshopController : ControllerBase
...
+                var isChatRoomValid = !(chatRoom is null) && await userHasRights(chatRoom).ConfigureAwait(false);
+
+                if (isChatRoomValid)
                 {
                     return Ok(chatRoom);
                 }
 
-                var messageToLog = $"{HttpContext.User.GetUserPropertyByClaimType(IdentityResourceClaimsTypes.Role)} with UserId:{HttpContext.User.GetUserPropertyByClaimType(IdentityResourceClaimsTypes.Sub)} is trying to get not his own chat room: {nameof(id)} {id}.";
-                logger.Warning(messageToLog);
+                this.LogWarningAboutUsersTryingToGetNotOwnChatRoomInformation(chatRoomId);
```
h4wk13, line 231: so if with such id still does not exist warning about access will be logged ?

OutOfSchool/OutOfSchool.WebApi/Controllers/ChatWorkshopController.cs
```diff
@@ -56,30 +56,179 @@ public class ChatWorkshopController : ControllerBase
...
+        private async Task<IActionResult> GetRoomByIdAsync(long chatRoomId, Func<ChatRoomWorkshopDto, Task<bool>> userHasRights)
         {
             try
             {
-                var chatRoom = await roomService.GetByIdAsync(id).ConfigureAwait(false);
+                var chatRoom = await roomService.GetByIdAsync(chatRoomId).ConfigureAwait(false);
 
-                if (!(chatRoom is null) &&
-                    await this.UserHasRigtsForChatRoomAsync(chatRoom.WorkshopId, chatRoom.ParentId).ConfigureAwait(false))
+                var isChatRoomValid = !(chatRoom is null) && await userHasRights(chatRoom).ConfigureAwait(false);
```
h4wk13, line 224: can we use `chatRoom != null` instead of using two logical operations ?

OutOfSchool/OutOfSchool.WebApi/Controllers/ChatWorkshopController.cs
OlhaHoliak: In each controller method, we catch `AuthenticationException` and log it, and return BadRequest to the user with the neutral message "Can not get some user's claims. Please check your authentication or contact technical support."

OutOfSchool/OutOfSchool.WebApi/Controllers/ChatWorkshopController.cs
OlhaHoliak: 1. In each controller method, we catch AuthenticationException (that is thrown by this method) and log it, and return BadRequest to the user.
2. This is a helper method, so we use it after all these methods (GetUserRole and GetUserId) have been already used before to figure out if the user has rights. Here we only log warnings about inappropriate user behavior.

OutOfSchool/OutOfSchool.WebApi/Controllers/ChatWorkshopController.cs
OlhaHoliak: No, if the chat room with such Id does not exist, or the user has no rights to it, we return NoContent. We don't tell the user about another, not his own existing chat rooms.
And we log that the user is trying to get forbidden to him chat room because if the room does not exist it is forbidden too.

OutOfSchool/OutOfSchool.WebApi/Controllers/ChatWorkshopController.cs
OlhaHoliak: This is a great idea. Also, this issue is common for the entire solution, so first, we need to rename the enum to lower case. I think we can make it in a separate task.

PR #340, opened 2021-09-20
Khasanov/TimeRanges

Bogdan-Hasanov:


OutOfSchool/OutOfSchool.DataAccess/Repository/WorkshopRepository.cs
```diff
@@ -11,9 +8,10 @@ namespace OutOfSchool.Services.Repository
     public class WorkshopRepository : EntityRepository<Workshop>, IWorkshopRepository
     {
         private readonly OutOfSchoolDbContext db;
```
VadymLevkovskyi, line 10: (To write less code and make linter happier) 
You could use getter-only property instead of backing field and expression-bodied getter-only property, like in my example (`TestPropertySetterOutsideCtor()` is used only to show that setting of the property is allowed only via ctor, it shouldn't be included in your code):

image

VadymLevkovskyi: And now I've realized, that you are using `OutOfSchoolDbContext db` everywhere. Then at least add empty line between field and property for linter.

OutOfSchool/OutOfSchool.IdentityServer/Data/Migrations/OutOfSchoolMigrations/20210920144554_TimeRanges.cs
```diff
@@ -0,0 +1,82 @@
...
+                table: "Workshops");
+
+            migrationBuilder.CreateTable(
+                name: "DateTimeRanges",
+                columns: table => new
+                {
+                    Id = table.Column<long>(type: "bigint", nullable: false)
+                        .Annotation("SqlServer:Identity", "1, 1"),
+                    StartTime = table.Column<TimeSpan>(type: "time", nullable: false),
+                    EndTime = table.Column<TimeSpan>(type: "time", nullable: false),
```
VadymLevkovskyi, line 21: shouldn't we use `.HasCheckConstraint("CK_DateTimeRanges_EndTimeIsAfterStartTime", "[EndTime] >= [StartTime]" )` for `EndTime` column?

OutOfSchool/OutOfSchool.IdentityServer/Data/Migrations/OutOfSchoolMigrations/20210920144554_TimeRanges.cs
```diff
@@ -0,0 +1,82 @@
...
+                });
+
+            migrationBuilder.CreateTable(
+                name: "Workdays",
+                columns: table => new
+                {
+                    Id = table.Column<long>(type: "bigint", nullable: false)
+                        .Annotation("SqlServer:Identity", "1, 1"),
+                    DateTimeRangeId = table.Column<long>(type: "bigint", nullable: false),
+                    DayOfWeek = table.Column<int>(type: "int", nullable: false)
```
VadymLevkovskyi, line 42: Shouldn't we constraint data for 0..6 range?

OutOfSchool/OutOfSchool.WebApi/Models/WorkdayDto.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace OutOfSchool.WebApi.Models
+{
+    public class WorkdayDto
+    {
+        public long Id { get; set; }
```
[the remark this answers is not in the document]
VadymLevkovskyi: (Additional consideration) to implement separate table to store list of max 7 bits (one bit per day) per TimeRange looks correct for db normalization approach, but looks very excessive from performance point of view (extra data, extra join in SQL, extra table blocking on updates...). 
Isn't it easier just to store `byte` data in db and write custom code to convert that byte to `List<DayOfWeek>`?

OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs
```diff
@@ -0,0 +1,84 @@
...
+using OutOfSchool.WebApi.Util.CustomComparers;
+
+namespace OutOfSchool.WebApi.Util
+{
+    public class MappingProfile : Profile
+    {
+        public MappingProfile()
+        {
+            CreateMap<WorkshopDTO, Workshop>()
+                .ForMember(dest => dest.Keywords, opt => opt.MapFrom(src => string.Join('¤', src.Keywords.Distinct())))
```
VadymLevkovskyi, line 16: `'¤'` - should be a char constant introduced somewhere in `WorkshopDto` or `Workshop`

OutOfSchool/OutOfSchool.DataAccess/Models/DateTimeRange.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        public long Id { get; set; }
+
+        public TimeSpan StartTime { get; set; }
+
+        public TimeSpan EndTime { get; set; }
+
+        public long WorkshopId { get; set; }
+
+        public virtual List<Workday> Workdays { get; set; }
```
VadymLevkovskyi, line 16: Suggestion:
Variant **0**. Leave as is
Variant **1**. instead of `public virtual List<Workday> Workdays { get; set; }`
use
```
public bool IsMonday {get; set;}
...
public bool IsSunday {get; set;}
```
and perform mapping to `List<DayOfWeek>` in `automapper` profile config

Variant **2**. instead of `public virtual List<Workday> Workdays { get; set; }`
use `public byte WorkDaysMask {get; set;}`
and perform mapping to `List<DayOfWeek>` in `automapper` profile config

Variant **3**. instead of `public virtual List<Workday> Workdays { get; set; }`
use `public byte WorkDaysMask {get; set;}`
and perform mapping to custom `DaysOfWeek` flag `Enum` in `automapper` profile config

OutOfSchool/OutOfSchool.DataAccess/Models/DateTimeRange.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public long Id { get; set; }
+
+        public TimeSpan StartTime { get; set; }
+
+        public TimeSpan EndTime { get; set; }
+
+        public long WorkshopId { get; set; }
+
+        public List<DaysBitMask> Workdays { get; set; }
```
VadymLevkovskyi, line 17: Why list?

Bogdan-Hasanov: It can be serialized and deserialized using default Newtonsoft list converter

OutOfSchool/OutOfSchool.IdentityServer/Data/Migrations/OutOfSchoolMigrations/OutOfSchoolDbContextModelSnapshot.cs
```diff
@@ -427,6 +427,34 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+                        .HasColumnType("bigint")
+                        .UseIdentityColumn();
+
+                    b.Property<TimeSpan>("EndTime")
+                        .HasColumnType("time");
+
+                    b.Property<TimeSpan>("StartTime")
+                        .HasColumnType("time");
+
+                    b.Property<byte?>("Workdays")
```
VadymLevkovskyi, line 443: Why nullable?

OutOfSchool/OutOfSchool.WebApi/Models/DateTimeRangeDto.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public class DateTimeRangeDto
+    {
+        public long Id { get; set; }
+
+        public TimeSpan StartTime { get; set; }
+
+        public TimeSpan EndTime { get; set; }
+
+        public List<DaysBitMask> Workdays { get; set; }
```
VadymLevkovskyi, line 15: Shouldn't it be `List<DayOfWeek>`?

OutOfSchool/OutOfSchool.DataAccess/Models/DateTimeRange.cs
VadymLevkovskyi: Resolved after discussion that code changes requires significant amount of extra code and extra dependency added

OutOfSchool/OutOfSchool.WebApi/Models/DateTimeRangeDto.cs
VadymLevkovskyi: Same reason of resolution - extra code/dependency needed to make it good.

VadymLevkovskyi (approved):

OutOfSchool/OutOfSchool.DataAccess/IUnitOfWork.cs
```diff
@@ -0,0 +1,9 @@
+using System.Threading.Tasks;
+
+namespace OutOfSchool.Services
+{
+    public interface IUnitOfWork
+    {
+        public Task CompleteAsync();
```
DmyMi, line 7: just for consistency maybe it's a good idea to have a sync version too

OutOfSchool/OutOfSchool.WebApi/Startup.cs
```diff
@@ -110,7 +111,7 @@ public void ConfigureServices(IServiceCollection services)
                         .AllowAnyHeader()
                         .AllowCredentials()));
 
-            services.AddControllers();
+            services.AddControllers().AddNewtonsoftJson();
```
DmyMi, line 114: We have two attributes [here](https://github.com/ita-social-projects/OoS-Backend/blob/develop/OutOfSchool/OutOfSchool.WebApi/Models/ApplicationDTO.cs) and [here](https://github.com/ita-social-projects/OoS-Backend/blob/develop/OutOfSchool/OutOfSchool.DataAccess/Enums/ApplicationStatus.cs) that depend on System.Text.Json. Probably need to refactor them to use Newtonsoft variation

DmyMi (approved):

PR #350, opened 2021-09-26
Dzhus/endpoints to upload the picture

VyacheslavDzhus:


OutOfSchool/OutOfSchool.WebApi/Common/Utilities/PictureManager.cs
```diff
@@ -0,0 +1,99 @@
...
+using System.Drawing.Drawing2D;
+using System.Drawing.Imaging;
+using System.IO;
+using System.Linq;
+
+namespace OutOfSchool.WebApi.Common.Utilities
+{
+    public class PictureManager
+    {
+        private const long Quality = 75;
```
DmyMi, line 11: maybe extract to options/appsettings too? and it's better to have `90`% quality. As for HD jpeg with 75% we'll be saving only around 50 KB space, but with substantial loss of quality compared to 90%.

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/PictureController.cs
```diff
@@ -0,0 +1,323 @@
...
+        [Route("provider/{providerId}/picture/{pictureId}")]
+        public async Task<IActionResult> Provider(long providerId, Guid pictureId)
+        {
+            try
+            {
+                var pictureData = await pictureService.GetPictureProvider(providerId, pictureId).ConfigureAwait(false);
+
+                return pictureData.ToFileResult();
+            }
+            catch (EntityNotFoundException)
```
VadymLevkovskyi, line 91: Regarding exception handling - see https://itamoedreamteam.slack.com/archives/C01HGKDMSFQ/p1632908126132700 seems you might simplify your `try/catch`s in this Controller or remove them at all.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/PictureController.cs
```diff
@@ -0,0 +1,323 @@
...
+        private string GetFormat(string fileName)
+        {
+            string contentType;
+
+            if (!contentTypeProvider.TryGetContentType(fileName, out contentType))
+            {
+                return string.Empty;
+            }
+
+            return contentType == string.Intern(PictureTypeNames.Jpeg) || contentType == string.Intern(PictureTypeNames.Png) ? contentType : string.Empty;
```
VadymLevkovskyi, line 320: There is no any need to use `string.Intern()` here

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/PictureController.cs
```diff
@@ -0,0 +1,323 @@
...
+        private bool IsExtensionValid(IFormFile picture)
+        {
+            var contentType = GetFormat(picture.FileName);
+
+            return string.IsNullOrEmpty(contentType) ? false : true;
+        }
+
+        private string GetFormat(string fileName)
+        {
+            string contentType;
```
VadymLevkovskyi, line 313: Variable (`contentType`) declaration should be as closer to it's 1st usage as possible, ideally in same line (declaration + initialization, like `var something = specificValue;`)

It's preferred to use `var` in variable declaration to enforce beforementioned rule

PR #386, opened 2021-10-08
Add logging for IdentityServer

vladbond007:
#86


vladbond007:
I have to get new develop changes before

Issue #25, opened 2021-02-02
Registration of more children 


Koshelenko-Tamara:
**As a** user **I want to** add more children **so that** they could become registered as well. 

**Description**
Adding more children in the process of registration of children. 
Predecessor:  US https://github.com/ita-social-projects/OoS-Backend/issues/20

**Assumptions** 
- If the customer presses the button "X" in case the child's data has already been filled in, the pop-up window "Are you sure you want to close the section?" appears.

**Business rules:**
 A customer can add up to 20 children.
 A customer can correct his and his child’s information in his personal account.  

Validation: 

Name | Type | Condition | Validation | Error message
-- | -- | -- | -- | --
Ім"я | Input field. Enabled | Mandatory | Min. 1 max. 30 characters (alphabetic and special characters ' - ) | Numeric characters and special characters (%$#@&*?!+=:"<>/) are not allowed
Прізвище | Input field. Enabled | Mandatory | Min. 1 max. 30 characters (alphabetic and special characters ' - ) | Numeric characters and special characters (%$#@&*?!+=:"<>/) are not allowed
По-батькові | Input field. Enabled | Mandatory | Min. 1 max. 30 characters (alphabetic and special characters ' - ) | Numeric characters and special characters (%$#@&*?!+=:"<>/) are not allowed
Згода на обробку даних | Check-box. Enabled | Mandatory

**Acceptance Criteria**
1. In case the customer wants to register one more child, he presses on the button "ADD MORE".

<img width="303" alt="Screenshot 2021-02-22 at 23 12 51" src="https://user-images.githubusercontent.com/77809312/108771463-5f3dd800-7564-11eb-8d10-0e4d1985ea24.png">

2. After that the new child's info section appears representing the information: surname, first name, patronymic, sex (male/female) and date of birth. 
3. The customer must choose the type of the social group the child belongs to. The groups are represented as a dropdown component with the following items: 1) "Діти із багатодітних сімей"; 2)"Діти із малозабезпечених сімей"; 3)"Діти з інвалідністю"; 4) "Діти-сироти"; 5) "Діти, позбавлені батьківського піклування".

<img width="245" alt="Screenshot 2021-02-22 at 23 16 40" src="https://user-images.githubusercontent.com/77809312/108771654-a62bcd80-7564-11eb-8f80-57f4d6c0ff18.png">

4. All the next children should be registered by the same principle, using the button "ADD MORE". 

[...]

Issue #20, opened 2021-02-01
Children registration 


Koshelenko-Tamara:
**As a** user (registered parent) **I want to** enter the child's data **so that** he will be registered as well.

**Description**
Predecessor: US  https://github.com/ita-social-projects/OoS-Backend/issues/13
Successor: US  https://github.com/ita-social-projects/OoS-Backend/issues/25
Mockup: https://ninjamock.com/s/QJ1HFHx

**Precondition:** A customer has already filled in his personal data and pressed on the button "REGISTER". 

**Assumptions:** 
- If the customer doesn't want to register a child, he must press the button "X" in the top right corner of the child's registration section. 

**Business rules:** Choosing a type of a social group must be available for multiple selection (more than one selection is available).
 
### Acceptance Criteria
1. When a customer (parent role) chooses the button "REGISTER A CHILD" a pop-up window appears. The information to be filled in about the child is: surname, first name, patronymic (father's name), sex (male/female) and date of birth. The date of birth must contain: date, month and year. 
 

<img width="300" alt="Screenshot 2021-02-17 at 00 52 57" src="https://user-images.githubusercontent.com/77809312/108135425-be14d480-70c0-11eb-982c-f29d58cea043.png">
2. The customer must choose the type of the social group the child belongs to. The groups are represented as a dropdown component with the following items: 1) "Діти із багатодітних сімей"; 2)"Діти із малозабезпечених сімей"; 3)"Діти з інвалідністю"; 4) "Діти-сироти"; 5) "Діти, позбавлені батьківського піклування".

3. The checkbox is used for customer confirmation of the possibility of processing his child’s personal data. Checkbox’s text is: “Даю згоду на обробку персональних даних.»  

4.  After pressing the button “COMPLETE” the pop-up window appears which says: “The letter with the following details has been sent to your email ...(customer’s email). The letter is sent to the mail, where the customer needs to confirm the registration. 

<img width="300" alt="Screenshot 2021-02-17 at 00 52 57 copy" src="https://user-images.githubusercontent.com/77809312/108135625-18ae3080-70c1-11eb-8426-3f0e02c9522e.png">
5. The customer follows the instructions in the email and complete the registration process. 

**Labels to be added**
"User story"

### Tasks 
1. - [ ] sample task.

Issue #13, opened 2021-01-25
Parent registration 


Koshelenko-Tamara:
**As a user** (Parent Role) I **want to be** able to provide my personal information **so that** I could be registered in the application and use it afterwards.
**Description**
Mockup:
https://ninjamock.com/s/QJ1HFHx

**Preconditions:**
 -  A user has already started registration by tapping on the "Register as a user" button. 

Validation 

Name | Type | Condition | Validation | Error message | Comments
-- | -- | -- | -- | -- | --
Ім'я | Input field. Enabled | Mandatory | Min. 1 max. 30 characters (alphabetic and special characters ' - ) | Numeric characters and special characters (%$#@&*?!+=:"<>/) are not allowed |  
Прізвище | Input field. Enabled | Mandatory | Min. 1 max. 30 characters (alphabetic and special characters ' - ) | Numeric characters and special characters (%$#@&*?!+=:"<>/) are not allowed |  
По-батькові | Input field. Enabled | Mandatory | Min. 1 max. 30 characters (alphabetic and special characters ' - ) | Numeric characters and special characters (%$#@&*?!+=:"<>/) are not allowed |  
Email | Input field. Enabled | Mandatory | Check the existence of @ character |   |  
телефон | Input field. Enabled | Mandatory | User should enter 9 numeric characters |   | autofil +380
Згода на обробку даних | Check-box. Enabled | Mandatory |   |   |  
Мені є 18 років | Check-box. Enabled | Mandatory |   |   |  
Зареєструватися | Button. Disabled |   |   |   | Enabled when all data is provided
Зареєструвати дитину | Button. Disabled |   |   |   | Enabled when all data is provided

Name | Type | Condition | Validation | Error message | Comments
-- | -- | -- | -- | -- | --
Пароль | Input field. Enabled | Mandatory | Min. 8 characters (alphabetic, numeric and special characters ' - ) |   | Enabled when all data is provided

### Acceptance Criteria
1. After tapping on "Register as a user", a visitor is redirected to the pop-up window containing further obligatory information for filling: surname, name, patronymic, email, phone number, password and password confirmation.
<img width="216" alt="Screenshot 2021-02-16 at 01 00 22" src="https://user-images.githubusercontent.com/77809312/108000420-99eac200-6ff2-11eb-9613-4bb7ff76ecd2.png">

2. After the password confirmation field a checkbox must be placed. The checkbox is used for customer confirmation of the possibility to process his personal data. 

[...]

PR #399, opened 2021-10-10
Provicevko/fix bug add child

provicevko:
#369 
Changed lengths for fields: FirstName, LastName, MiddleName --> [1-30] in ChildDTO
Changed lengths of Child entity for fields: FirstName, LastName, MiddleName --> [1-30]

#367 
Changed validation for ChildDTO

Based on:
#20
#13
#25

### DIFF
--- a/OutOfSchool/OutOfSchool.DataAccess/Models/Child.cs
+++ b/OutOfSchool/OutOfSchool.DataAccess/Models/Child.cs
@@ -9,18 +9,18 @@ namespace OutOfSchool.Services.Models
     {
         public long Id { get; set; }
 
-        [Required(ErrorMessage = "First name is required")]
-        [MaxLength(40)]
-        [RegularExpression(@"[\w\-\']*", ErrorMessage = "First name cannot contains digits and special symbols")]
+        [Required]
+        [MinLength(1)]
+        [MaxLength(30)]
         public string FirstName { get; set; } = string.Empty;
 
-        [Required(ErrorMessage = "Last name is required")]
-        [MaxLength(40)]
-        [RegularExpression(@"[\w\-\']*", ErrorMessage = "Last name cannot contains digits and special symbols")]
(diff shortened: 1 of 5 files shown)


OutOfSchool/OutOfSchool.DataAccess/Models/Child.cs
```diff
@@ -9,18 +9,18 @@ public class Child
     {
         public long Id { get; set; }
 
-        [Required(ErrorMessage = "First name is required")]
-        [MaxLength(40)]
-        [RegularExpression(@"[\w\-\']*", ErrorMessage = "First name cannot contains digits and special symbols")]
+        [Required]
+        [MinLength(1)]
+        [MaxLength(30)]
```
DmyMi, line 14: Shouldn't we have additional validation in data access if somehow someone manages create/modify child without going through child controller path (and childdto validation)? i.e., some other custom logic that will somehow modify child data.
p.s. question for discussion, not immediate action.

OutOfSchool/OutOfSchool.WebApi/Models/ChildDTO.cs
```diff
@@ -10,17 +10,17 @@ public class ChildDto
         public long Id { get; set; }
 
         [Required(ErrorMessage = "First name is required")]
-        [MaxLength(40)]
-        [RegularExpression(@"[\w\-\']*", ErrorMessage = "First name cannot contains digits")]
+        [StringLength(30, MinimumLength = 1)]
+        [RegularExpression(@"^[А-ЯҐЄІЇ](([\'\-][А-ЯҐЄІЇа-яґєії])?[а-яґєії]*)*$", ErrorMessage = "First name contains invalid characters")]
```
DmyMi, line 14: Do we really need to enforce first capital letter?
p.s. question for discussion, not immediate action.

PR #408, opened 2021-10-13
Necheporenko/direction_fix

BHy4uk:
Changed WorkshopCard`s property Direction to DirectionId. Fixed MappingProfile


OutOfSchool/OutOfSchool.WebApi/Models/WorkshopCard.cs
```diff
@@ -37,7 +37,7 @@ public class WorkshopCard
         [Range(0, 10000, ErrorMessage = "Field value should be in a range from 1 to 10 000")]
         public decimal Price { get; set; } = default;
 
-        public string Direction { get; set; }
+        public string DirectionId { get; set; }
```
Bogdan-Hasanov, line 40: Also type should be long

DmyMi: +1 :)

yfedo (approved):

PR #420, opened 2021-10-31
tatokhin/authorization update from role-based to permission-based

toherberg:
Summary:
Hard-coded role-based authorization system (i. e. [Authorize(Roles = "admin, provider")]) altered with more flexible and easier to support permission-based (i.e [HasPermission(Permissions.WorkshopAddNew)] ) .

What was done to achieve this:
-Added new IdentityResource to generate additional claim "Permissions" and send it to client via token
-Created tools for permissions-based authorization - detailed enum with grouped permissions for all currently available actions in our application, permission packers, custom attribute, policy provider, permission checker etc
-Updated db with new table "PermissionsForRole" to store data about each role permissions
-Implemented interface, service and api-controller for system-admin to manage permissions for user roles in our application
-Changed attributes in all controllers to new custom permissions based.


OutOfSchool/OutOfSchool.Common/PermissionModule/PermissionManagement/Permissions.cs
```diff
@@ -0,0 +1,130 @@
...
+        [Display(GroupName = "User", Name = "Add new", Description = "Can add a new User item")]
+        UserAddNew = 82,
+        [Display(GroupName = "User", Name = "Remove", Description = "Can remove User data")]
+        UserRemove = 83,
+        #endregion
+
+        #region Workshop Control permissions #10
+        [Display(GroupName = "Workshop", Name = "Read", Description = "Can read Workshop")]
+        WorkshopRead = 90,
+        [Display(GroupName = "Workshop", Name = "Add new", Description = "Can edit a Workshop item")]
```
provicevko, line 114: Name  = "Edit"?

OutOfSchool/OutOfSchool.Common/PermissionModule/PermissionManagement/PermissionChecker.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.ComponentModel;
+using System.Linq;
+using OutOfSchool.Common.PermissionsModule;
+
+namespace OutOfSchool.Common.PermissionsModule
+{
+    public static class PermissionChecker
+    {
+        /// <summary>
+        /// This is used by the policy provider to check the permission name string
```
DmyMi, line 12: either fix comments (args, return types, throws, etc.) or remove it

OutOfSchool/OutOfSchool.Common/PermissionModule/PermissionManagement/Permissions.cs
```diff
@@ -0,0 +1,130 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace OutOfSchool.Common.PermissionsModule
+{
+    public enum Permissions : short
```
DmyMi, line 6: Just a question for discussion: separate groups for permissions (have separate enums for each group). That way enums gonna be smaller and we can use flag enums for fast bitwise permissions checks within a group.

OutOfSchool/OutOfSchool.IdentityServer/ProfileService.cs
```diff
@@ -36,5 +45,12 @@ public async Task IsActiveAsync(IsActiveContext context)
             var user = await userManager.FindByIdAsync(sub);
             context.IsActive = user != null;
         }
+
+        // get's list of permissions for current user's role from db
+        private async Task<string> GetPermissionsForRole(string roleName)
+        {
+            var permissionsForRole = (await repository.GetByFilter(p => p.RoleName == roleName)).FirstOrDefault();
```
provicevko, line 52: Just use First()

OutOfSchool/OutOfSchool.WebApi/Services/PermissionsForRoleService.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            return permissionsForRole.FirstOrDefault().ToModel();
+        }
+
+        /// <inheritdoc/>
+        public async Task<PermissionsForRoleDTO> Create(PermissionsForRoleDTO dto)
+        {
+            logger.LogInformation($"Permissions for Role - {dto?.RoleName} creating was started.");
+
+            var allPermissions = await repository.GetAll().ConfigureAwait(false);
```
provicevko, line 71: Should we check dto for null not to make a request in this case? You'll get an exception after it. And do we need to return allPermisions into our memory, you can check it in a DB.

OutOfSchool/OutOfSchool.Common/PermissionModule/PermissionManagement/Permissions.cs
provicevko, line 114: 114. Name = "Edit"

PR #436, opened 2021-11-16
Added workshop`s rating for short model called by StatisticController

BHy4uk:


fkma13552 (approved):

OutOfSchool/OutOfSchool.WebApi.Tests/Services/StatisticServiceTest.cs
```diff
@@ -22,24 +23,30 @@ public class StatisticServiceTest
 
         private Mock<IApplicationRepository> applicationRepository;
         private Mock<IWorkshopRepository> workshopRepository;
+        private Mock<IRatingService> ratingService;
```
VadymLevkovskyi, line 26: `ratingService` seems to be not ever used outside `SetUp` method. Thus might be moved inside it as local var

VadymLevkovskyi: Same for `logger`

VadymLevkovskyi: Same for `mapper`

VadymLevkovskyi (approved):
There small issues with tests, might be fixed in other PRs

DmyMi (approved):

toherberg (approved):

PR #440, opened 2021-11-22
Provicevko/ Upload/Get workshop images

provicevko:
The first realization of image services


DmyMi (approved):

OutOfSchool/OutOfSchool.DataAccess/Repository/ExternalImageStorage.cs
```diff
@@ -0,0 +1,60 @@
...
+                throw new ImageStorageException(ex);
+            }
+        }
+
+        public async Task<string> UploadImageAsync(Stream contentStream, CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                contentStream.Position = uint.MinValue;
+                var objectId = await gridFsBucket.UploadFromStreamAsync(Guid.NewGuid().ToString(), contentStream, cancellationToken: cancellationToken).ConfigureAwait(false);
```
DmyMi, line 41: probably not relevant if we move to SQL storage, but maybe leave a comment so we don't forget. If we stick to Mongo then using objectId might not be cool as it's technically not an uuid and they are predictable. We can use the file name as identifier (and it's possible to search by filename) as it's almost unique :)

DmyMi, line 41: not relevant if we move to MySQL storage, but just a note (maybe we return to Mongo). Maybe leave a comment here.
Mongo default object ids are not unique and can be predicted (not as easy as integer but still). We can search by file name and use it as an identifier as uuid is pretty unique :)

OutOfSchool/OutOfSchool.WebApi/Services/Images/ImageService.cs
```diff
@@ -0,0 +1,263 @@
...
+            this.imageRepository = imageRepository;
+            this.serviceProvider = serviceProvider;
+            this.logger = logger;
+        }
+
+        /// <inheritdoc/>
+        public async Task<Result<ImageStorageModel>> GetByIdAsync(Guid imageId)
+        {
+            logger.LogInformation($"Getting image by id = {imageId} was started.");
+            var imageMetadata = await imageRepository.GetMetadataById(imageId).ConfigureAwait(false);
```
DmyMi, line 46: probably leave a comment here too if we stick with Mongo.
Two step image retrieval (get info from sql and then image from mongo) will be a very resource intensive operation. As this method is used in public endpoint - we can use just image id and go to mongo directly.

DmyMi (approved):

vladbond007 (approved):

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopController.cs
```diff
@@ -170,10 +184,24 @@ public async Task<IActionResult> Create(WorkshopDTO dto)
 
             var workshop = await combinedWorkshopService.Create(dto).ConfigureAwait(false);
 
+            IDictionary<short, OperationResult> imageBadResults = new Dictionary<short, OperationResult>();
+            if (dto.ImageFiles != null && dto.ImageFiles.Count > 0)
+            {
+                if (!ValidCountOfFiles(dto.ImageFiles.Count))
+                {
```
yfedo, line 191: maybe it's better to count size of all files in bytes?)

OutOfSchool/OutOfSchool.WebApi/Services/Images/IImageService.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <summary>
+        /// Uploads images for the chosen workshop and then updates it.
+        /// </summary>
+        /// <param name="workshopId">It's an Id of Workshop.</param>
+        /// <param name="fileCollection">Contains images to upload.</param>
+        /// <returns>The instance of <see cref="Dictionary{TKey,TValue}"/> that contains results (type of <see cref="OperationResult"/>) for uploading any image.
+        /// This is the pair of keys from fileCollection and appropriate OperationResult.
+        /// It returns the pair of (-1, <see cref="OperationResult"/>) if there are no ways to update the current workshop.</returns>
+        Task<IDictionary<short, OperationResult>> UploadManyWorkshopImagesWithUpdatingEntityAsync(
+            Guid workshopId,
```
yfedo, line 33: can -1 status code be part of OperationResult?)

OutOfSchool/OutOfSchool.WebApi/Services/Images/ImageValidatorService.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                logger.LogError($"Unable to validate stream {ex.Message}");
+                return OperationResult.Failed(new OperationError {Code = ImageResourceCodes.UnexpectedValidationError, Description = Resources.ImageResource.UnexpectedValidationError});
+            }
+        }
+
+        /// <inheritdoc/>
+        public bool ValidateImageSize(long size)
+        {
+            return size <= options.MaxSizeBytes;
```
yfedo, line 71: nice))

yfedo (approved):

PR #445, opened 2021-11-26
Necheporenko/Add a rejection message

BHy4uk:
Added a rejection message after application status is changed to rejected


OutOfSchool/OutOfSchool.DataAccess/Repository/ApplicationRepository.cs
```diff
@@ -50,6 +51,13 @@ public new async Task<Application> Update(Application entity)
 
             dbContext.Entry(application).State = EntityState.Modified;
 
+            if (application.Status != ApplicationStatus.Rejected)
```
DmyMi, line 54: repository shouldn't contain logic. Need to check the status in controller or service

DmyMi (changes requested):

DmyMi (approved):

toherberg (approved):

SergeyNovitsky (approved):

yfedo (approved):

PR #447, opened 2021-11-28
Tatokhin/Unit tests + refactored controllers and services(Provider, InstitutionStatus, PermissionsForRole) 

toherberg:
-Fixed existing and created additional unit tests to examine the logic of controllers and services, uppermost, Providers, and then also for InstitutionStatus and PermissionsForRole objects in our app. To achieve this, created the system with helper methods as data generators, randomizers, etc., following the example of similar features for other objects. Also produced generic methods for easy assertions of expected and actual results in situations, where we have to examine multiple conditions (e.g. - for controllers - check response type and also value inside it, both if it is a single object, collection or even data retrieved from bad request.

-Removed user Role from parameters of IProviderService Update method. Deleted check for the role during method execution because currently, our service allows Users with the "Admin" role to update any entity from the Providers list. My changes caused that provider entities can be edited only by the user-owner.

-Altered ProviderController actions to handle exceptions thrown during service usage and generate BadRequest response with a documented specific problem. Deleted code for getting current user's role, because we don't check role anymore. So the current flow is - we need specific "ProviderEdit" permission to use controller's "Update" action and also logged user's Id must be similar to UserId of Provider entity we want to change.

-------------------

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/InstitutionStatusControllerTests.cs
+++ b/OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/InstitutionStatusControllerTests.cs
@@ -7,8 +7,10 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Localization;
 using Moq;
 using NUnit.Framework;
+using OutOfSchool.Services.Models;
 using OutOfSchool.Tests.Common;
 using OutOfSchool.WebApi.Controllers.V1;
+using OutOfSchool.WebApi.Extensions;
 using OutOfSchool.WebApi.Models;
 using OutOfSchool.WebApi.Services;
 
@@ -21,8 +23,8 @@ namespace OutOfSchool.WebApi.Tests.Controllers
         private Mock<IStatusService> service;
         private Mock<IStringLocalizer<SharedResource>> localizer;
 
-        private IEnumerable<InstitutionStatusDTO> institutionStatuses;
-        private InstitutionStatusDTO institutionStatus;
+        private IEnumerable<InstitutionStatus> institutionStatuses;
+        private InstitutionStatus institutionStatus;
 
 
 
@@ -42,13 +44,14 @@ namespace OutOfSchool.WebApi.Tests.Controllers
         public async Task GetInstitutionStatuses_WhenCalled_ReturnsOkResultObject()
         {
             // Arrange
-            service.Setup(x => x.GetAll()).ReturnsAsync(institutionStatuses);
+            var expected = institutionStatuses.Select(x => x.ToModel());
+            service.Setup(x => x.GetAll()).ReturnsAsync(institutionStatuses.Select(x => x.ToModel()));
 
             // Act
             var response = await controller.Get().ConfigureAwait(false);
 
             // Assert
-            response.GetAssertedResponseOkAndValidValue<IEnumerable<InstitutionStatusDTO>>();
+            response.AssertResponseOkResultAndValidateValue(expected);
         }
 
         [Test]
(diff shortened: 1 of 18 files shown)


OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ProviderControllerTests.cs
```diff
@@ -1,519 +1,290 @@
...
+        {
+            // Arrange
+            providerService.Setup(x => x.GetAll()).ReturnsAsync(providerDtos);
+
+
+            // Act
+            var result = (await providerController.Get().ConfigureAwait(false) as ObjectResult).Value as IEnumerable<ProviderDto>;
+
+            // Assert
+            Assert.That(providerDtos.Union(result).Count, Is.EqualTo(providerDtos.Count()));
```
h4wk13, line 111: What does count exactly says here ? 
We should check all entries for equality.
Not sure if order is matter.

toherberg: Union method returns IEnumerable based on two other collections members, ignoring duplicates. Due to that logic, as we have same collection length in result Value as expected collection count is, my idea was that union of two totally similar collections produces third exactly same collection)

h4wk13: ok, does that mean that 2 collections with the same number of entries are equals ?

DmyMi (approved):

BHy4uk (approved):

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/PermissionsForRoleControllerTests.cs
```diff
@@ -65,104 +71,138 @@ public async Task GetPermissionsForRoles_WhenEmptyCollection_ReturnsNoContentRes
...
         }
 
+
         [Test]
         [TestCase("Admin")]
         public async Task GetByRoleName_WhenRoleNameIsValid_ReturnOkResultObject(string roleName)
         {
             // Arrange
-            service.Setup(x => x.GetByRole(roleName)).ReturnsAsync(permissionsForAllRoles.SingleOrDefault(x => x.RoleName == roleName));
+            var expected = permissionsForAllRoles.Where(s => s.RoleName == roleName).Select(p => p.ToModel()).FirstOrDefault();
```
h4wk13, line 92: `First()` ?

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/PermissionsForRoleControllerTests.cs
```diff
@@ -65,104 +71,138 @@ public async Task GetPermissionsForRoles_WhenEmptyCollection_ReturnsNoContentRes
...
 
             // Act
             var response = await controller.GetByRoleName(roleName).ConfigureAwait(false);
 
             // Assert
-            response.GetAssertedResponseOkAndValidValue<PermissionsForRoleDTO>();
+            response.AssertResponseOkResultAndValidateValue(expected);
         }
 
         [Test]
```
h4wk13, line 102: please remove

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/PermissionsForRoleControllerTests.cs
```diff
@@ -65,104 +71,138 @@ public async Task GetPermissionsForRoles_WhenEmptyCollection_ReturnsNoContentRes
...
         [TestCase("fakeRole")]
```
h4wk13, line 103: please use random string instead

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/InstitutionStatusControllerTests.cs
```diff
@@ -152,60 +162,62 @@ public async Task DeleteInstitutionStatus_WhenIdIsValid_ReturnsNoContentResult(l
...
+            return new InstitutionStatus()
             {
                 Id = 1,
-                Name = "Test",
+                Name = TestDataHelper.GetRandomWords(),
             };
         }
 
-        private IEnumerable<InstitutionStatusDTO> FakeInstitutionStatuses()
+        private IEnumerable<InstitutionStatus> FakeInstitutionStatuses()
```
h4wk13, line 203: why it is still here ?

PR #472, opened 2021-12-22
Refactor/google logging

DmyMi:
* Refactored logging config
* Fixed building container with `libgdiplus`

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Services/Images/ImageService.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Services/Images/ImageService.cs
@@ -125,7 +125,7 @@ namespace OutOfSchool.WebApi.Services.Images
             }
             catch (Exception ex)
             {
-                logger.LogError($"Exception while uploading images for workshopId = {workshopId}: {ex.Message}");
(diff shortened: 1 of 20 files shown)


OutOfSchool/OutOfSchool.WebApi/Services/Images/ImageService.cs
```diff
@@ -106,7 +106,7 @@ public async Task<Result<ImageDto>> GetByIdAsync(string imageId)
                     var validationResult = validator.Validate(stream);
                     if (!validationResult.Succeeded)
                     {
-                        logger.LogError($"Image with {nameof(fileCollection)} id = {i} isn't valid.");
+                        logger.LogError($"Image with {nameof(fileCollection)} id = {i} isn't valid: {string.Join(",", validationResult.Errors)}");
```
provicevko, line 109: we will return only the first error if not succeeded in this case

provicevko (approved):

OutOfSchool/OutOfSchool.WebApi/Services/Images/ImageService.cs
DmyMi: join is more general then `array[0]` :)

PR #484, opened 2022-01-05
Update workshop images

provicevko:
Changed Image services structures.
Added changing images while updating workshop.
Fixed xml docs.

Changed resources logic


OutOfSchool/OutOfSchool.WebApi/Common/Resources/Describers/ImagesErrorDescriber.cs
```diff
@@ -0,0 +1,124 @@
+﻿using Microsoft.AspNetCore.Identity;
+
+namespace OutOfSchool.WebApi.Common.Resources.Describers
+{
+    public class ImagesErrorDescriber
```
DmyMi, line 5: maybe use a factory pattern with a method that receives error type and returns operationerror object?

OutOfSchool/OutOfSchool.WebApi/Controllers/V2/WorkshopController.cs
```diff
@@ -164,18 +163,22 @@ public async Task<IActionResult> GetByFilter([FromQuery] WorkshopFilter filter)
         [Consumes("multipart/form-data")]
         public async Task<IActionResult> Create([FromForm] WorkshopCreationDto dto) // TODO: validate by request size
         {
-            // TODO: Also should check for AddressDto == null
             if (!ModelState.IsValid)
             {
                 return BadRequest(ModelState);
             }
 
+            if (dto.Address == null)
```
DmyMi, line 171: Maybe add a `required` attribute to address field in model and validate it during model validation?
or it doesn't work with form data?

OutOfSchool/OutOfSchool.WebApi/Controllers/V2/WorkshopController.cs
```diff
@@ -233,28 +221,44 @@ public async Task<IActionResult> GetByFilter([FromQuery] WorkshopFilter filter)
         /// <response code="400">If the model is invalid, some properties are not set etc.</response>
         /// <response code="401">If the user is not authorized.</response>
         /// <response code="403">If the user has no rights to use this method, or sets some properties that are forbidden to change.</response>
+        /// <response code="413">If the request break the limits, set in configs.</response>
         /// <response code="500">If any server error occures.</response>
         [HasPermission(Permissions.WorkshopEdit)]
-        [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(WorkshopDTO))]
+        [ProducesResponseType(StatusCodes.Status200OK)]
```
DmyMi, line 227: better to leave the type in produces response. If you are using anonymous object then create a type for it.

OutOfSchool/OutOfSchool.WebApi/appsettings.Development.json
```diff
@@ -41,22 +41,23 @@
...
-      "MaxWidthPixels": 10000,
-      "MinHeightPixels": 250,
-      "MaxHeightPixels": 8000,
-      "MaxSizeBytes": 16777216,
-      "MaxWidthHeightRatio": 2, 
-      "SupportedFormats": [
-        "jpeg",
-        "png"
-      ]
+      "Specs": {
```
DmyMi, line 46: if the config is the same as in basic `appsettings.json` you don't really need to repeat this in other settings files.

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Common/Resources/Describers/ImagesErrorDescriber.cs
provicevko: Sounds really cool but in this way we won't be able to build messages dynamically (I mean calling methods with some parameters and then use string formatting/interpolation to create something new, I've used it in ExceedingCountOfImagesError(int countOfImages)). Of course in this case we could create a little bit more general message and delete parameters. So we should discuss it :)

OutOfSchool/OutOfSchool.WebApi/Controllers/V2/WorkshopController.cs
provicevko: I hadn't been sure that we use dto.Address in all situations as required before your wrote this. I will change it on attribute)

OutOfSchool/OutOfSchool.WebApi/Controllers/V2/WorkshopController.cs
provicevko: I agree with you that we should use the concrete type, but for this I will create new classes only for returning from controllers (or using in tests)

OutOfSchool/OutOfSchool.WebApi/Common/Resources/Describers/ImagesErrorDescriber.cs
yfedo:
```
public enum OperationError : byte
{
		[AppResourcesKey("txtDefaultError")]
		DefaultError = 0,

		[AppResourcesKey("txtUploadingError")]
		UploadingError = 1,
}
```

```
public class AppResourcesKeyAttribute : Attribute
{
		#region Constructor

		/// <summary>
		/// Initializes a new instance of the <see cref="AppResourcesKeyAttribute"/> class.
		/// </summary>
		/// <param name="appResourcesKey">The application resources key.</param>
		public AppResourcesKeyAttribute(string appResourcesKey) => AppResourcesKey = appResourcesKey;

		#endregion

		#region Public properties

		/// <summary>
		/// Gets or sets the application resources key.
		/// </summary>
		public string AppResourcesKey { get; set; }

		#endregion
}
```
```
/// <summary>
		/// Returns text from AppResources by AppResourceKey
		/// </summary>
		/// <param name="en">target enum</param>
		/// <returns>AppResources by AppResourceKey</returns>
		public static string GetAppResource(this Enum en)
		{
			var appResourceKeyAttribute = en.GetAttribute<AppResourcesKeyAttribute>();

			return appResourceKeyAttribute == null
				? string.Empty
				: AppResources.ResourceManager.GetString(appResourceKeyAttribute.AppResourcesKey);
		}
```
we can do something like this and use the factory method))
so we will be able to add new errors easily))

yfedo (approved):

PR #489, opened 2022-01-10
Fix tests

alrom:
Fix tests for Child and Parent controllers, Workshop controller and Statistic service


VadymLevkovskyi (approved):
Some magic constants are introduced (for new code it is treated as code smell), but this legacy approach seems to be spread over whole tests, so can't blame the particular developer

provicevko (approved):

PR #497, opened 2022-01-13
Got rid of the CreateMultiple applications method

BHy4uk:


OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -389,19 +389,7 @@ public async Task<IActionResult> Delete(Guid id)
...
-            {
-                ChildId = child.Id,
-                CreationTime = DateTimeOffset.UtcNow,
-                WorkshopId = applicationApiModel.WorkshopId,
-            });
-
-            return applications.ToList();
-        }
-
+
```
DmyMi, line 392: maybe remove not only the method but every other objects that were used like `ApplicationApiModel`? Unless it is used somewhere else

DmyMi (approved):

PR #501, opened 2022-01-15
Provicevko/change image service structure

provicevko:
-> Made images generic service
-> Added deleting images while deleting workshop

--  NOT ADDED distributed transactions


VadymLevkovskyi:
> If you have some propositions how to do it in another way, I'll be happy to see it :).

Correct me if I'm wrong you are now using inheritance - to put common code into base class and specialize behavior (mentioned `GetFilterForSearchingEntityByIdWithIncludedImages` and `GetEntityImages` methods) in descendants by overriding abstract ones. 

This might always be refactored to composition (see Strategy pattern) or even as simple as passing extra parameters (like `Action<>` or `Func<>` in a specific methods). While latter seems not feasible for your case, former sounds good for me.

Hope it will help )

DmyMi (approved):

provicevko:
> Correct me if I'm wrong you are now using inheritance - to put common code into base class and specialize behavior (mentioned `GetFilterForSearchingEntityByIdWithIncludedImages` and `GetEntityImages` methods) in descendants by overriding abstract ones.

Yes, you're exactly right that's I try to do it. Could you explain how I can use composition with strategy in my case? These 2 methods depends on generic parameters (TEntity, TKey). For this case I can only create something like ITypeStrategy<TEntity, TKey> with these methods but it doesn't have any sense because each pair of (TEntity, TKey) has to have 1 implementation. As a result I won't see difference between overriding (like template method) or with strategy with this interface. The generic parameters are need because we don't have one base class for entities with images.
Or maybe you can offer something how to change logic of these methods, like GetEntityImages<TEntity>(TEntity entity) => entity.(NameOfEmagesField) in order to remove dependencies of generic parameters.
Thanks a lot :)

VadymLevkovskyi:
> ...(skipped)... Could you explain how I can use composition with strategy in my case? These 2 methods depends on generic parameters (TEntity, TKey). For this case I can only create something like ITypeStrategy<TEntity, TKey> with these methods but it doesn't have any sense because each pair of (TEntity, TKey) has to have 1 implementation. As a result I won't see difference between overriding (like template method) or with strategy with this interface. 

Yes, each pair has to have 1 strategy implementation. Same situation as you have currently with `WorkshopImagesInteractionService` - specific implementation per specific pair. Yes, strategy pattern will add `1` extra interface and `n` specific pair interface implementations. In general, rule `composition over inheritance` proves it's correctness when code has to be tested/changed. 

Do you need it? I don't know, until you specify the goal of refactoring or, in other words, what problem are we trying to solve. And I'm afraid, but currently your goals and attacked problems are unclear (at least for me). 

_Subsequent considerations are not worth of implementing, they are just thoughts I have re your code and examples of things that doesn't let me see the whole picture/problems._

F.i. I don't see the goal of having `TRepository` as generic type parameter. For me it looks like it could be used 
`protected IEntityRepositoryBase<TKey, TEntity> Repository { get; }`
thus simplifying several generics. Probably I don't see real obstacles (as I didn't checked out your branch in my local VS with all dependencies and compiler support). Probably I don't see the developer's intention.

From other point I don't see the goal to have separate `TEntity, TKey` generic parameters while they goes in pair everywhere. As I got the domain - the `TKey` is inner thing of `TEntity`. I assume it's impossible to have 1 specific `TEntity` with several `TKeys`. Thus instead of 
`SomeGenericClass<TEntity, TKey> where TEntity : class, new()`
it might be 
`SomeGenericClass<TEntity, TKey> where TEntity : TKeyedEntity<TKey>`
or even 
`SomeGenericClass<TEntity> where TEntity : TKeyedEntity`

Definitely last variant requires way too big refactoring and probably inevitable due to  
- forcing specific `WorkshopDto` / `Workshop` to implement ` : TKeyedEntity<Guid>, TKeyedEntity` - first interface to set returning type for `GetKey()` method / `Id` property, last interface just marker one to simplify constraints where key is not necessary)
- changing lot of places in code where `TKey` is used (but the same pattern is everywhere - `TKey entityId` is taken and passed to either `GetFilterForSearchingEntityByIdWithIncludedImages` or `GetEntityWithIncludedImages` (that calls the former method under the hood).

_So you still know your problem better than me and could try to think outside the box re code and find better solution than I've suggested._

toherberg (approved):

SergeyNovitsky (approved):

Issue #490, opened 2022-01-11
Change status code if user exceeds the application limit


DmyMi:
If user sends more requests than the limit per week then we should send a [429 status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) with Retry-After header stating when he/she can retry to send application to the workshop

PR #504, opened 2022-01-18
#490 Change status code if user exceeds the application limit

alrom:


OutOfSchool/OutOfSchool.WebApi/Models/ModelWithAdditionalData.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace OutOfSchool.WebApi.Models
+{
+    public class ModelWithAdditionalData<TModel, TData>
+        where TModel : class
+    {
+        public string Description { get; set; } = string.Empty;
+
+        public TModel Model { get; set; } = null;
+
+        public TData AdditionalData { get; set; }
```
DmyMi, line 12: possible future improvement: have additional data as dictionary, so we can store multiple named additional values if needed

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Models/ModelWithAdditionalData.cs
alrom: Our point was, if we need a collection as AdditionalData - we set out result like "new ModelWithAdditionalData<ApplicationDto, Dictionary<string, int>>()". Or we miss something?

provicevko (approved):

PR #526, opened 2022-01-25
Added WorkshopStatus

BHy4uk:
- Added a check for the rights to change the status regarding the type of ownership
#519


OutOfSchool/OutOfSchool.DataAccess/Enums/WorkshopStatus.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace OutOfSchool.Services.Enums
+{
+    [JsonConverter(typeof(StringEnumConverter))]
+    public enum WorkshopStatus
+    {
+        Opened = 1,
```
toherberg, line 9: shouldn't be better to use "Open" in this case? I mean, that's value is to save current status e. g. 
‘Open’ is an adjective that means ‘not closed’.
‘Opened’ is the past tense of the verb ‘open’.

toherberg (approved):

provicevko (approved):

VadymLevkovskyi (approved):

PR #538, opened 2022-01-31
Provicevko/teacher images

provicevko:
Added teacher images
Added support v1 CRUD for workshop (controllers and services)


DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
```diff
@@ -220,6 +306,39 @@ public async Task Delete(Guid id)
...
+            try
+            {
+                await workshopRepository.Delete(entity).ConfigureAwait(false);
+                logger.LogInformation($"{nameof(Workshop)} with Id = {id} successfully deleted.");
+            }
+            catch (DbUpdateConcurrencyException ex)
+            {
+                logger.LogError(ex, $"Deleting {nameof(Workshop)} with Id = {id} failed.");
+                throw;
+            }
```
SergeyNovitsky, line 339: I think we need a transaction here or maybe there is a cascade deletion and will be enough to delete workshop.

OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs
```diff
@@ -120,5 +136,21 @@ public MappingProfile()
             #warning The next mapping is here to test UI Admin features. Will be removed or refactored
             CreateMap<ShortUserDto, AdminDto>();
         }
+
+        public static List<Teacher> WorkshopTeachersMapperFunction(List<Teacher> dtoTeachers, List<Teacher> dest)
```
SergeyNovitsky, line 140: maybe make this method private?

SergeyNovitsky (approved):

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
provicevko: Now it will fall down if unreal to delete workshop images -> any teacher (images + data) -> workshop data so we can have a situation when a workshop isn't removed but some images/teachers are deleted (the partial delete). In this way we won't have unreferenced data in our db (garbage). We will create a transaction for these cases in the closest PRs when move our images to another storage.

PR #553, opened 2022-02-10
Romanchuk Lukashov/Notification hub

alrom:
Functionality for notifications was added. Draft version for frontend.


VadymLevkovskyi (approved):

OutOfSchool/OutOfSchool.WebApi/Startup.cs
```diff
@@ -102,6 +102,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, IApiVers
             {
                 endpoints.MapControllers();
                 endpoints.MapHub<ChatWorkshopHub>("/chathub/workshop");
+                endpoints.MapHub<NotificationHub>("/notificationhub");
```
DmyMi, line 105: maybe if notifications are disabled - do not add hub?

Also, after POC we can do [multiple hubs on single websocket connection](https://docs.microsoft.com/en-us/aspnet/signalr/overview/guide-to-the-api/hubs-api-guide-server#multiple-hubs).

As our messages are fairly small (chat & notifications) we should not have any performance issues, but can support more users (1 open connection vs 2)

DmyMi (approved):

PR #565, opened 2022-02-16
Provicevko/upgrade to mysql images

provicevko:


OutOfSchool/OutOfSchool.WebApi/Util/Transactions/TransactionPoolManager.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// Commits all transaction that are currently in the pool.
+        /// </summary>
+        public void CommitAll()
+        {
+            foreach (var dbTransaction in transactions)
+            {
+                DbManagerHelper.CommitTransaction(dbTransaction);
+            }
+
+            transactions.Clear();
```
DmyMi, line 64: comment not directly tied to this line, but i haven't seen any of the transactions being disposed. Might lead to some connections not being closed properly.

PR #575, opened 2022-02-22
Tkachenko/direction filtering

fkma13552:
Added pagination and filtering by string to Direction, Class and Department

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Services/DirectionService.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Services/DirectionService.cs
@@ -104,6 +104,22 @@ namespace OutOfSchool.WebApi.Services
             return directions.Select(entity => entity.ToModel()).ToList();
         }
 
+        public async Task<IEnumerable<DirectionDto>> GetByFilter(OffsetFilter filter)
+        {
+            logger.LogInformation("Getting Directions by filter started.");
+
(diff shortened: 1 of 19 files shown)


OutOfSchool/OutOfSchool.WebApi/Services/DirectionService.cs
```diff
@@ -143,6 +193,36 @@ public async Task<DirectionDto> Update(DirectionDto dto)
             }
         }
 
+        public async Task<SearchResult<DirectionDto>> FilterByName(DirectionFilter filter)
```
DmyMi, line 196: it's mostly identical to `GetByFilter`, do we need two methods?:)

OutOfSchool/OutOfSchool.WebApi/Services/DirectionService.cs
```diff
@@ -22,6 +24,8 @@ public class DirectionService : IDirectionService
         private readonly IWorkshopRepository repositoryWorkshop;
         private readonly ILogger<DirectionService> logger;
         private readonly IStringLocalizer<SharedResource> localizer;
+        private readonly IMapper mapper;
+        private readonly string includePropertiesForGetByFilterWithName = $"";
```
DmyMi, line 28: this property not being used?

DmyMi (approved):

PR #598, opened 2022-05-02
Added checking statuses of previously sent applications

provicevko:


OutOfSchool/OutOfSchool.DataAccess/Enums/ApplicationStatus.cs
```diff
@@ -4,7 +4,8 @@
 namespace OutOfSchool.Services.Enums
 {
     // TODO: Swagger ignores this attribute on model property in webapi layer
-    // TODO: Ask Frontend if we can just make all enums as strings instead of magic numbers
+    // Notice that statuses less or equal StudyingForYears are used to block ability to create a new application in AllowedNewApplicationByChildStatus method
```
DmyMi, line 7: maybe either use binary flags and check those (will need to alter existing variables in test DB) or a simple method with switch inside the service. Comments are ok, but the alternative solutions will work even if the enum is changed

DmyMi (approved):

OutOfSchool/OutOfSchool.DataAccess/Enums/ApplicationStatus.cs
VadymLevkovskyi: Thanks Dmytro, this comment scratches my attention too.
I could suggest use extension method. Something like 
```
static class ApplicationStatusExtension 
{
    public static bool IsAllowedNewApplicationByChildStatus(this ApplicationStatus status)
        => status <= ApplicationStatus.StudyingForYears;
}
```
Thus, as benefits, you could 
- expose your business logic in a code, not a comment, 
- write related unit-test 
- have one place of changes if/when business logic is changed
- extract this check from `ApplicationService.AllowedNewApplicationByChildStatus` implementation
- and reuse it in case of necessity without code duplication.

Such approach has obvious faults:
- extra level of abstraction,
- small perf. loss,
- compiler doesn't prohibit developer from writing raw comparison (`<= ApplicationStatus.StudyingForYears`) in a code, so extra efforts should be spent on code review.

provicevko: Okey, firstly, thank you everyone for your answers :)
I've just checked your proposals.
I'm really confused about what solution to choose here.

Vadym, your offer looks good in performance but I'm not sure that somebody else will remember that method to use it as well.
But also we have problem that this check must be converted into SQL code made by EF Core that's why we cannot use custom methods in such form.
To fix it, I must return Expression<Func<Application, bool>> from this method, something like
```
static class ApplicationExtension 
{
    public static Expression<Func<Application, bool>> IsNewApplicationAllowedByChildStatus()
       a => a.Status <= ApplicationStatus.StudyingForYears;
}
```
I guess that is awkward because of the point I said at first (remember this method) and also this fix reduces
amount of potential usage, I mean, the situation we use returning `Expression<Func<Application, bool>>` is almost one and it's in the ApplicationService (call using LINQ to Entity).
On the other hand, I agree that one method of magic check is better than possible multiple if we write as I did in my code.
For this case I'm able to write some tests that will check this method and also fall down whether ApplicationStatus enum is changed but allowed and blocked statuses are forgotten to adapt in tests.

The another solution as mentioned Dmytro is making this check explict in order to guarantee stability. I mean, I can check everything like `a.Status <= ApplicationStatus.Pending && a.Status <= ApplicationStatus.Approved && ...` etc.
There will be 4 checks.
Futhermore, to improve consistency in all parts, I can create static array of "stopping application statuses":
```
static class ApplicationStatusExtension
{
    public static IReadOnlyList<ApplicationStatus> GetStoppingNewApplicationStatuses { get; } =
        new List<ApplicationStatus> { ApplicationStatus.Pending, ApplicationStatus.Approved }.AsReadOnly();
}
```
This must be a thread-safety collection of allowed statuses, thus, will be only 1 instance in memory.
Then I can use this collection with Contains() method that will be easily transleted into SQL.
But the main minus of explict checking in ApplicationService class is that we do increase processing by CPU (maybe on 0.00000000...1 :) ) with useless operations (4 conditions against 1 for 1 row in db)
The problem of remembering is here as well.

It mustn't have been so long, nonetheless, unfortunately, I can't decide.
I need one more answer for it, so I'm looking forward for your answers. Thanks)

DmyMi (approved):

PR #599, opened 2022-05-04
RomanchukLukashov/BlockedProviderParent

Sandaljero:
Functionality for blocking provider + parent

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/BlockedProviderParentController.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/BlockedProviderParentController.cs
@@ -1,8 +1,11 @@
 ﻿using System;
+using System.Security.Authentication;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authorization;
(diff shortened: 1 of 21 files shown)


OutOfSchool/OutOfSchool.WebApi/Controllers/V1/BlockedProviderParentController.cs
```diff
@@ -0,0 +1,124 @@
...
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        public async Task<IActionResult> GetBlock(Guid parentId, Guid providerId)
+        {
+            return Ok(await blockedProviderParentService.GetBlock(parentId, providerId).ConfigureAwait(false));
+        }
+
+        private string GetUserId()
```
DmyMi, line 116: i think we already have an extension for that. Please check common project or webapi extensions.

DmyMi (approved):

Issue #465, opened 2021-12-17
History of changes Ministry Admin


ulianachukhna:
As a user (Ministry-admin) I want to see the history of changes so that to know what changes were made whom and when 

**Role user** 
(Ministry-admin) 

**Precondition** 
Changes were made on the platform 

 **Acceptance Criteria**

SCENARIO #1 The Ministry-admin views changes in the cabinet  

GIVEN admin enters the cabinet 
AND opens the history of changes tab 

AND selects period 

WHEN clicks “Показати історію змін” 

THEN history of changes is displayed 

<img width="463" alt="history of changes" src="https://user-images.githubusercontent.com/88023871/146541360-f18f7c0c-7bd6-4f2e-b1bb-8da3aa25469b.png">

**Business Rule**
History of changes can be viewed by; 
Tech-admin 
Ministry admin 
District admin 
Service provider 
Information fields to display: 
(Дата /час/хто робив зміну) 
Назва 
ЄРДПО 
ПІБ Директора 
Підпорядкування 
Юридична адресса 
Заміна Адмінів (for example, there was one admin in the workshop, then they changed it to another, or, for example, the director of an institution was changed) 
Зміна статусів заявок 

Filters: 
Надавачі 
1)Назва 
2)ЄРДПО 
3)ПІБ Директора 
4)Підпорядкування 
5)Юридична адресса 
Адміни 
1)Заміна Адмінів 
Заяви 
1)Зміна статусів заявок 

According to the permissions matrix:
https://docs.google.com/spreadsheets/d/1yD1PH5b-4RjtNjNkTZYhOERCRvD3aFK71KVL2noImAw/edit#gid=404315089

**Epic link**
E.g.: Epic #100 [Epic](https://github.com/ita-social-projects/OoS-Backend/labels/100)

**Labels to be added**
"User story", Correspondind Functionality (e.g. User), Priority ("pri: "), "Corresponding Epic"

### Tasks 
1. - [ ] sample task.

PR #604, opened 2022-05-11
History of changes

i-rakityanskiy:
https://github.com/ita-social-projects/OoS-Backend/issues/457
https://github.com/ita-social-projects/OoS-Backend/issues/465

### DIFF
--- a/OutOfSchool/OutOfSchool.Common/Extensions/StringExtensions.cs
+++ b/OutOfSchool/OutOfSchool.Common/Extensions/StringExtensions.cs
@@ -16,7 +16,7 @@ namespace OutOfSchool.Common.Extensions
         }
 
         public static string Limit(this string value, int maxLength)
-            => value.Length <= maxLength || maxLength <= 0
+            => value == null || value.Length <= maxLength || maxLength <= 0
             ? value
             : value.Substring(0, maxLength);
     }
--- a/OutOfSchool/OutOfSchool.Common/Extensions/StringExtensions.cs
+++ b/OutOfSchool/OutOfSchool.Common/Extensions/StringExtensions.cs
@@ -16,8 +16,10 @@ namespace OutOfSchool.Common.Extensions
         }
 
         public static string Limit(this string value, int maxLength)
-            => value == null || value.Length <= maxLength || maxLength <= 0
-            ? value
-            : value.Substring(0, maxLength);
+            => maxLength < 0
+            ? throw new ArgumentOutOfRangeException(nameof(maxLength), maxLength, "Max length cannot be less than 0.")
+            : value == null || value.Length <= maxLength || maxLength == 0
+                ? value
+                : value.Substring(0, maxLength);
     }
 }
--- a/OutOfSchool/OutOfSchool.Common/Extensions/StringExtensions.cs
+++ b/OutOfSchool/OutOfSchool.Common/Extensions/StringExtensions.cs
@@ -14,5 +14,10 @@ namespace OutOfSchool.Common.Extensions
 
             return Enum.TryParse<TEnum>(value, true, out var result) ? result : defaultValue;
         }
+
+        public static string Limit(this string value, int maxLength)
+            => value.Length <= maxLength || maxLength <= 0
+            ? value
+            : value.Substring(0, maxLength);
     }
 }
--- a/OutOfSchool/OutOfSchool.DataAccess/Extensions/EntityEntryExtensions.cs
+++ b/OutOfSchool/OutOfSchool.DataAccess/Extensions/EntityEntryExtensions.cs
@@ -1,0 +1,16 @@
+﻿using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.ChangeTracking;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Metadata;
+
+namespace OutOfSchool.Services.Extensions
+{
+    public static class EntityEntryExtensions
+    {
+        public static string GetTableName(this EntityEntry entry) =>
+            entry.Metadata.GetAnnotation("Relational:TableName").Value.ToString();
+
+        public static string GetColumnName(this PropertyEntry entry) =>
+            entry.Metadata.GetColumnName(StoreObjectIdentifier.Create(entry.EntityEntry.Metadata, StoreObjectType.Table).Value);
+    }
+}
--- a/OutOfSchool/OutOfSchool.DataAccess/Extensions/EntityEntryExtensions.cs
+++ b/OutOfSchool/OutOfSchool.DataAccess/Extensions/EntityEntryExtensions.cs
@@ -1,15 +1,11 @@
 ﻿using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.ChangeTracking;
-using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Metadata;
 
 namespace OutOfSchool.Services.Extensions
 {
     public static class EntityEntryExtensions
     {
-        public static string GetTableName(this EntityEntry entry) =>
-            entry.Metadata.GetAnnotation("Relational:TableName").Value.ToString();
-
         public static string GetColumnName(this PropertyEntry entry) =>
             entry.Metadata.GetColumnName(StoreObjectIdentifier.Create(entry.EntityEntry.Metadata, StoreObjectType.Table).Value);
     }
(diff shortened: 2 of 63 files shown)


OutOfSchool/OutOfSchool.DataAccess/Extensions/EntityEntryExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+using Microsoft.EntityFrameworkCore.ChangeTracking;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Metadata;
+
+namespace OutOfSchool.Services.Extensions
+{
+    public static class EntityEntryExtensions
+    {
+        public static string GetTableName(this EntityEntry entry) =>
+            entry.Metadata.GetAnnotation("Relational:TableName").Value.ToString();
```
provicevko, line 11: Can you use `entry.Metadata.GetTableName()` here? Or at least you're able to use RelationalAnnotationNames.TableName instead of string

i-rakityanskiy: Good point. But this method is no longer used, so it will be removed.

OutOfSchool/OutOfSchool.Common/Extensions/StringExtensions.cs
```diff
@@ -14,5 +14,10 @@ public static TEnum ToEnum<TEnum>(this string value, TEnum defaultValue)
 
             return Enum.TryParse<TEnum>(value, true, out var result) ? result : defaultValue;
         }
+
+        public static string Limit(this string value, int maxLength)
```
provicevko, line 18: Just a comment. Looks a little bit strange when you use `string.Limit(-2)` then it returns value. Maybe should think about returning an empty string if <= 0

OutOfSchool/OutOfSchool.DataAccess/Extensions/QueryableExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public static TSource If<TSource>(this TSource source, bool condition, Func<TSource, TSource> filter)
+            where TSource : IQueryable
+        {
+            if (condition)
+            {
+                return filter(source);
+            }
+
+            return source;
```
provicevko, line 16: Doesn't matter but looks like `return condition ? filter(source) : source;`

OutOfSchool/OutOfSchool.DataAccess/Repository/ApplicationRepository.cs
```diff
@@ -22,35 +21,38 @@ public ApplicationRepository(OutOfSchoolDbContext dbContext)
...
         /// <param name="entity">Entity to update.</param>
+        /// <param name="onSaveChanges">Called before SaveChangesAsync method is invoked.</param>
         /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.
         /// The task result contains the entity that was updated.</returns>
         /// <exception cref="DbUpdateException">An exception that is thrown when an error is encountered while saving to the database.</exception>
         /// <exception cref="DbUpdateConcurrencyException">If a concurrency violation is encountered while saving to database.</exception>
-        public new async Task<Application> Update(Application entity)
+        public async Task<Application> Update(Application entity, Action<Application> onSaveChanges)
         {
             var application = dbSet.Find(entity.Id);
```
provicevko, line 47: If this method is async then maybe use `FindAsync()`?

OutOfSchool/OutOfSchool.DataAccess/Repository/ChangesLogRepository.cs
```diff
@@ -0,0 +1,118 @@
...
+
+            return properties.Concat(references);
+        }
+
+        private (Guid? entityIdGuid, long? entityIdLong) GetEntityId(EntityEntry entityEntry)
+        {
+            Guid? entityIdGuid = null;
+            long? entityIdLong = null;
+
+            var idProperty = entityEntry.Property("Id");
```
provicevko, line 80: I think it'd be better to get `string idColumnName` as parameter to expand usage.
And also just a comment I suppose you define which type id is? And then you add 2 ids as record in table, one of which is null. So it could be used with `string Id` and one more column as `IdType` (using enum). But I guess it adds a little bit more complexity, so just a comment. Maybe somebody can offer something.
Or if I don't understand, I'm sorry.

OutOfSchool/OutOfSchool.IdentityServer/Services/ProviderAdminChangesLogService.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        private readonly IEntityRepository<ProviderAdminChangesLog> providerAdminChangesLogRepository;
+
+        public ProviderAdminChangesLogService(IEntityRepository<ProviderAdminChangesLog> providerAdminChangesLogRepository)
+        {
+            this.providerAdminChangesLogRepository = providerAdminChangesLogRepository;
+        }
+
+        public async Task<int> SaveChangesLogAsync(ProviderAdmin entity, string userId, OperationType operationType)
+        {
```
provicevko, line 21: Check `ProviderAdmin entity` on null

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/ChangesLogRepositoryTests.cs
```diff
@@ -0,0 +1,253 @@
...
+
+        #region AddChangesLogToDbContext
+        [Test]
+        public async Task AddChangesLogToDbContext_WhenEntityIsModified_AddsLogToDbContext()
+        {
+            // Arrange
+            using var context = GetContext();
+            var changesLogRepository = GetChangesLogRepository(context);
+            var trackedProperties = new[] { "FullTitle", "EdrpouIpn", "Director", "LegalAddress" };
+            var provider = await context.Providers.FirstOrDefaultAsync();
```
provicevko, line 47: In some places `First()` is enough

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/ChangesLogRepositoryTests.cs
```diff
@@ -0,0 +1,253 @@
...
+            provider.FullTitle += "new";
+            provider.Director += "new";
+
+            var newFullTitle = provider.FullTitle;
+            var newDirector = provider.Director;
+
+            var added = changesLogRepository.AddChangesLogToDbContext(provider, user.Id, trackedProperties, valueProjector.ProjectValue);
+
+            // Assert
+            var fullTitleChanges = context.ChangesLog.Local
```
provicevko, line 62: In some places `nameof(Provider)`, `nameof(Provider.FullTitle)` and the same in `trackedProperties`

i-rakityanskiy: I agree. But, in my opinion, using string literals in tests is more readable. In any case, it does not have a negative impact.

OutOfSchool/OutOfSchool.IdentityServer/Services/ProviderAdminChangesLogService.cs
i-rakityanskiy: Given service usage scenarios, we should never have nulls when calling this method. I would leave it as is.

provicevko: Yeah, but you know this. What will happen if I don't know and my entity will be null because some methods allows null? Your method will throw null reference that is much worse for understanding that argument null exception. You cannot predict how somebody uses that considering its `public` modifier which is accesable outside of other projects.

OutOfSchool/OutOfSchool.DataAccess/Repository/ChangesLogRepository.cs
provicevko: Agree with that, that's why I've said 'just a comment for considering' :)

OutOfSchool/OutOfSchool.DataAccess/Repository/ChangesLogRepository.cs
```diff
@@ -0,0 +1,118 @@
...
+namespace OutOfSchool.Services.Repository
+{
+    public class ChangesLogRepository : EntityRepository<ChangesLog>, IChangesLogRepository
+    {
+        public ChangesLogRepository(OutOfSchoolDbContext dbContext)
+            : base(dbContext)
+        {
+        }
+
+        public ICollection<ChangesLog> AddChangesLogToDbContext<TEntity>(
```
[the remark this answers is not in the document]
provicevko: You fixed but it can be rewritten in short form as (for future):
```
_ = trackedProperties ?? throw new ArgumentNullException(nameof(trackedProperties));
_ = valueProjector?? throw new ArgumentNullException(nameof(valueProjector));
```
Can be marked as resolved after reading this comment.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/ChangesLogRepositoryTests.cs
provicevko: The main problem literals in such form is that they must be changed in all places every time your property name is changed. You've already used `nameof()` in `ArgumentNullException()`. You could use literal string there as well, why couldn't you? :)

DmyMi (approved):

PR #607, opened 2022-05-14
Rbiyovskiy/feature 292 about support laws and regulations pages

RBiyovskiy:
Implemented a general logic for the AboutPortal, SupportInformation and Laws and Regulations pages.
Changes in db: 
 - renamed next tables
    AboutPortal -> CompanyInformation
   AboutPortalItems -> CompanyInformationItems
- deleted SupportInformation table.
Added a base CompanyInformationController for the AboutPortalController, SupportInformationController and LawsAndRegulationsController that implemented GET and PUT methods.
So in the swagger we see the same description for the AboutPortal, SupportInformation and LawsAndRegulations controllers.
Please notice it and maybe, you will offer a more informative description.


OutOfSchool/OutOfSchool.WebApi/Services/CompanyInformationService.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// </summary>
+        /// <param name="companyInformationRepository">CompanyInformation repository.</param>
+        /// <param name="logger">Logger.</param>
+        /// <param name="mapper">Mapper.</param>
+        public CompanyInformationService(
+            IEntityRepository<CompanyInformation> companyInformationRepository,
+            ILogger<CompanyInformationService> logger,
+            IMapper mapper)
+        {
+            this.companyInformationRepository = companyInformationRepository;
```
VadymLevkovskyi, line 38: Please ensure all ctor arguments are correct, you can use `?? throw new ArgumentNullException(...)` 

This has to be done, as through all places in class you are using repo/mapper/logger without null-check so your class does require them to be not-null and it's better to know immediately if they are not.

OutOfSchool/OutOfSchool.WebApi/Services/CompanyInformationService.cs
```diff
@@ -0,0 +1,115 @@
...
+using OutOfSchool.Services.Models;
+using OutOfSchool.Services.Repository;
+using OutOfSchool.WebApi.Models;
+
+namespace OutOfSchool.WebApi.Services
+{
+    /// <summary>
+    /// Implements the interface with CRUD functionality for the CompanyInformation entities (AboutPortal, SupportInformation and LawsAndRegulations).
+    /// </summary>
+    public class CompanyInformationService : ICompanyInformationService
```
VadymLevkovskyi, line 19: Please cover this class with tests

VadymLevkovskyi (changes requested):
Small changes required. Not sure about our current policy re tests - please check with @DmyMi

OutOfSchool/OutOfSchool.WebApi/Extensions/MappingExtensions.cs
```diff
@@ -179,6 +179,15 @@ public static WorkshopCard ToModelSimple(this Workshop workshop)
             });
         }
 
+        public static CompanyInformationDto ToModel(this CompanyInformation companyInformation)
```
VadymLevkovskyi, line 182: Please don't introduce these converters on project level as they might be used not only for testing purposes. 
An issue with such implementation is performance penalty for case of processing big amount of entities, that's why we prepare and register mapper once in DI and not recreate it for each processed DTO/model.

So either move and isolate this code in tests or rewrite tests (if possible, ofc)

OutOfSchool/OutOfSchool.WebApi.Tests/Services/CompanyInformationServiceTests.cs
```diff
@@ -0,0 +1,193 @@
...
+            var changedEntity = GetEntityDto(type);
+
+            // Act
+            mapper.Setup(m => m.Map<IEnumerable<CompanyInformationItem>>(It.IsAny<IEnumerable<CompanyInformationItemDto>>())).Returns(changedEntity.ToDomain().CompanyInformationItems);
+            mapper.Setup(m => m.Map<CompanyInformationDto>(It.IsAny<CompanyInformation>())).Returns(changedEntity);
+
+            var result = await service.Update(changedEntity, type);
+
+            // Assert
+            Assert.That(changedEntity.Title, Is.EqualTo(result.Title));
```
VadymLevkovskyi, line 99: This set of `Assert`s is duplicated in next test method. 2 times duplication is moderately ok, but for 3rd and subsequent time consider to reuse extracted helper method that perform these asserts in one place. 

Moreover it looks like you have another helper - to assert equality of `CompanyInformationItem`.

And also consider collection equality asserts provided by framework or write your own (or maybe any is already written by other devs - I don't remember if this is a case).

VadymLevkovskyi (approved):
Approved with minor comments.

OutOfSchool/OutOfSchool.WebApi/Extensions/MappingExtensions.cs
DmyMi: Curently we're trying to switch to using mapping profile and inject the mapper instead of static extensions.
Mapping profile is here: [https://github.com/ita-social-projects/OoS-Backend/blob/develop/OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs](https://github.com/ita-social-projects/OoS-Backend/blob/develop/OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs)
Example usage: [https://github.com/ita-social-projects/OoS-Backend/blob/develop/OutOfSchool/OutOfSchool.WebApi/Services/ClassService.cs#L51](https://github.com/ita-social-projects/OoS-Backend/blob/develop/OutOfSchool/OutOfSchool.WebApi/Services/ClassService.cs#L51)

DmyMi (approved):

PR #621, opened 2022-05-21
Provicevko/refactor images

provicevko:
**Please note that this PR does not change internal logic of images, it just a refactoring with improvements.**
**It seems large because of:**
1) Migration.
2) Some moved blocks (for example interfaces) were recreated as new.

**The second (final stage) of refactoring files/images/gcp.**
- Changed Gcp storage lifetime.
- Changed IImageDependent interface, now only this interface must be used for entity with images.
- Encapsulated logic from WorkshopService/TeacherService ChangeCoverImage() into ImageDependentEntityImagesInteractionMediator class. This class is in charge of operating with all images process. It is a mediator between Entity from service and ImageService (just uploads/removes). It's internal because just hides all alghoritms of checking/uploading/removing.
- Now ImageDependentEntityImagesInteractionMediator does not update entity by itself so this change decrease amount of queries to db.
- Removed unnecessary files (interfaces, empty descandants of ImageDependentEntityImagesInteractionMediator those had been made to identify; now I register in DI as generic).
- Rewrote logs.

**Result:** 
- simplified structure with ability to extend.
- increased performance.

**Now [in process]:** _I'm going to test/fix potential issues._

**Future plans:**
- Adding Provider images.
- Adding related tests.
- Solving a problem of the desynchronization.


OutOfSchool/OutOfSchool.WebApi/Resources/Images/ImageResource.uk.resx
```diff
@@ -153,10 +150,10 @@
   <data name="UnexpectedValidationError" xml:space="preserve">
     <value>Будь-ласка, переконайтесь у коректності фото</value>
```
VadymLevkovskyi, line 151: (Finally I see some refactoring PR touching this file)
Please fix spelling - dash is not used here: 
[Будь ласка](http://slovopys.kubg.edu.ua/bud-laska-chy-bud-laska/#:~:text=%D0%91%D1%83%D0%B4%D1%8C%20%D0%BB%D0%B0%D1%81%D0%BA%D0%B0%20%D1%86%D0%B5%20%D0%B2%D0%B0%D0%BC%20%D0%BD%D0%B5,%D1%87%D0%B5%D1%80%D0%B5%D0%B7%20%D0%B4%D0%B5%D1%84%D1%96%D1%81%2C%20%D0%B0%20%D0%BA%D0%B0%D1%82%D0%B5%D0%B3%D0%BE%D1%80%D0%B8%D1%87%D0%BD%D0%BE%20%D0%BE%D0%BA%D1%80%D0%B5%D0%BC%D0%BE.)
Also (few lines above) write words separately: 
[Нажаль](https://ukr-mova.in.ua/library/orfografiya/na-zhal)

OutOfSchool/OutOfSchool.WebApi/Services/Images/ImageDependentEntityImagesInteractionMediator.cs
```diff
@@ -21,152 +21,168 @@ namespace OutOfSchool.WebApi.Services.Images
...
     /// <summary>
-    /// Represents a base class for operations with images.
+    /// Represents a mediator class for operations with images.
+    /// This instance does not save the given entity changes.
     /// </summary>
     /// <typeparam name="TEntity">Entity type.</typeparam>
-    /// <typeparam name="TKey">The type of entity Id.</typeparam>
-    public class ImageInteractionBaseService<TEntity, TKey> : IImageInteractionService<TKey>
-        where TEntity : class, IKeyedEntity<TKey>, IImageDependentEntity<TEntity>, new()
+    internal class ImageDependentEntityImagesInteractionMediator<TEntity> : IImageDependentEntityImagesInteractionMediator<TEntity>
```
VadymLevkovskyi, line 28: Not a call to action, just consideration.
It's definitely not a [Mediator pattern](https://refactoring.guru/design-patterns/mediator). 
Could other name be chosen (and why `service` is not good)?
Could all logic (ImageService + this Mediator) be in one `service` that just implements 2 different interfaces?

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
```diff
@@ -109,7 +105,10 @@ public async Task<WorkshopCreationResultDto> CreateV2(WorkshopDTO dto)
             // In case if DirectionId and DepartmentId does not match ClassId
             await FillDirectionsFields(dto).ConfigureAwait(false);
 
-            Func<Task<Workshop>> operation = async () =>
+            MultipleImageUploadingResult uploadingResult = null;
```
VadymLevkovskyi, line 108: (Another styling consideration, not a change request, just thing that irritates my eye, but I'm not insisting)
Variables `uploadingResult` and `uploadingCoverImageResult` are both results of `TransactionOperation` (latter name btw could be more specific - like `CreateWorkshopAndDependencies`) execution.
They are not set before it's execution (so they are not hidden parameters).
Why not make `TransactionOperation` more `pure` function ( [wiki](https://en.wikipedia.org/wiki/Pure_function) ) and return all results as, well, function result? Something like
`
async Task<(Workshop createdWorkshop, MultipleImageUploadingResult imagesUploadResult, Result<string> coverImageUploadResult)> TransactionOperation()
`
and then use it as
`
var (newWorkshop, imagesUploadResult, coverImageUploadResult) = await workshopRepository.RunInTransaction(TransactionOperation).ConfigureAwait(false);
`

OutOfSchool/OutOfSchool.WebApi/Services/Images/ImageDependentEntityImagesInteractionMediator.cs
```diff
@@ -21,152 +21,168 @@ namespace OutOfSchool.WebApi.Services.Images
...
+
+            var result = new ImageChangingResult();
+
+            if (!string.IsNullOrEmpty(oldImageId))
             {
-                return Result<string>.Failed(ImagesOperationErrorCode.NoGivenImagesError.GetOperationError());
+                result.RemovingResult = await RemoveImageProcessAsync(entity, oldImageId).ConfigureAwait(false);
+                if (!result.RemovingResult.Succeeded)
+                {
+                    return result;
```
VadymLevkovskyi, line 82: Shouldn't another `Log.Trace` be called here?

VadymLevkovskyi: See similar pattern in next functions.

Probably we need more error-prone approach ,and from my experience I see two and half possible ways to avoid such problem of missing `Log.Trace()` near all returns: 

1. either use explicitly `try-finally` block wrapping whole method body like:
```
public async Task<ImageChangingResult> ChangeImageAsync(TEntity entity, string oldImageId, IFormFile newImage)
{
    _ = entity ?? throw new ArgumentNullException(nameof(entity));
    try 
    {
        Logger.LogTrace("Updating an image the for entity was started");
        //... method code with return operators in different places
    }
    finally
    {
        Logger.LogTrace("Updating an image the for entity was finished");
    }
}
```

1,5. same as previous but implicit `try-finally` created by `using` block with special disposable class (like mentioned below `MethodTraceLog` that might log "start" on constructing and "finish" on `Dispose`) - bit worse from CPU and used memory but neater in code, esp. when you can use new `using` syntax (without block {}):
```
public async Task<ImageChangingResult> ChangeImageAsync(TEntity entity, string oldImageId, IFormFile newImage)
{
    _ = entity ?? throw new ArgumentNullException(nameof(entity));
    using(var traceLog = new MethodTraceLog("Updating an image the for entity"))
    {
        //... method code with return operators in different places
    }
}
```

2. or using some post processors (PostSharp or any other)

Calling @DmyMi to thread, probably he can add more reasons pro or contra

OutOfSchool/OutOfSchool.WebApi/Services/Images/ImageDependentEntityImagesInteractionMediator.cs
```diff
@@ -21,152 +21,168 @@ namespace OutOfSchool.WebApi.Services.Images
...
 
         /// <inheritdoc/>
-        public async Task<ImageUploadingResult> UploadManyImagesAsync(TKey entityId, IList<IFormFile> images)
+        public async Task<Result<string>> AddCoverImageAsync(TEntity entity, IFormFile image)
         {
-            if (images == null || images.Count <= 0)
+            _ = entity ?? throw new ArgumentNullException(nameof(entity));
+            _ = image ?? throw new ArgumentNullException(nameof(image));
+
+            if (!string.IsNullOrEmpty(entity.CoverImageId))
```
VadymLevkovskyi, line 140: No usage of `Trace.Log` here. Bit inconsistent, I would say

OutOfSchool/OutOfSchool.WebApi/Services/Images/ImageDependentEntityImagesInteractionMediator.cs
```diff
@@ -21,152 +21,168 @@ namespace OutOfSchool.WebApi.Services.Images
...
         {
-            if (imageIds == null || imageIds.Count <= 0)
+            _ = entity ?? throw new ArgumentNullException(nameof(entity));
+            if (string.IsNullOrEmpty(entity.CoverImageId))
             {
-                return new ImageRemovingResult
-                {
-                    MultipleKeyValueOperationResult = new MultipleKeyValueOperationResult { GeneralResultMessage = ImagesOperationErrorCode.NoGivenImagesError.GetResourceValue() },
-                };
+                throw new InvalidOperationException(@"Image id must be a non empty string");
```
VadymLevkovskyi, line 161: Probably better to use `ArgumentException`, as it's an issue with input argument, not a processing code

OutOfSchool/OutOfSchool.WebApi/Services/Images/ImageDependentEntityImagesInteractionMediator.cs
```diff
@@ -21,152 +21,168 @@ namespace OutOfSchool.WebApi.Services.Images
...
-        /// <summary>
-        /// Gets the entity with included images.
-        /// </summary>
-        /// <param name="entityId">Entity id.</param>
-        /// <returns>The instance of <see cref="Task"/> that represents <c>TEntity</c>.</returns>
-        /// <exception cref="InvalidOperationException">When filter is not specified.</exception>
-        protected virtual async Task<TEntity> GetEntityWithIncludedImages(TKey entityId)
-        {
-            return (await Repository.GetByFilter(x => x.Id.Equals(entityId), nameof(IImageDependentEntity<TEntity>.Images)).ConfigureAwait(false)).FirstOrDefault();
+            if (string.Equals(dtoImageId, entity.CoverImageId, StringComparison.Ordinal) &&
```
VadymLevkovskyi, line 181: Please swap && operands. 
It makes sense to check `(!string.IsNullOrEmpty(dtoImageId) || newImage == null)` first and only then use more complicated `string.Equals`

OutOfSchool/OutOfSchool.WebApi/Services/Images/ImageDependentEntityImagesInteractionMediator.cs
```diff
@@ -21,152 +21,168 @@ namespace OutOfSchool.WebApi.Services.Images
...
+
+        /// <inheritdoc/>
+        public async Task<OperationResult> RemoveImageAsync(TEntity entity, string imageId)
+            => await RemoveImageProcessAsync(entity, imageId).ConfigureAwait(false);
+
+        public async Task<ImageChangingResult> ChangeImageAsync(TEntity entity, string oldImageId, IFormFile newImage)
         {
-            if (image == null)
+            _ = entity ?? throw new ArgumentNullException(nameof(entity));
+            Logger.LogTrace("Updating an image the for entity was started");
```
VadymLevkovskyi, line 73: "Updating an image the for entity" - wording issue, probably excessive `the` article or missed noun

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
VadymLevkovskyi: I see, and agree, my comment is irrelevant. 

In general I'll stand for less code but more diversity when possible - so `option 1` looks better.

Downsides of it are:
- possible necessity to explicitly specify method type - need to check how compiler wants it to be used, 
- and this is not backward compatible change - check please scope of changes (how many places are calling this `RunInTransaction<T>(Func<T> func)` from `EntityBaseRepository<TEntity>`, and should any additional code fix be done). 

If changes are way too big or syntax is too awkward - then `option 2`.

And ofc, I don't insist to do it in this PR.

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Services/Images/ImageDependentEntityImagesInteractionService.cs
```diff
@@ -369,7 +369,11 @@ protected virtual bool AllowedToUploadGivenAmountOfFiles(TEntity entity, int cou
 
                 await ImageService.RemoveManyImagesAsync(imageIds).ConfigureAwait(false);
 
-                var imagesRemovingResult = new MultipleImageRemovingResult();
+                var imagesRemovingResult = new MultipleImageRemovingResult
```
VadymLevkovskyi, line 372: This looks like `MultipleImageRemovingResult` ctor responsibility. 
I mean next rule: constructor should create valid object; if object is invalid after constructing and you need to set some properties to make it valid - you have issues with constructing.

PR #623, opened 2022-05-23
Additional condition to avoid "NullReferenceException" has been added

valerabad:


OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
```diff
@@ -127,7 +127,7 @@ public async Task<IEnumerable<ProviderDto>> GetAll()
 
             foreach (var provider in providersDTO)
             {
-                if (averageRatings.Any())
+                if (averageRatings.Any(r => r.Key == provider.Id))
```
VadymLevkovskyi, line 130: Please improve performance of this code. Now you search for `averageRatings` with `Key == provider.Id` twice - in `Any()` and then in `FirstOrDefault()`. Please use only `FirstOrDefault()` and in `if` ensure that result is not `null` (be aware that deconstructing probably should be done after check for `null`, but I'm not sure as I can't see `averageRatings` type).

VadymLevkovskyi (changes requested):
See comment

PR #628, opened 2022-05-23
Add possibility to override connection string with a strongly typed options

DmyMi:
- Base options needed for every service
- Possible to have different options per service
- Possible to change any database connection option by changing a single environment variable instead of generating a new connection string all the time even for a small change

### DIFF
--- a/OutOfSchool/OutOfSchool.Common/Config/IMySqlConnectionOptions.cs
+++ b/OutOfSchool/OutOfSchool.Common/Config/IMySqlConnectionOptions.cs
@@ -14,6 +14,6 @@ namespace OutOfSchool.Common.Config
 
         public string Password { get; set; }
 
-        public bool IsGuidFormatBinary16 { get; set; }
+        public string GuidFormat { get; set; }
     }
 }
--- a/OutOfSchool/OutOfSchool.Common/Config/IMySqlConnectionOptions.cs
+++ b/OutOfSchool/OutOfSchool.Common/Config/IMySqlConnectionOptions.cs
@@ -1,0 +1,19 @@
+namespace OutOfSchool.Common.Config
+{
+    public interface IMySqlConnectionOptions
+    {
+        public bool UseOverride { get; set; }
+
+        public string Server { get; set; }
+
+        public uint Port { get; set; }
+
+        public string Database { get; set; }
+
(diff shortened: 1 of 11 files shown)


OutOfSchool/OutOfSchool.Common/Config/IMySqlConnectionOptions.cs
```diff
@@ -0,0 +1,19 @@
+namespace OutOfSchool.Common.Config
+{
+    public interface IMySqlConnectionOptions
+    {
+        public bool UseOverride { get; set; }
+
+        public string Server { get; set; }
+
+        public uint Port { get; set; }
```
VadymLevkovskyi, line 9: Wow `uint`! First time see this type really used in C# project :)

DmyMi: it is used in the Pomelo `MySqlConnectionBuilder` class :(

OutOfSchool/OutOfSchool.Common/Extensions/Startup/ConnectionStringExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+                    ConnectionString = config.GetConnectionString(connectionStringKey),
+                };
+            }
+
+            if (string.IsNullOrEmpty(connectionStringBuilder.ConnectionString))
+            {
+                throw new Exception("Provide a valid connection string or options");
+            }
+
+            if (!connectionStringBuilder.ContainsKey("guidformat") || connectionStringBuilder["guidformat"].ToString().ToLower() != "binary16")
```
VadymLevkovskyi, line 46: Please don't use `.ToLower()` for comparing strings, please use `String.Equals(..., ..., StringComparison.OrdinalIgnoreCase)`

VadymLevkovskyi (approved):

OutOfSchool/OutOfSchool.Common/Config/IMySqlConnectionOptions.cs
VadymLevkovskyi: Well, then all credits go to Pomelo author.

PR #631, opened 2022-05-25
renamed and changed enum

RBiyovskiy:
I have created pull request only to seen changes, without merge in a future.


OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ProviderController.cs
```diff
@@ -116,7 +116,8 @@ public async Task<IActionResult> GetProfile()
         /// </summary>
         /// <param name="providerModel">Entity to add.</param>
         /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
-        [HasPermission(Permissions.ProviderAddNew)]
+        //[HasPermission(Permissions.ProviderAddNew)]
```
alrom, line 119: )) You forget something to return

PR #633, opened 2022-05-25
Provicevko/provider images

provicevko:
`ProviderController`, `ProviderService` methods was just copied into V2 version.
`ProviderController` changes:
- Some messages in logs were replaced
- Deleted unnecessary checking dto on ModelState.IsValid or null
- Added `FromForm`, `Consumes` attributes

`ProviderService` changes:
- Added `private UpdateProvider()` that contains `try{}/catch{}` inside
- Added some checking arguments on null

Provider images:
- Added `IImageDependent` interface to `Provider` model
- Made migration
- Registered images service for `Provider` in DI
- Added relevant method calls (using images) in `ProviderService`


OutOfSchool/OutOfSchool.WebApi/Services/IProviderService.cs
```diff
@@ -18,6 +18,13 @@ public interface IProviderService
         /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
         Task<ProviderDto> Create(ProviderDto providerDto);
 
+        /// <summary>
+        /// Add entity to the database.
+        /// </summary>
+        /// <param name="dto">Entity to add.</param>
+        /// <returns>The <see cref="Task"/> that represents the asynchronous operation, containing the <see cref="ProviderDto"/>.</returns>
+        Task<ProviderDto> CreateV2(ProviderDto dto);
```
VadymLevkovskyi, line 26: Please consider if there are any benefits to introduce separate `...V2` interface, and either use it as base in `IProvideService` (not a good idea, just in case) or implement it in specific provider classes?

OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
```diff
@@ -108,6 +115,57 @@ public async Task<ProviderDto> Create(ProviderDto providerDto)
...
+            var providerDomainModel = providerDto.ToDomain();
+
+            // BUG: concurrency issue:
+            //      while first repository with this particular user id is not saved to DB - we can create any number of repositories for this user.
+            if (providerRepository.SameExists(providerDomainModel))
+            {
+                throw new InvalidOperationException(localizer["There is already a provider with such a data"]);
+            }
+
+            var users = await usersRepository.GetByFilter(u => u.Id.Equals(providerDto.UserId)).ConfigureAwait(false);
```
VadymLevkovskyi, line 145: All code before `var users = await usersRepository.GetByFilter...` seems to be related to `providerDto` validation and copy of same code in `Create` method. Could you extract method and thus avoid code duplication?

OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
```diff
@@ -269,6 +327,93 @@ public async Task<ProviderDto> Update(ProviderDto providerDto, string userId)
             }
         }
 
+        /// <inheritdoc/>
+        public async Task<ProviderDto> UpdateV2(ProviderDto providerDto, string userId)
```
VadymLevkovskyi, line 331: Please check if there are possibilities to avoid code duplication in `Update` and `UpdateV2`

OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
```diff
@@ -292,6 +437,38 @@ public async Task Delete(Guid id)
             }
         }
 
+        /// <inheritdoc/>
+        public async Task DeleteV2(Guid id)
+        {
```
VadymLevkovskyi, line 442: Same here, consider if you could avoid code duplication

OutOfSchool/OutOfSchool.WebApi/Services/IProviderService.cs
provicevko: Okey, sounds really better, I'll rewrite

OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
```diff
@@ -396,10 +285,10 @@ public async Task<ProviderDto> UpdateV2(ProviderDto providerDto, string userId)
                         mapper.Map(providerDto, checkProvider);
                     }
 
-                    await providerImagesService.ChangeImagesAsync(checkProvider, providerDto.ImageIds ?? new List<string>(), providerDto.ImageFiles)
-                        .ConfigureAwait(false);
-
-                    await providerImagesService.ChangeCoverImageAsync(checkProvider, providerDto.CoverImageId, providerDto.CoverImage).ConfigureAwait(false);
+                    if (actionBeforeUpdating != null)
+                    {
+                        await actionBeforeUpdating.Invoke(checkProvider).ConfigureAwait(false);
```
VadymLevkovskyi, line 290: `.Invoke` seems to be excessive here, you could just 
`await actionBeforeUpdating(checkProvider).ConfigureAwait(false);`

VadymLevkovskyi (approved):
Minor code style issue, see comment

PR #638, opened 2022-05-25
RomanchukLukashov/AdminStructure

Sandaljero:
First step!

### DIFF
--- a/OutOfSchool/OutOfSchool.DataAccess/Models/SubordinationStructure/InstitutionFieldDescription.cs
+++ b/OutOfSchool/OutOfSchool.DataAccess/Models/SubordinationStructure/InstitutionFieldDescription.cs
@@ -3,7 +3,7 @@ using System.ComponentModel.DataAnnotations;
 
 namespace OutOfSchool.Services.Models.SubordinationStructure
 {
-    public class InstitutionFieldDescription
+    public class InstitutionFieldDescription : IKeyedEntity<Guid>
     {
         public Guid Id { get; set; }
 
--- a/OutOfSchool/OutOfSchool.DataAccess/Models/SubordinationStructure/InstitutionFieldDescription.cs
+++ b/OutOfSchool/OutOfSchool.DataAccess/Models/SubordinationStructure/InstitutionFieldDescription.cs
@@ -1,0 +1,21 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace OutOfSchool.Services.Models.SubordinationStructure
+{
+    public class InstitutionFieldDescription
+    {
+        public Guid Id { get; set; }
+
+        [MinLength(1)]
+        [MaxLength(100)]
+        public string Title { get; set; }
+
+        public int HierarchyLevel { get; set; }
+
+        [Required]
+        public Guid InstitutionId { get; set; }
+
+        public virtual Institution Institution { get; set; }
+    }
+}
--- a/OutOfSchool/OutOfSchool.WebApi/Services/SubordinationStructure/InstitutionHierarchyService.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Services/SubordinationStructure/InstitutionHierarchyService.cs
@@ -40,11 +40,11 @@ namespace OutOfSchool.WebApi.Services.SubordinationStructure
             IMapper mapper)
         {
             this.repository = repository ?? throw new ArgumentNullException(nameof(repository));
-            this.repositoryWorkshop = repositoryWorkshop;
-            this.repositoryProvider = repositoryProvider;
-            this.localizer = localizer;
-            this.logger = logger;
-            this.mapper = mapper;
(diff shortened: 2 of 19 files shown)


OutOfSchool/OutOfSchool.DataAccess/Models/SubordinationStructure/InstitutionFieldDescription.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace OutOfSchool.Services.Models.SubordinationStructure
+{
+    public class InstitutionFieldDescription : IKeyedEntity<Guid>
+    {
+        public Guid Id { get; set; }
```
DmyMi, line 8: do we need guid for every entity here? if some of those won't be exposed directly - we can use longs maybe?

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Services/SubordinationStructure/InstitutionHierarchyService.cs
```diff
@@ -0,0 +1,146 @@
...
+        public InstitutionHierarchyService(
+            ISensitiveEntityRepository<InstitutionHierarchy> repository,
+            IWorkshopRepository repositoryWorkshop,
+            IProviderRepository repositoryProvider,
+            ILogger<InstitutionHierarchyService> logger,
+            IStringLocalizer<SharedResource> localizer,
+            IMapper mapper)
+        {
+            this.repository = repository ?? throw new ArgumentNullException(nameof(repository));
+            this.repositoryWorkshop = repositoryWorkshop;
```
VadymLevkovskyi, line 43: Please either validate all arguments using `?? throw new ArgumentNullException` in constructor or check them for `null` everywhere in code (like in line 53 you could get `NullReferenceException` if logger is null).

OutOfSchool/OutOfSchool.WebApi/Services/SubordinationStructure/InstitutionHierarchyService.cs
```diff
@@ -0,0 +1,146 @@
...
+using Microsoft.Extensions.Localization;
+using Microsoft.Extensions.Logging;
+using OutOfSchool.Services.Models.SubordinationStructure;
+using OutOfSchool.Services.Repository;
+using OutOfSchool.WebApi.Common;
+using OutOfSchool.WebApi.Models.SubordinationStructure;
+
+namespace OutOfSchool.WebApi.Services.SubordinationStructure
+{
+    public class InstitutionHierarchyService : IInstitutionHierarchyService
```
VadymLevkovskyi, line 16: Is there any possibility to cover this class functionality with unit-test?

OutOfSchool/OutOfSchool.WebApi/Services/SubordinationStructure/InstitutionHierarchyService.cs
```diff
@@ -0,0 +1,146 @@
...
+        {
+            logger.LogInformation("Getting all InstitutionHierarchies started.");
+
+            var institutionHierarchies = await repository.GetAll().ConfigureAwait(false);
+
+            logger.LogInformation(!institutionHierarchies.Any()
+                ? "InstitutionHierarchy table is empty."
+                : $"All {institutionHierarchies.Count()} records were successfully received from the InstitutionHierarchy table.");
+
+            return institutionHierarchies.Select(entity => mapper.Map<InstitutionHierarchyDto>(entity)).ToList();
```
VadymLevkovskyi, line 97: Please be consistent - either change method result type to `Task<List<....` or don't do premature materialization in return using `.ToList()`. This might affect performance a bit.

VadymLevkovskyi: Similar point for `GetChildren`

OutOfSchool/OutOfSchool.WebApi/Services/SubordinationStructure/InstitutionHierarchyService.cs
alrom: Done

OutOfSchool/OutOfSchool.WebApi/Services/SubordinationStructure/InstitutionHierarchyService.cs
alrom: Done

OutOfSchool/OutOfSchool.WebApi/Services/SubordinationStructure/InstitutionHierarchyService.cs
alrom: Done

VadymLevkovskyi (approved):

PR #646, opened 2022-05-30
rbiyovskiy/642 service provider registration add non existent functionality

RBiyovskiy:
Done next:
- added the ProviderSectionItem entity and implemented logic;
- completely transferred the ProviderService class to use the new maper (replaced the implementation of the ToModel / ToDomain extension methods to use the AutoMaper)
- Fixed the ServiceProvider tests (all 14 are running successfully).

### DIFF
--- a/OutOfSchool/OutOfSchool.DataAccess/Models/ProviderSectionItem.cs
+++ b/OutOfSchool/OutOfSchool.DataAccess/Models/ProviderSectionItem.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace OutOfSchool.Services.Models
+{
+    public class ProviderSectionItem : IKeyedEntity<Guid>
+    {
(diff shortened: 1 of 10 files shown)


OutOfSchool/OutOfSchool.DataAccess/Models/ProviderSectionItem.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace OutOfSchool.Services.Models
+{
+    public class ProviderSectionItem : IKeyedEntity<Guid>
+    {
+        public Guid Id { get; set; }
```
DmyMi, line 8: as these items won't be directly exposed via a controller, maybe we can use Long as ID? save a few bytes per record. Not critical comparing to the size of other fields, but just a thought :)

DmyMi (approved):

VadymLevkovskyi (approved):

PR #656, opened 2022-06-04
Provicevko/gcp synchronization

provicevko:
**Changes:**
- moved default quartz config to separate extension method
- added custom quartz task pool that is used to add initial jobs with one default quartz
- created gcp sync data repository in order to make explict queries and avoid unwanted inner joins we have if wanna get images from entity with EntityRepositoryBase methods
- created quartz job for synchronize gcp storage images with main db OutOfSchool
- created relevant service for completing sync operation

**The default values for testing period:**
- sync every day at 00:00
- 25 objects per page from gcp

**Estimated values those must be after testing:**
 - sync every 3-6 months at 00:00
 - 500 objects per page (should be tested in related tests)

**Pay attention to:**
- `GcpImagesStorageSynchronizationService.SynchronizeAsync()` method.

**Last update:** deleted multithreading process for now.

**Future: Tests will be written after we agree the final version.**


OutOfSchool/OutOfSchool.DataAccess/Repository/Files/GcpFilesStorageBase.cs
```diff
@@ -25,6 +32,12 @@ protected GcpFilesStorageBase(IGcpStorageContext storageContext)
 
         private protected string BucketName { get; }
 
+        public async Task<IAsyncEnumerable<Objects>> GetBulkListsOfObjectsAsync(string prefix = null, ListObjectsOptions options = null)
+        {
+            var objects = await Task.Run(() => StorageClient.ListObjectsAsync(BucketName, prefix: prefix, options: options));
+            return objects.AsRawResponses();
+        }
```
i-rakityanskiy, line 39: Calling `Task.Run` looks redundant here. This method could be simplified.
```suggestion
        public IAsyncEnumerable<Objects> GetBulkListsOfObjects(string prefix = null, ListObjectsOptions options = null)
            => StorageClient.ListObjectsAsync(BucketName, prefix: prefix, options: options).AsRawResponses();
```

provicevko: Yeah, that's indeed true, there are no network requests before the enumerating. Thanks

OutOfSchool/OutOfSchool.WebApi/Services/Gcp/GcpImagesStorageSynchronizationService.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        /// <inheritdoc/>
+        public async Task SynchronizeAsync(CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                logger.LogDebug("Gcp storage synchronization was started");
+
+                // gcp returns data by local time
+                var dateTime = DateTime.Now.AddMinutes(-1);
```
DmyMi, line 67: maybe skip more then a minute (e.g., 1 hour or even 1 day) to be sure we won't delete files that are in process of writing to storage&database (if this job might take long time to complete)

OutOfSchool/OutOfSchool.WebApi/Services/Gcp/GcpImagesStorageSynchronizationService.cs
```diff
@@ -0,0 +1,138 @@
...
+                    searchIds.ExceptWith(syncResult);
+                }
+            }
+
+            return searchIds;
+        }
+
+        private IAsyncEnumerable<Objects> GetListsOfObjects()
+        {
+            var options = new ListObjectsOptions
```
DmyMi, line 113: not a call to immediate action. maybe we can store different "image types" (workshop cover, workshop images, teacher avatars, etc.) under different prefixes and? or go even deeper and prefix workshops for each ministry, etc. It will probably add some complexity to create/update, but will considerably reduce the amount of data that we get during cleanup job. And we can have different jobs for each prefix

OutOfSchool/OutOfSchool.DataAccess/Repository/Files/GcpImagesSyncDataRepository.cs
```diff
@@ -0,0 +1,88 @@
...
+        #endregion
+
+        private async Task<List<string>> GetIntersectEntityCoverImagesIds<TEntity>(
+            IQueryable<TEntity> dbSet,
+            IEnumerable<string> searchIds)
+            where TEntity : class, IKeyedEntity, IImageDependentEntity<TEntity>, new()
+        {
+            _ = searchIds ?? throw new ArgumentNullException(nameof(searchIds));
+
+            return await dbSet
```
h4wk13, line 69: this `await` looks redundant

OutOfSchool/OutOfSchool.DataAccess/Repository/Files/GcpImagesSyncDataRepository.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        private async Task<List<string>> GetIntersectEntityImagesIds<TEntity>(
+            IQueryable<Image<TEntity>> dbSet,
+            IEnumerable<string> searchIds)
+            where TEntity : class, IKeyedEntity, IImageDependentEntity<TEntity>, new()
+        {
+            _ = searchIds ?? throw new ArgumentNullException(nameof(searchIds));
+
+            return await dbSet
```
h4wk13, line 82: this `await` looks redundant

OutOfSchool/OutOfSchool.WebApi/Extensions/ElasticsearchSynchronizationExtension.cs
```diff
@@ -32,44 +33,14 @@ public static class ElasticsearchSynchronizationExtension
...
-                            }));
-                    q.SetProperty("quartz.jobStore.clustered", "true");
-                    q.SetProperty("quartz.jobStore.driverDelegateType", "Quartz.Impl.AdoJobStore.StdAdoDelegate, Quartz");
-                    q.SetProperty("quartz.jobStore.useProperties", "true");
-                    q.UseMicrosoftDependencyInjectionJobFactory();
-                    q.AddJob<ElasticsearchSynchronizationQuartz>(j => j.WithIdentity(jobKey));
-                    q.AddTrigger(t => t
+                QuartzPool.AddJob<ElasticsearchSynchronizationQuartz>(j => j.WithIdentity(jobKey));
+                QuartzPool.AddTrigger(t => t
                        .WithIdentity("elasticsearchJobTrigger")
```
h4wk13, line 40: `elasticsearchJobTrigger` looks like a `const`

OutOfSchool/OutOfSchool.WebApi/Extensions/ElasticsearchSynchronizationExtension.cs
```diff
@@ -32,44 +33,14 @@ public static class ElasticsearchSynchronizationExtension
 
             if (elasticSynchronizationSchedulerConfig.UseQuartz)
```
h4wk13, line 34: if `elasticSynchronizationSchedulerConfig` is controlled by the end user - it is possible it might be null, so it should be initialized with some default value in case of `null`, otherwise NRE will be thrown

OutOfSchool/OutOfSchool.WebApi/Extensions/Startup/FileStorageExtensions.cs
```diff
@@ -39,5 +48,26 @@ public static IServiceCollection AddImagesStorage(this IServiceCollection servic
             return services.AddScoped<IImageFilesStorage, GcpImagesStorage>(provider
                 => new GcpImagesStorage(provider.GetRequiredService<IGcpStorageContext>()));
         }
+
+        public static IServiceCollection AddGcpSynchronization(this IServiceCollection services, Microsoft.Extensions.Configuration.IConfiguration configuration)
```
h4wk13, line 52: xml-doc ?

OutOfSchool/OutOfSchool.WebApi/Services/Gcp/GcpImagesStorageSynchronizationService.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                logger.LogDebug("Gcp storage synchronization was finished");
+            }
+        }
+
+        private async Task<HashSet<string>> SynchronizeAll(HashSet<string> searchIds)
+        {
+            foreach (var syncFunction in GetAllSyncFunctions)
+            {
+                var syncResult = await syncFunction(searchIds).ConfigureAwait(false);
```
h4wk13, line 101: Looks pretty confusing, why do we have to wait `sync` function ?

OutOfSchool/OutOfSchool.WebApi/Util/FakeImplementations/FakeImagesStorage.cs
```diff
@@ -1,13 +1,26 @@
...
+using Google.Cloud.Storage.V1;
 using OutOfSchool.Services.Models.Images;
 using OutOfSchool.Services.Repository.Files;
+using Object = Google.Apis.Storage.v1.Data.Object;
 
 namespace OutOfSchool.WebApi.Util.FakeImplementations
 {
+    /// <summary>
+    /// Only for development purposes. Used as fake storage whenever no need to interplay with gcp storage.
+    /// </summary>
```
h4wk13, line 16: Please wrap this and such sources with the `#if DEBUG # endif` directive so they won't be present in the production code.

OutOfSchool/OutOfSchool.WebApi/Util/QuartzPool.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// </summary>
+        public static IReadOnlyList<Action<IServiceCollectionQuartzConfigurator>> GetQuartzConfigActions
+            => quartzConfigActions.AsReadOnly();
+
+        /// <summary>
+        /// Adds a job to the quartz pool.
+        /// </summary>
+        /// <param name="configure">Configures a given job.</param>
+        /// <typeparam name="T">Quartz job.</typeparam>
+        public static void AddJob<T>(Action<IJobConfigurator> configure = null)
```
h4wk13, line 26: In case that method will be called into multiple threads `quartzConfigActions` will be broken as the `List` is not thread safe collection.
You can use `ConcurrentBag<>` instead.

OutOfSchool/OutOfSchool.WebApi/Services/Gcp/GcpImagesStorageSynchronizationService.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        /// <inheritdoc/>
+        public async Task SynchronizeAsync(CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                logger.LogDebug("Gcp storage synchronization was started");
+
+                // gcp returns data by local time
+                var dateTime = DateTime.Now.AddHours(-1);
```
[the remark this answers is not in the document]
provicevko: -1 is needed to give all started processes oportunity to finish. Whenever we start this job and some requests are made by users, we can exclude the situation we create files in gcp but not save info about them in our main database because of 1 hour gap.
I've already tried to use UTC but gcp returns data using local server TZ so I guess we can use it? The problem can be if TZ is changed in the middle of synchronization that is unbelievable. If TZ is changed in server then another TZ will be chosen by gcp.
Maybe I've missed something in my thoughts.

OutOfSchool/OutOfSchool.WebApi/Services/Gcp/GcpImagesStorageSynchronizationService.cs
provicevko: `sync` here means synchronizeFunctions not synchronous. See previous comment about `await` if it must be synchronous in executing.

OutOfSchool/OutOfSchool.WebApi/Extensions/Startup/QuartzExtension.cs
```diff
@@ -0,0 +1,60 @@
...
+            this IServiceCollection services,
+            IConfiguration configuration)
+        {
+            _ = services ?? throw new ArgumentNullException(nameof(services));
+
+            services.AddQuartz(q =>
+            {
+                q.SchedulerId = "DefaultAppQuartz";
+                q.SchedulerName = "DefaultAppQuartz";
+                q.SetProperty("quartz.serializer.type", "json");
```
h4wk13, line 26: It seems you can read all the quartz configuration from the external config and use `SetProperty` for each.
It will bring mode flexibility to the server administrator, but it should be well documented then.
In case you want to restrict user to config everything - a way better to have a particular config model.

OutOfSchool/OutOfSchool.WebApi/Services/Gcp/GcpImagesStorageSynchronizationService.cs
h4wk13: That was a hint of renaming this `syncFunction` to something like a `synchronizationFunction` or to the name you already showed in your answer )
Also, if you agree with renaming - please rename property or field `GetAllSyncFunctions` as well

OutOfSchool/OutOfSchool.WebApi/Services/Gcp/GcpImagesStorageSynchronizationService.cs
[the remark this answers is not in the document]
VadymLevkovskyi: You've stepped into the usual problem with TZ with DST (day saving time). 
An issue might be when TZ is not changed, but your DST is switched on/off. I.e. in Ukraine you idea with -1 hour will not work twice a year at 3 AM last Sunday on March and October.

It's minor, but it's better to avoid it using UTC (and ask gcp about UTC time, if it is possible)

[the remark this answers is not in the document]
h4wk13: We can continue this discussion about `DateTime` and .NET for days or weeks or... ))
I will try to rephrase my question: 
1. why not -5 ? or -3.762056 ?
2. Also 'unbelievable' for me literally means: we know there is a bug which will bubble up once per few years and nobody knows/remembers what and why, and everyone will continue to suffer )

Also, just imagine how much efforts we will have to put here to identify all places which should be updated in case if GCP changed their free client library or data format or timing format ?
Better option, as for me, is to have a single place for converting input data to format that we will use in the whole application, so the only single place to update in case of any changes.
Does that make sense ?

OutOfSchool/OutOfSchool.DataAccess/Repository/Files/GcpImagesSyncDataRepository.cs
provicevko: True, I haven't even thought about it before

OutOfSchool/OutOfSchool.WebApi/Services/Gcp/GcpImagesStorageSynchronizationService.cs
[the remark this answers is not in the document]
provicevko: 1. **Agree with bouth of you, thanks.** I've just converted `DateTime` that I get from gcp to utc where I compare and filtering with UtcNow.
2. @h4wk13 About -5, -3.762056 I moved it into constants but I also want to determine what I mean. This isn't about time zone evaluating. It's about the situation when some user starts operation for example 'creating workshop' then some images will be uploaded into gcp storage but there are also few long time tasks before changes are applied to our database so without 'operation gap' can be moments when we delete files from gcp by synchronizer because they are missing in our db then after some time we commit changes into our db. So we have ids in our db which were deleted by synchronizer.
3. That's because we decided not to have a distributed transaction between our db and gcp **at least for now** and synchronizer can be triggered when users do their jobs (we don't close ports or anything else while syncronization).
Also that's why the trigger must be fired at night.

OutOfSchool/OutOfSchool.WebApi/Extensions/Startup/QuartzExtension.cs
provicevko: rewrote this

DmyMi (approved):

PR #658, opened 2022-06-06
Romanchuk Lukashov/Institution hierarchy (add new actions)

alrom:


OutOfSchool/OutOfSchool.IdentityServer/db-institutionshierarchyscript.txt
```diff
@@ -0,0 +1,1034 @@
+﻿USE out_of_school;
```
DmyMi, line 1: rename the extension to `.sql` and maybe move to `/Config` folder in repository root as it's not part of the identity server.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/SubordinationStructure/InstitutionHierarchyController.cs
```diff
@@ -14,7 +17,7 @@ namespace OutOfSchool.WebApi.Controllers.V1.SubordinationStructure
     [ApiController]
     [ApiVersion("1.0")]
     [Route("api/v{version:apiVersion}/[controller]/[action]")]
-    [HasPermission(Permissions.SystemManagement)]
+    [Authorize]
```
DmyMi, line 20: it's either we have a certain permission or leave it for anonymous because just having an account in the application is not secure enough :). It's ok if actions need different permission (get is anonymous, create is admin only, etc.)

OutOfSchool/OutOfSchool.WebApi/Extensions/ElasticsearchMappingExtensions.cs
```diff
@@ -20,7 +20,9 @@ public static WorkshopES ToESModel(this WorkshopDTO workshopDto)
                         dest => dest.Keywords,
                         opt =>
                             opt.MapFrom(src =>
-                                string.Join('¤', src.Keywords.Distinct())));
+                                string.Join('¤', src.Keywords.Distinct())))
```
DmyMi, line 23: i totally forgot about this file :) can you add a `TODO` somewhere that some tine later we need to create a mapping profile instead of extensions.

DmyMi (approved):

OutOfSchool/OutOfSchool.IdentityServer/db-institutionshierarchyscript.txt
Sandaljero: Ok ;)

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/SubordinationStructure/InstitutionHierarchyController.cs
Sandaljero: Ok ;)

OutOfSchool/OutOfSchool.WebApi/Extensions/ElasticsearchMappingExtensions.cs
Sandaljero: Ok ;)

PR #659, opened 2022-06-07
#859 social group multiselection

trdmitry:
* added ChildRepository
* added ChildrenSocialGroups
* removed ChildrensId from SocialGroupDTO


OutOfSchool/OutOfSchool.DataAccess/Repository/ChildRepository.cs
```diff
@@ -0,0 +1,26 @@
...
+        private readonly OutOfSchoolDbContext db;
+
+        public ChildRepository(OutOfSchoolDbContext dbContext)
+         : base(dbContext)
+        {
+            db = dbContext;
+        }
+
+        public override Task<Child> Create(Child child)
+        {
```
provicevko, line 21: You should check child on null before

provicevko (approved):

OutOfSchool/OutOfSchool.WebApi/Models/ChildDTO.cs
```diff
@@ -36,7 +37,7 @@ public class ChildDto
 
         public Guid ParentId { get; set; } = default;
 
-        public long? SocialGroupId { get; set; } = default;
+        public List<SocialGroupDto> SocialGroups { get; set; } = new List<SocialGroupDto>();
```
h4wk13, line 40: Such initializers telling me that this model can have different states when some of fields can be uninitialized and causes NRE in some parts of code.
Maybe we have to have more strict checks in the code to avoid such  initializers ?

trdmitry:
Comments in the new pull request.

PR #671, opened 2022-06-14
rbiyovskiy/653 kodeficator create db model and parser

RBiyovskiy:
Created only model and migration.
Project and scripts that generated by project will be send to Dmytro Minochkin.

### DIFF
--- a/OutOfSchool/OutOfSchool.DataAccess/Enums/CodeficatorCategory.cs
+++ b/OutOfSchool/OutOfSchool.DataAccess/Enums/CodeficatorCategory.cs
@@ -1,0 +1,23 @@
+﻿namespace OutOfSchool.Services.Enums
+{
+    public static class CodeficatorCategory
+    {
+        public const string Region = "O"; // Автономна Республіка Крим, області.
+
+        public const string SpecialStatusCity = "K"; // Міста, що мають спеціальний статус.
+
+        public const string District = "P"; // Райони в областях та Автономній Республіці Крим.
+
+        public const string TerritorialCommunity = "H"; // Території територіальних громад (назви територіальних громад) в областях, територіальні громади Автономної Республіки Крим.
+
+        public const string City = "M"; // Міста.
+
+        public const string UrbanSettlement = "T"; // Селища міського типу.
+
+        public const string Village = "C"; // Села.
+
+        public const string Settlement = "X"; // Селища.
+
+        public const string CityDistrict = "B";  // Райони в містах.
+    }
+}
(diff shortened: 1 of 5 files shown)


OutOfSchool/OutOfSchool.DataAccess/Enums/CodeficatorCategory.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace OutOfSchool.Services.Enums
+{
+    public static class CodeficatorCategory
```
DmyMi, line 3: it will be hard to check what are the allowed values for category. Maybe think about custom enumaration and/or custom value converter for entity framework, or some other way we can have strong type while saving string to database.

DmyMi (approved):

OutOfSchool/OutOfSchool.DataAccess/Enums/CodeficatorCategory.cs
RBiyovskiy: Why, will it be hard? You will enter the name of the CodeficatorCategory class and see all possible values.
I did not find another way.
I wanted to save in a DB string value of the category, instead of number. But we can't use a string in the enum. If we need to list categories, we can add a method that will return a collection of possible values ​​(for example, for use in DDL).

provicevko: 1. You could use a enum with attributes and save in database numbers then using extension methods easily convert numbers->strings or vice versa strings->numbers on the server side if you need that. You can check _WebApi/Common/Resources/Codes/ImagesOperationErrorCode_ as example.
2. I'm not sure about this solution to this case but also exists a package for `SmartEnum` that allows create things very similar to enums but with parameters (like in Java, Scala etc.)

provicevko: Or even if you wanna save a string to the database, enum with attribute can do it I guess but more costly.

RBiyovskiy: Yes, of cours, it is very costly using enum for saving data in a string format. 
My way is the regular way for this, I think. For example the Debug is using similar construction. Debug.Info, Debug.Error, Debug.Warning. 
I repeat I wanted to have Category in a string format not a number. 
But if you all suggest that is not a good idea I will change it. In this case do not use scripts before I do not made changes.

provicevko: I don't see your usages so it's just some offers how to use them with strings, I do not insist on immediate changes.

RBiyovskiy: Yeah, I understand. It is a discussion. And I try to explain why I made it.

Issue #479, opened 2022-01-03
Area admin creates service provider (director) 5.3.5 


OlhaYuzefiv:
As a user (area-admin) I want to create a service provider (director) to be able to manage the workshop's data and information

Acceptance Criteria

The area-admin creates a service provider

GIVEN Area admin opens the cabinet

AND clicks the tab "Надавачі послуг"

(Design in progress)

AND clicks the button “Додати надавача послуг”

AND service provider (director) creation page opens
![додати надавача](https://user-images.githubusercontent.com/89577816/147986670-ef959113-e1ef-495b-b82b-611b3714f7d7.png)

AND fills in all the data
WHEN clicks “Зберегти”
THEN new service provider (director) is created
AND shown in the list

(Design in progress)

Epic link
E.g.: Epic #100 Epic

Labels to be added
"User story", Correspondind Functionality (e.g. User), Priority ("pri: "), "Corresponding Epic"

Issue #348, opened 2021-09-24
Ministry admin confirms a service provider registration (Paid/free Service provider approval, license approval)


ulianachukhna:
**As a** user (ministry administrator) **I want to** confirm the service provider **so that** to manage workshops data and information 

Role 
user (ministry admin) 

**Acceptance Criteria** 

Scenario #1 

GIVEN Ministry admin opens the cabinet 
AND clicks the tab “Надавачі послуг” 

<img width="1440" alt="клік на надивачі" src="https://user-images.githubusercontent.com/88023871/134671359-507afdc5-ef00-4dee-9b9f-b4fd4fef8c98.png">

AND clicks on interested Service providers request with the status “Очікує підтвердження” 
WHEN clicks “Підтвердити” 
 
<img width="1440" alt="підтвердити" src="https://user-images.githubusercontent.com/88023871/134671481-d6fc744e-ebd4-4eab-839f-3ec987b706f0.png">

AND information about the Service provider is displayed 

<img width="238" alt="інфо" src="https://user-images.githubusercontent.com/88023871/134671525-99c9b51e-b801-4943-aadf-fe6f860959e4.png">
 
AND checks all the data 
AND all data is correct 
WHEN clicks “Підтвердити” 
THEN new Service provider is approved 
AND New Service provider can see a notification 

<img width="157" alt="notification" src="https://user-images.githubusercontent.com/88023871/134672476-c7dab299-2ebb-41e0-897b-ec50d2b9f4ee.png">

AND New Service provider can see in the cabinet message 

<img width="432" alt="профайл погоджено" src="https://user-images.githubusercontent.com/88023871/134672524-374d8bc6-aeaf-42b6-8d6f-27fad4935a4c.png">

**As a** (ministry admin) **I want to** send back new paid Service provider application to check and edit information **so that** to manage and keep information about the workshop right 

**Acceptance Criteria** 

SCENARIO #2 
(Ministry admin) checks created New Service provider and noticed incorrect information and changes status of the New Service Provider to “На редагуванні” so that New Service provider could double-check and edit profiles information. 

GIVEN Ministry admin opens the cabinet 
AND clicks the tab “Надавачі послуг” 

<img width="1440" alt="клік на надивачі" src="https://user-images.githubusercontent.com/88023871/134672690-0746083e-5561-496f-99fe-4c689f973183.png">

AND clicks on interested Service providers request with the status “Очікує підтвердження” 
AND information about the Service provider is displayed 

[...]

Issue #264, opened 2021-08-06
Paid Service provider with paid workshops


Nikita-Babenko:
As a user I want to create Paid Service provider so that to be able to create paid workshops and get paid members

**Role**
user (paid service provider) , tech admin

### Acceptance Criteria
SCENARIO #1 User registers New Service provider 
 
GIVEN New Service provider registers profile
![1ічфчф](https://user-images.githubusercontent.com/12609814/128507234-5c11c25f-7c89-4a09-b359-0175325b784b.png)
AND selects  “Приватна”
AND additional non required drop-down field “Завантажити ліцензію” displays under the field “Приватна”
AND inputs all needed data 
WHEN clicks “Зберегти” on the final registration page
THEN new Service provider is created
AND message in the cabinet of the new Service provider displayed
![2021-08-03 17_16_47-ПІдтвердити реєстрацію (надіслати на підтвердження) png ‎- Photos](https://user-images.githubusercontent.com/12609814/128507318-1bb35e52-c126-4089-8045-e6611c8be60c.png)
AND approval request is sent to district admin
AND Service provider waits until (district admin/ministry admin/tech admin)  approves the profile 
 

As (tech admin) I want to approve Paid Service provider so that to manage and keep information about the workshop right

**Role**
user (paid service provider) , tech admin

### Acceptance Criteria
SCENARIO #2 Tech admin approves new Service providers profile 

GIVEN tech admin opens cabinet
AND can see the notification
![2021-08-03 17_56_00-MoE _ Dev Team – Figma](https://user-images.githubusercontent.com/12609814/128507371-a45e2cd3-94e0-49aa-8cc3-a919135e72e3.png)
AND selects tab  “Надавачі послуг”
AND a list of Service providers displayed
![7777 Підтвердити профіль](https://user-images.githubusercontent.com/12609814/128507408-0592cacd-3f20-4689-9e86-2561fe40c04f.png)
AND tech admin wants to check information about the new Service provider
AND selects the Service provider
AND clicks on interested Service providers request with the status “Очікує підтвердження”
AND information about the Service provider is displayed 
![7777 Підтвердити профіль (Відправити на редагування)](https://user-images.githubusercontent.com/12609814/128507665-15d4f451-6ee3-4e97-82a8-737124a35e67.png)
AND tech admin wants to check information about the new Service provider
AND selects the Service provider
AND clicks on interested Service providers request with the status “Очікує підтвердження”
AND information about the Service provider is displayed 
[...]

Issue #257, opened 2021-08-02
Tech-admin approves the registration of service provider 


ulianachukhna:
**As a** user (tech-admin) **I want** to approve the registration of service provider **so that** to be in the list of service providers 

**Acceptance Criteria**
Tech-admin approves a service provider 

GIVEN Tech admin opens the cabinet 

AND sees notification sign on the tab “Надавачі послуг” 

<img width="596" alt="нотифікація" src="https://user-images.githubusercontent.com/88023871/127832374-cadedaad-a155-4535-9242-446c49188182.png">

AND clicks the tab “Надавачі послуг” 

AND sees the list of registered service providers 
<img width="912" alt="підтвердити надавача" src="https://user-images.githubusercontent.com/88023871/127832479-bd63b1be-3428-47f6-9990-5fd634571596.png">
AND clicks the right service provider 

AND clicks the tab “Підтвердити реєстрацію” 

AND service provider is registered 

THEN tech-admin sees the status “Підтверджено” 

<img width="93" alt="підтверджено" src="https://user-images.githubusercontent.com/88023871/127833270-d54294b0-b4f6-4850-aba1-14c765bf00f6.png">

AND service provider is added to the list of service providers 

**Epic link**
E.g.: Epic #100 [Epic](https://github.com/ita-social-projects/OoS-Backend/labels/100)

**Labels to be added**
"User story", Correspondind Functionality (e.g. User), Priority ("pri: "), "Corresponding Epic"

### Tasks 
1. - [ ] sample task.

PR #677, opened 2022-06-16
Service provider approval

i-rakityanskiy:
### (!) This PR contains braking changes in `Provider/Get` `v1` and `v2` endpoints that must be applied at front-end. 

https://github.com/ita-social-projects/OoS-Backend/issues/257
**https://github.com/ita-social-projects/OoS-Backend/issues/264**
https://github.com/ita-social-projects/OoS-Backend/issues/348
https://github.com/ita-social-projects/OoS-Backend/issues/479

In scope of this PR was not implemented sending notifications to District Admin

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
@@ -258,7 +258,7 @@ namespace OutOfSchool.WebApi.Services
             var users = await usersRepository.GetByFilter(u => u.Id.Equals(providerDto.UserId)).ConfigureAwait(false);
             providerDomainModel.User = users.Single();
             providerDomainModel.User.IsRegistered = true;
-            providerDomainModel.Status = ProviderApprovalStatus.Pending;
+            providerDomainModel.Status = ProviderStatus.Pending;
             providerDomainModel.LicenseStatus = providerDomainModel.License == null
                 ? ProviderLicenseStatus.NotProvided
                 : ProviderLicenseStatus.Pending;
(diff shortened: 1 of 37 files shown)


OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
```diff
@@ -178,6 +181,96 @@ public async Task<ProviderDto> Update(ProviderDto providerDto, string userId)
         public async Task<Guid> GetProviderIdForWorkshopById(Guid workshopId) =>
             await workshopServiceCombiner.GetWorkshopProviderId(workshopId).ConfigureAwait(false);
 
+        public async Task<ProviderStatusDto> UpdateStatus(ProviderStatusDto dto, string userId)
```
DmyMi, line 184: not a call to action, but userId is used only for logging purposes? If it's going to be used later to verify if the given admin can or can not update this status - maybe add a TODO so we won't forget :)

OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
```diff
@@ -178,6 +181,96 @@ public async Task<ProviderDto> Update(ProviderDto providerDto, string userId)
...
+        {
+            _ = dto ?? throw new ArgumentNullException(nameof(dto));
+
+            var provider = await providerRepository.GetById(dto.ProviderId).ConfigureAwait(false);
+
+            if (provider is null)
+            {
+                logger.LogError($"Provider(id) {dto.ProviderId} not found. User(id): {userId}");
```
DmyMi, line 192: we do like to use error logs throughout the application :) but here probably nothing bad has happened, it might be just an user error or someone deleted the provider before admin could click approve. As user experience or application functioning is not impacted - maybe use `info` or `warning`?

OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
```diff
@@ -178,6 +181,96 @@ public async Task<ProviderDto> Update(ProviderDto providerDto, string userId)
...
+
+                return null;
+            }
+
+            if (string.IsNullOrEmpty(provider.License) && dto.LicenseStatus != ProviderLicenseStatus.NotProvided)
+            {
+                throw new ArgumentException("Provider license is not provided. It cannot be approved.");
```
DmyMi, line 222: here is a good place for log, though it's an expected part of the application flow, so info might be enough.

DmyMi (approved):

PR #681, opened 2022-06-17
Romanchuk/Add Redis to institution hierarchy

alrom:
Add Redis to InstitutionHierarchy services:
- InstitutionFieldDescriptionService.GetByInstitutionId
- InstitutionHierarchyService.GetChildren
- InstitutionHierarchyService.GetParents
- InstitutionHierarchyService.GetAllByInstitutionAndLevel
- InstitutionService.GetAll

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Services/SubordinationStructure/InstitutionHierarchyService.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Services/SubordinationStructure/InstitutionHierarchyService.cs
@@ -6,6 +6,7 @@ using AutoMapper;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Localization;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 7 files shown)


OutOfSchool/OutOfSchool.WebApi/Services/SubordinationStructure/InstitutionHierarchyService.cs
```diff
@@ -102,6 +107,17 @@ public async Task<List<InstitutionHierarchyDto>> GetChildren(Guid? parentId)
         {
             logger.LogInformation("Getting all children InstitutionHierarchies started.");
 
+            string cacheKey = $"InstitutionHierarchyService_GetChildren_{parentId}";
```
provicevko, line 110: Maybe we will need to save such cacheKey templates in one place and then use `string.Format` for retrieving string we want.

provicevko (approved):

PR #684, opened 2022-06-20
rbiyovskiy/680 implement codeficator service

RBiyovskiy:
Implemented:
- get list of regions (when id is null);
- get children entities by parentId;
- get list of the regions  as pair values: Id, Name  (when id is null);
- get list of the regions  as pair values: Id, Name by parentId;
- get parents by entityId (ierarchical structure).


OutOfSchool/OutOfSchool.WebApi/Models/Codeficator/CodeficatorDto.cs
```diff
@@ -0,0 +1,28 @@
...
+        [MaxLength(30)]
+        public string Name { get; set; }
+
+        public double Latitude { get; set; }
+
+        public double Longitude { get; set; }
+
+        public ulong GeoHash { get; set; }
+
+        public bool NeedCheck { get; set; }
```
DmyMi, line 26: same as geoshash

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Models/Codeficator/CodeficatorDto.cs
RBiyovskiy: fixed all

PR #695, opened 2022-06-23
update logging

DmyMi:
For future production we already have an Elasticsearch so we can use it as a log storage & analysis tool.

First iteration of using ES:
- add formatter
- add app version to logs so can filter them
- add exception properties for easier serialization to json

For testing, at first this formatter will be used to write to file in development env.

### DIFF
--- a/OutOfSchool/OutOfSchool.IdentityServer/appsettings.json
+++ b/OutOfSchool/OutOfSchool.IdentityServer/appsettings.json
@@ -10,9 +10,12 @@
       "GuidFormat": "Binary16"
     }
   },
+  "AppDefaults": {
+    "Version": "replace_me"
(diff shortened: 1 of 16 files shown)


OutOfSchool/OutOfSchool.IdentityServer/appsettings.json
```diff
@@ -10,9 +10,12 @@
       "GuidFormat": "Binary16"
     }
   },
+  "AppDefaults": {
+    "Version": "replace_me"
```
provicevko, line 14: `"Version": "replace_me"` 😀 here and below

provicevko (approved):

RBiyovskiy (approved):

PR #703, opened 2022-06-24
rbiyovskiy/689 we receive provider subrole in token it could be either deputy pr provideradmin

RBiyovskiy:
- added action ManagedWorkshops to the ProviderAdminController;
- added action GetByProviderAdminId to the ApplicationController;
- fixed validation message and regex for the phone number in time ProviderAdmin registration;
- added explanetion message in time ProviderAdmin registration (if user with the same email already exists);
- fixed two ApplicationController tests.

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
@@ -153,10 +153,10 @@ namespace OutOfSchool.WebApi.Controllers.V1
         }
 
         /// <summary>
-        /// Get Applications by Provider or Workshop Id.
+        /// Get Applications by Provider, Workshop or ProviderAdmin Id.
         /// </summary>
-        /// <param name="property">Property to find by (workshop or provider).</param>
-        /// <param name="id">Provider or Workshop id.</param>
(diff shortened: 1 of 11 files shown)


OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -366,32 +375,41 @@ private async Task<SearchResult<ApplicationDto>> GetByProviderId(Guid id, Applic
...
+            var applications = await applicationService.GetAllByProviderAdmin(userId, filter, providerAdmin.ProviderId, providerAdmin.IsDeputy).ConfigureAwait(false);
+
+            return applications;
+        }
+
+        private async Task CheckUserRights(Guid parentId = default, Guid providerId = default, Guid workshopId = default, string providerAdminId = default)
         {
             var userId = GettingUserProperties.GetUserId(User);
 
             bool userHasRights = true;
```
DmyMi, line 398: not your code, but maybe the default value should be false, it's better to deny access by default in case something goes wrong. Also maybe it's better to refactor the exception flow in case of `userHasRights` is `false` to return a `403` code instead of a more general bad request as it is now.

DmyMi (approved):

PR #705, opened 2022-06-26
Romanchuk/Add pagination and filter

alrom:
Add pagination and filter to:
- Providers;
- Parents.

Add multiordering to EntityRepositoryBase.


OutOfSchool/OutOfSchool.WebApi/Services/ParentService.cs
```diff
@@ -86,6 +87,37 @@ public async Task<IEnumerable<ParentDTO>> GetAll()
...
+            logger.LogInformation("Getting all Parents started (by filter).");
+
+            if (filter is null)
+            {
+                filter = new SearchStringFilter();
+            }
+
+            var filterPredicate = PredicateBuild(filter);
+
+            int count = await repositoryParent.Count(filterPredicate).ConfigureAwait(false);
```
h4wk13, line 102: Why do we need 2 calls to the data source ?
Correct me if I'm wrong but the code below gets the collection of items from the data source by the same filter.
Is that the same count ?
Is it possible to get this count from items in the collection ?

OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
```diff
@@ -96,21 +95,43 @@ public async Task<IEnumerable<ProviderDto>> GetAll()
...
-                    provider.Rating = rating;
-                    provider.NumberOfRatings = numberOfVotes;
-                }
+                filter = new SearchStringFilter();
             }
 
-            return providersDTO;
+            var filterPredicate = PredicateBuild(filter);
+
+            int count = await providerRepository.Count(filterPredicate).ConfigureAwait(false);
```
h4wk13, line 115: same question about count

OutOfSchool/OutOfSchool.WebApi/Services/ParentService.cs
```diff
@@ -147,5 +179,29 @@ public async Task<ShortUserDto> Update(ShortUserDto dto)
...
+            var predicate = PredicateBuilder.True<Parent>();
+
+            if (!string.IsNullOrWhiteSpace(filter.SearchString))
+            {
+                var tempPredicate = PredicateBuilder.False<Parent>();
+
+                foreach (var word in filter.SearchString.Split(' ', ',', StringSplitOptions.RemoveEmptyEntries))
+                {
+                    tempPredicate = tempPredicate.Or(
+                        x => x.User.FirstName.StartsWith(word, StringComparison.InvariantCulture)
```
h4wk13, line 194: You can add the extension method to the string type, smth like a InvariantStartWith

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ParentController.cs
```diff
@@ -62,6 +62,31 @@ public async Task<IActionResult> Get()
...
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        public async Task<IActionResult> GetByFilter([FromQuery] SearchStringFilter filter)
+{
+            var parents = await serviceParent.GetByFilter(filter).ConfigureAwait(false);
+
+            if (parents.TotalAmount < 1)
+            {
+                return NoContent();
```
DmyMi, line 84: not exactly comment for you to fix everywhere, but a TODO/ to ask, as frontend mentioned it would be cool if the result is empty - return object with `TotalAmount` 0 and empty `Entities` list instead of `NoContent` as they need less checks for pagination.

OutOfSchool/OutOfSchool.WebApi/Services/ApplicationService.cs
```diff
@@ -207,21 +208,28 @@ public async Task<SearchResult<ApplicationDto>> GetAllByParent(Guid id, Applicat
...
+
+            var totalAmount = await applicationRepository.Count(where: predicate).ConfigureAwait(false);
 
-            var applications = applicationRepository.Get<int>(skip: filter.From, take: filter.Size, where: applicationFilter, includeProperties: "Workshop,Child,Parent");
+            var applications = await applicationRepository.Get<Application>(
+                skip: filter.From,
+                take: filter.Size,
+                where: predicate,
+                includeProperties: "Workshop,Child,Parent",
+                orderBy: sortPredicate).ToListAsync().ConfigureAwait(false);
```
DmyMi, line 223: not a call to immediate action, but as per today's discussion maybe these 3 methods you've rewrote are good canditates for `asNoTracking` flag

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
```diff
@@ -155,14 +149,18 @@ public async Task<SearchResult<WorkshopDTO>> GetAll(OffsetFilter offsetFilter)
...
             var count = await workshopRepository.Count().ConfigureAwait(false);
             var workshops =
-                workshopRepository.Get(
+                workshopRepository.Get<Workshop>(
                     skip: offsetFilter.From,
                     take: offsetFilter.Size,
                     includeProperties: includingPropertiesForMappingDtoModel,
-                    orderBy: x => x.Id,
-                    ascending: true)
+                    orderBy: sortExpression)
```
DmyMi, line 163: `asNoTracking` might be good here too.

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ParentController.cs
alrom: I did it in Parent and Provider controllers GetByFilter methods. Wrote to Sasha about it.

OutOfSchool/OutOfSchool.WebApi/Services/ApplicationService.cs
alrom: I implemented it to Application, Parent and Provider controllers. Lets look at the result ))

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
alrom: Done

OutOfSchool/OutOfSchool.WebApi/Services/ParentService.cs
alrom: I can't do it right now. There is an exception when i use extension method. But I'll try to find the solution for this.

OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
```diff
@@ -358,5 +379,46 @@ private void LogProviderChanges(Provider provider, string userId)
...
+        private Expression<Func<Provider, bool>> PredicateBuild(SearchStringFilter filter)
+        {
+            var predicate = PredicateBuilder.True<Provider>();
+
+            if (!string.IsNullOrWhiteSpace(filter.SearchString))
+            {
+                var tempPredicate = PredicateBuilder.False<Provider>();
+
+                foreach (var word in filter.SearchString.Split(' ', ',', StringSplitOptions.RemoveEmptyEntries))
+                {
```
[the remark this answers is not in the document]
alrom: They are not identical in the properties Phone and Email. And in the future they might be more different. I don't know by which fields front will make theirs search queries.

OutOfSchool/OutOfSchool.WebApi/Services/ParentService.cs
h4wk13: Could you please show a small code example when the exception is thrown ?

PR #713, opened 2022-06-29
fix quartz extension to work with cloud and less hardcode

DmyMi:


OutOfSchool/OutOfSchool.WebApi/Startup.cs
```diff
@@ -358,7 +358,10 @@ public void ConfigureServices(IServiceCollection services)
             services.AddAutoMapper(typeof(MappingProfile));
 
             services.AddSignalR();
-            services.AddDefaultQuartz(Configuration);
+
+            var quartzConfig = Configuration.GetSection(QuartzConfig.Name).Get<QuartzConfig>();
+            services.AddDefaultQuartz(Configuration, quartzConfig.ConnectionStringKey);
+            services.AddGcpSynchronization(quartzConfig);
```
provicevko, line 364: Looks like this will not work because `services.AddDefaultQuartz(Configuration, quartzConfig.ConnectionStringKey)` method uses `QuartzPool.QuartzConfigActions` to setup all jobs&triggers. In this case you'll add a job&triggers for GcpSynchronization after quartz's setupped so they will not be registered. I mean that `services.AddDefaultQuartz(Configuration, quartzConfig.ConnectionStringKey)` must be after `services.AddGcpSynchronization(quartzConfig)`

provicevko: Or we must reconsider our way of registering jobs&triggers to Quartz in another way for ElasticSearch, Gcp, ... to be order independent.

PR #716, opened 2022-06-30
GetChildrenByWorkshopId

maksgritchin:


OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ChildControllerTests.cs
```diff
@@ -33,7 +35,9 @@ public class ChildControllerTests
         public void Setup()
         {
             service = new Mock<IChildService>();
-            controller = new ChildController(service.Object);
+            providerService = new Mock<IProviderService>();
```
VadymLevkovskyi, line 38: `providerService` and `providerAdminService` might be done local method variables, not class fields. Encapsulation might be improved

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ChildControllerTests.cs
```diff
@@ -23,6 +23,8 @@ public class ChildControllerTests
     {
```
VadymLevkovskyi, line 23: Please add constructor tests. (Assert that it fails with null passed as any service parameter)

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChildController.cs
```diff
@@ -101,6 +108,59 @@ public async Task<IActionResult> GetUsersChildById(Guid id)
...
+        /// <param name="offsetFilter">Filter to get a part of all children that were found.</param>
+        /// <returns>The result is a <see cref="SearchResult{ChildDto}"/> that contains the count of all found children and a list of children that were received.</returns>
+        [HasPermission(Permissions.WorkshopEdit)]
+        [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(SearchResult<ChildDto>))]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        [HttpGet("{workshopId}")]
+        public async Task<IActionResult> GetApprovedByWorkshopId(Guid workshopId, [FromQuery] OffsetFilter offsetFilter)
```
VadymLevkovskyi, line 124: I believe you should add test(s) to cover code of added new public method

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChildController.cs
```diff
@@ -101,6 +108,59 @@ public async Task<IActionResult> GetUsersChildById(Guid id)
...
+            var userHasRights = await this.IsUserProvidersOwnerOrAdmin(workshopId).ConfigureAwait(false);
+            if (!userHasRights)
+            {
+                return StatusCode(403, "Forbidden for another providers.");
+            }
+
+            return Ok(await service.GetApprovedByWorkshopId(workshopId, offsetFilter).ConfigureAwait(false));
+        }
+
+        private async Task<bool> IsUserProvidersOwnerOrAdmin(Guid workshopId)
```
VadymLevkovskyi, line 135: (Code style readability concern) 
1. If `User` is not in `provider` role, you return `true` for method named `IsUserProvidersOwnerOrAdmin`. I'm bit concerned. Either change method name or store result of check (`User.IsInRole(nameof(Role.Provider).ToLower())`) into variable with meaningful name
2. If `User` is not in `provider` role, then there is no sense to call `providerService.GetProviderIdForWorkshopById()`, you are not using its result for this case. So, you could call it later inside `try`/`catch` after `providerService.GetByUserId()`
3. You have single `return true` after `if (User.IsInRole(...)) { ... }` with more than 20 lines of code. Isn't it better to invert condition in mentioned `if` and have small one-line `return true`? You'll definitely improve your code readability.
4. Code like 
```
if (providerId != provider.Id)
{
    return false;
} 
```
with subsequent `return true` looks like long way to write simple `return providerId == provider.Id;`
5. Same for `if (!isUserRelatedAdmin)`
6. Inconsistent usage of `.ConfigureAwait(false)`, it is not used for `providerService.GetProviderIdForWorkshopById()`. Is there any reason?

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChildController.cs
```diff
@@ -101,6 +108,59 @@ public async Task<IActionResult> GetUsersChildById(Guid id)
...
+        {
+            Guid providerId = await providerService.GetProviderIdForWorkshopById(workshopId);
+
+            if (User.IsInRole(nameof(Role.Provider).ToLower()))
+            {
+                var userId = User.FindFirst("sub")?.Value;
```
VadymLevkovskyi, line 141: "sub" - is a string constant, that is probably used not only here. Please extract it or surrounding code (i.e. `User.FindFirst("sub")?.Value`) into some "entity" (const/helper method/extension method/your choice).

OutOfSchool/OutOfSchool.WebApi/Services/ChildService.cs
```diff
@@ -26,10 +29,11 @@ public class ChildService : IChildService
         /// <param name="childRepository">Repository for the Child entity.</param>
         /// <param name="parentRepository">Repository for the Parent entity.</param>
         /// <param name="logger">Logger.</param>
-        public ChildService(IEntityRepository<Child> childRepository, IParentRepository parentRepository, ILogger<ChildService> logger)
+        public ChildService(IEntityRepository<Child> childRepository, IParentRepository parentRepository, IApplicationRepository applicationRepository, ILogger<ChildService> logger)
```
VadymLevkovskyi, line 32: Class public API was changed (new constructor parameter and new method added), thus tests should be updated as well

OutOfSchool/OutOfSchool.WebApi/Services/ChildService.cs
```diff
@@ -160,6 +164,49 @@ public async Task<SearchResult<ChildDto>> GetByUserId(string userId, OffsetFilte
...
+        /// <exception cref="ArgumentException">If one of the parameters was not valid.</exception>
+        /// <exception cref="SqlException">If the database cannot execute the query.</exception>
+        public async Task<SearchResult<ChildDto>> GetApprovedByWorkshopId(Guid workshopId, OffsetFilter offsetFilter)
+        {
+            ValidateOffsetFilter(offsetFilter);
+
+            logger.LogDebug($"Getting Children by WorkshopId: {workshopId} started. Amount of children to take: {offsetFilter.Size}, skip first: {offsetFilter.From}.");
+
+            var applications = await applicationRepository.GetByFilter(p => p.WorkshopId == workshopId && p.Status == ApplicationStatus.Approved).ConfigureAwait(false);
+            HashSet<Guid> childrenGuids = new HashSet<Guid>();
```
VadymLevkovskyi, line 184: As we are not payed for code lines count next code 
```
HashSet<Guid> childrenGuids = new HashSet<Guid>();
foreach (var app in applications)
{
    childrenGuids.Add(app.ChildId);
}
```
could (I believe) be rewritten into
```
var childrenGuids = new HashSet<Guid>(applications.Select(app => app.ChildId));
```

VadymLevkovskyi (commented):
Code style and logic concerns. Missing unit tests. See comments

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChildController.cs
alrom: We have a module for it: OutOfSchool.WebApi.Common.GettingUserProperties

OutOfSchool/OutOfSchool.WebApi/Services/ChildService.cs
maksgritchin: It will be done in scope of another task

VadymLevkovskyi (approved):
I've expected to have more assert in success path test, also have fail path tests, workshop dto/user claims generator... but this is probably out of this task scope and I don't want to bother developer any longer. 
Approved

PR #720, opened 2022-07-01
Migrate to ASP.NET Core 6.0

i-rakityanskiy:


DmyMi (approved):

OutOfSchool/Tests/OutOfSchool.Tests.Common/TestDataGenerators/AddressGenerator.cs
```diff
@@ -11,7 +11,7 @@ namespace OutOfSchool.Tests.Common.TestDataGenerators
     public static class AddressGenerator
     {
         private static readonly Faker<Address> faker = new Faker<Address>()
-            .RuleFor(x => x.Id, f => f.IndexFaker)
+            .RuleFor(x => x.Id, f => f.IndexFaker++)
```
provicevko, line 14: For what reason?

provicevko (approved):

OutOfSchool/Tests/OutOfSchool.Tests.Common/TestDataGenerators/AddressGenerator.cs
i-rakityanskiy: When the IndexFaker is used to generate Address.Id it causes an exception `System.InvalidOperationException : The instance of entity type 'Address' cannot be tracked because another instance with the key value '{Id: 1}' is already being tracked. When attaching existing entities, ensure that only one entity instance with a given key value is attached.`
I saw a workaround to avoid this exception here https://github.com/ita-social-projects/OoS-Backend/blob/c906c6081c2d9711aa81b6263f5be9f175eea6f2/OutOfSchool/Tests/OutOfSchool.WebApi.IntegrationTests/ProviderServiceIntergrationTests/ProviderServiceUpdate.cs#L43
Using `f.IndexFaker++` solves this.

h4wk13 (approved):

OutOfSchool/Tests/OutOfSchool.Tests.Common/TestDataGenerators/AddressGenerator.cs
h4wk13: It is not really good idea to change the generator trying to fix the problem with data.
Seems it was used in a wrong way.
It is better to find the reason why collection already contains the address with id 1

i-rakityanskiy: This happens because IndexFaker has an initial value of 0. When entity with Id=0 is added to the DB context, EF implicitly changes Id value to 1. When the second entity with Id=1 is generated and added to the DB context it causes the exception. I think that for the testing purposes that fix is sufficient enough. At the same time I agree that better solution can be found.

i-rakityanskiy: @h4wk13 Please check this PR https://github.com/ita-social-projects/OoS-Backend/pull/733

PR #728, opened 2022-07-05
update CI/CD to work with .Net 6

DmyMi:


cloudbuild-auth.yml
```diff
@@ -6,10 +6,10 @@ steps:
     - build
     - gcr.io/$PROJECT_ID/oos-auth:$SHORT_SHA
     - --cache-image=gcr.io/$PROJECT_ID/oos-auth:cache
-    # TODO: after paketo-buildpacks/dotnet-core buildpack fix revert to using latest version
     # - --buildpack=gcr.io/paketo-buildpacks/dotnet-core
-    - --buildpack=docker:///gcr.io/paketo-buildpacks/dotnet-core:0.16.1
-    - --builder=paketobuildpacks/builder:base
+    - --buildpack=docker:///gcr.io/paketo-buildpacks/dotnet-core:0.11.3
```
VadymLevkovskyi, line 10: Just wondering, what was the reason to decrease version?

VadymLevkovskyi (approved):

cloudbuild-auth.yml
DmyMi: @VadymLevkovskyi  I will add this as a coment so we won't forget. There is a [bug in later versions of .Net 6 SDK 6.0.200 or higher](https://github.com/dotnet/sdk/issues/23461) (or rather Nuget toolset). It works fine if you are using `MSBuild`, but sometimes it fails to restore packages if you are using `dotnet` CLI.
Both Nuget team (hopefully) and Paketo Buildpack team ([issue](https://github.com/paketo-buildpacks/dotnet-core/issues/670)) are working on a permanent solution, but till then there are 2 workarounds - either use older build tools version does not have this bug yet or compile the code outside of container build tools (which still means the bug could manifest on Linux). By trial & error i've found a stable combination of older build tools.

VadymLevkovskyi: @DmyMi thanks, it's really important piece of knowledge!

PR #737, opened 2022-07-07
rbiyovskiy/736 codeficator improvements

RBiyovskiy:
Search Codeficators by part of name action changed

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Models/Codeficator/CodeficatorFilter.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Models/Codeficator/CodeficatorFilter.cs
@@ -1,6 +1,16 @@
+﻿using System.ComponentModel.DataAnnotations;
+﻿using OutOfSchool.Services.Enums;
+
 ﻿namespace OutOfSchool.WebApi.Models.Codeficator;
 
-public class CodeficatorFilter : SizeFilter
+public class CodeficatorFilter
 {
+    [Required]
+    [MinLength(3)]
     public string Name { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Gets or sets сategories. By default - 'MTCXK'".
+    /// </summary>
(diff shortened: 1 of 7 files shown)


OutOfSchool/OutOfSchool.WebApi/Models/Codeficator/CodeficatorFilter.cs
```diff
@@ -1,6 +1,16 @@
+﻿using System.ComponentModel.DataAnnotations;
+﻿using OutOfSchool.Services.Enums;
+
 ﻿namespace OutOfSchool.WebApi.Models.Codeficator;
 
-public class CodeficatorFilter : SizeFilter
+public class CodeficatorFilter
 {
+    [Required]
+    [MinLength(3)]
```
DmyMi, line 9: just a question, maybe check length in logic and return `204` status instead of failed model validation if less than 3 characters. Frontend will use it in a "autocomplete" style search field and will need to check if the `400` they got is because they sent less than 3, or because the request failed by any other reason.

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Models/Codeficator/CodeficatorFilter.cs
RBiyovskiy: I am not sure that it is correct return No Content if we will not execute request. Like for me it can confused. 
Maybe will be better ask at frontend devs, what will be better fot them?
I will ask.

RBiyovskiy: Asked by @Oleksandra Lytvynets. She agrees that returning badRequest is clearer. I looked at how it currently works for a city (asked by @Oleksandra Lytvynets) and it also returned a badRequest when there were not enough characters.

PR #743, opened 2022-07-07
Added usage SkiaSharp

provicevko:
`System.Drawing.Image.FromStream` is not supported by default for cross platform in .Net 6


OutOfSchool/OutOfSchool.WebApi/OutOfSchool.WebApi.csproj
```diff
@@ -59,6 +59,7 @@
     <PackageReference Include="Serilog.Settings.Configuration" Version="3.3.0" />
     <PackageReference Include="Serilog.Sinks.File" Version="5.0.0" />
     <PackageReference Include="Serilog.Sinks.GoogleCloudLogging" Version="5.0.0" />
+    <PackageReference Include="SkiaSharp" Version="2.88.0" />
```
DmyMi, line 62: as the test is failing, probably need the `SkiaSharp.NativeAssets.Linux` package in addition to this

PR #749, opened 2022-07-12
rbiyovskiy/739 workshop control of the available places

### LINKED ISSUE #739: Workshop: Control of the available places
**Story link**
https://github.com/ita-social-projects/OoS-Backend/issues/206

### Task description
Need to add control of the "Доступних місць". It need to do for the all workshops.
For the some workshops option "Доступних місць" should be disabled (As result conversation with @DmyMi we will use maxInt value for these workshops).

As taken, we should count applications with statuses:  Approved and StudyingForYears.

### To do 
- after if amount of the "Доступних місць" equal max possible - change workshop status to close.
- if amount of the "Доступних місць" less than max possible - change workshop status to open.
- For the form we should prepare amount of the available places (max - taken).
- A notification should be send the same as in case when a provider changed the status by personally.

RBiyovskiy:
- Added the new field workshop.AvailableSeats;
- added logic that will change workshop status automatically if it needed;
- changed tests and added one new more.


OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationServiceTests.cs
```diff
@@ -439,6 +442,42 @@ public async Task UpdateApplication_WhenIdIsValid_ShouldReturnApplication()
...
+        // Arrange
+        var id = new Guid("1745d16a-6181-43d7-97d0-a1d6cc34a8bd");
+        var entity = WithApplication(id);
+        var changedEntity = WithApplication(id, ApplicationStatus.Approved);
+        var userId = Guid.NewGuid().ToString();
+        var workshop = WithWorkshop(new Guid("0083633f-4e5b-4c09-a89d-52d8a9b89cdb"));
+
+        var applicationsMock = WithApplicationsList().AsQueryable().BuildMock();
+
+        applicationRepositoryMock.Setup(r => r.Get(
```
valerabad, line 457: Are you sure we need this setup? I guess this setup isn't matched for this test. Please check it using applicationRepositoryMock.VerifyAll(); in assert section

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationServiceTests.cs
```diff
@@ -439,6 +442,42 @@ public async Task UpdateApplication_WhenIdIsValid_ShouldReturnApplication()
...
+                It.IsAny<string>(),
+                It.IsAny<Expression<Func<Application, bool>>>(),
+                It.IsAny<Dictionary<Expression<Func<Application, object>>, SortDirection>>(),
+                It.IsAny<bool>()))
+            .Returns(applicationsMock)
+            .Verifiable();
+
+        applicationRepositoryMock.Setup(a => a.Update(It.IsAny<Application>(), It.IsAny<Action<Application>>())).ReturnsAsync(changedEntity);
+        applicationRepositoryMock.Setup(a => a.GetById(It.IsAny<Guid>())).ReturnsAsync(entity);
+        applicationRepositoryMock.Setup(a => a.Count(x => x.WorkshopId == workshop.Id && (x.Status == ApplicationStatus.Approved || x.Status == ApplicationStatus.StudyingForYears))).ReturnsAsync(1);
```
valerabad, line 469: The same

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationServiceTests.cs
RBiyovskiy: 1. I read that using VerifyAll() is a bad idea.
For example:
https://hamidmosalla.com/2019/02/17/moq-whats-wrong-with-using-verifyall/
2. Using VerifyAll does not indicate that we can abandon setup methods.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationServiceTests.cs
RBiyovskiy: See above...

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationServiceTests.cs
RBiyovskiy: You are right. setup the Count method is not needed (because returned result always equal 0).

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationServiceTests.cs
RBiyovskiy: You are right. setup the Get method is not needed. It does not run in scope of Update method.

PR #750, opened 2022-07-12
Baditsa/some improvements

valerabad:
1) Attribute [EnumDataType]  has been added to enums in DTO to avoid setting incorrect int values to DB
2) Improve mapping

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Models/ApplicationDTO.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Models/ApplicationDTO.cs
@@ -12,7 +12,7 @@ public class ApplicationDto
 {
     public Guid Id { get; set; }
 
-    [JsonConverter(typeof(StringEnumConverter))]
+    [EnumDataType(typeof(ApplicationStatus), ErrorMessage = "Status should be in enum range")]
     public ApplicationStatus Status { get; set; } = ApplicationStatus.Pending;
 
     [MaxLength(500)]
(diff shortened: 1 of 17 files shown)


OutOfSchool/OutOfSchool.WebApi/Models/ApplicationDTO.cs
```diff
@@ -12,7 +12,7 @@ public class ApplicationDto
 {
     public Guid Id { get; set; }
 
-    [JsonConverter(typeof(StringEnumConverter))]
+    [EnumDataType(typeof(ApplicationStatus), ErrorMessage = "Status should be in enum range")]
```
DmyMi, line 15: are these mutually exclusive? Because FE right now expects enums as string and most enums are sent as string

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Models/ChildDTO.cs
```diff
@@ -29,7 +29,7 @@ public class ChildDto
     public DateTime DateOfBirth { get; set; } = default;
 
     [Required(ErrorMessage = "Gender is required")]
-    [Range(0, 1)]
+    [EnumDataType(typeof(Gender), ErrorMessage = "Gender should be in enum range")]
```
i-rakityanskiy, line 32: There is a possibility to use a unified error message for all of the EnumDataType attributes. Please, create a string constant for the error message.
```suggestion
    [EnumDataType(typeof(Gender), ErrorMessage = "{0} should be in enum range")]
```

RBiyovskiy (approved):

i-rakityanskiy (approved):

PR #764, opened 2022-07-21
Provicevko/create tech admin net6

provicevko:
- Added tech admin role.
- Added necessity to change password after first login.


OutOfSchool/OutOfSchool.IdentityServer/Services/UserManagerAdditionalService.cs
```diff
@@ -0,0 +1,71 @@
...
+        this.userManager = userManager ?? throw new ArgumentNullException(nameof(userManager));
+        this.storeContext = storeContext ?? throw new ArgumentNullException(nameof(storeContext));
+    }
+
+    /// <inheritdoc/>
+    public async Task<IdentityResult> ChangePasswordWithRequiredMustChangePasswordAsync(User user, string currentPassword, string newPassword)
+    {
+        _ = user ?? throw new ArgumentNullException(nameof(user));
+        _ = currentPassword ?? throw new ArgumentNullException(nameof(currentPassword));
+        _ = newPassword ?? throw new ArgumentNullException(nameof(newPassword));
```
RBiyovskiy, line 26: You could use new format for check  the  argument is null.

alrom (approved):

PR #766, opened 2022-07-23
Romanchuk/Change type for edrpou property

alrom:


OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
```diff
@@ -467,7 +467,7 @@ await SendNotification(checkProvider, NotificationAction.Update, statusChanged,
         licenseChanged = false;
 
         if (!(checkProvider.FullTitle == providerDto.FullTitle
-              && checkProvider.EdrpouIpn == long.Parse(providerDto.EdrpouIpn)))
+              && checkProvider.EdrpouIpn == providerDto.EdrpouIpn))
```
provicevko, line 470: Maybe it's better to use `checkProvider.EdrpouIpn.Equals(providerDto.EdrpouIpn)`

provicevko (approved):

OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
provicevko: Okay, after some tests I've found that `==` can be enough for this case, so this comment no longer useful.

VadymLevkovskyi (approved):

PR #767, opened 2022-07-24
Added ability to see providers' chats for ministry admins

provicevko:
**_For Ministry Admins:_**
- Added ability to get chat rooms by **Provider Id**  if **Provider** is a subordinate of **Ministry Admin**.
- Added ability to see chat messages in allowed rooms (where **Provider** is a subordinate).
- Added **Operation handler** in order to avoid duplicates.

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChatWorkshopController.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChatWorkshopController.cs
@@ -58,13 +58,13 @@ public class ChatWorkshopController : ControllerBase
         IProviderAdminService providerAdminService,
         IMinistryAdminService ministryAdminService)
     {
-        this.messageService = messageService;
-        this.roomService = roomService;
-        this.validationService = validationService;
-        this.localizer = localizer;
-        this.logger = logger;
-        this.providerAdminService = providerAdminService;
-        this.ministryAdminService = ministryAdminService;
+        this.messageService = messageService ?? throw new ArgumentNullException(nameof(messageService));
+        this.roomService = roomService ?? throw new ArgumentNullException(nameof(roomService));
+        this.validationService = validationService ?? throw new ArgumentNullException(nameof(validationService));
+        this.localizer = localizer ?? throw new ArgumentNullException(nameof(localizer));
+        this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        this.providerAdminService = providerAdminService ?? throw new ArgumentNullException(nameof(providerAdminService));
+        this.ministryAdminService = ministryAdminService ?? throw new ArgumentNullException(nameof(ministryAdminService));
     }
 
     /// <summary>
--- a/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChatWorkshopController.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChatWorkshopController.cs
@@ -5,6 +5,7 @@ using System.Security.Authentication;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.JsonPatch.Operations;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Localization;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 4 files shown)


OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChatWorkshopController.cs
```diff
@@ -46,20 +48,23 @@ public class ChatWorkshopController : ControllerBase
...
         ILogger<ChatWorkshopController> logger,
-        IProviderAdminService providerAdminService)
+        IProviderAdminService providerAdminService,
+        IMinistryAdminService ministryAdminService)
     {
-        this.messageService = messageService ?? throw new ArgumentNullException(nameof(messageService));
-        this.roomService = roomService ?? throw new ArgumentNullException(nameof(roomService));
-        this.validationService = validationService ?? throw new ArgumentNullException(nameof(validationService));
-        this.localizer = localizer ?? throw new ArgumentNullException(nameof(localizer));
+        this.messageService = messageService;
```
VadymLevkovskyi, line 61: Not sure if it is a good tendency to drop all null-checks. 
You are using implicit assumption that might shoot your leg when some part of framework/used 3rd libraries are changed

VadymLevkovskyi (approved):
From logic perspective - approved. From style/safe-coding perspective - see my other comment.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChatWorkshopController.cs
provicevko: Agree with services, repositories, other public classes, but I'm not sure about controllers that are tight coupled with Asp.Net Core. The only one and main argument to do it is like you said if some code inside framework changes. To be honest, I want to check such parameters on null but it does no sense if we don't check on null parameters in controllers' methods (many developers don't check on null in controllers' methods because the binding mechanism creates non-nullable arguments' instances and then ApiController attribute validates modelstate without having explict code check). So if it was normal practice to check on null in action methods then I'd be fully agree with this defensive programming approach regarding controllers. What do you think?

VadymLevkovskyi: There are lot of thoughts. 
TLDR - it's better to not rely on implicit guarantees that might be changed, especially when extra check is cheap.

---

Well, convince me that I'm wrong, otherwise I'm sure that all people (developers as well) are illogical, lazy and habit-driven species )). That's why I'd like to check any tech practice from these points perspective.

Thus idea of partial rule applying (to do something everywhere except some places) - converts the good simple lazy habit (do null check everywhere and don't overthink) to more complex and (from my perspective) worse habit (should I do it or don't? as latter is easier, so don't do). 

From technical perspective this partial rule applying idea makes code maintaining, i.e. moving code from controllers to services/etc harder, as code that was technically "correct" in one place is no longer correct in other.

There are additional points. All abstractions are leaky (see term definition in  [The Leaky Abstraction Law](https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/) - it is really worth to be read), and "explicit is better than implicit". So please spend extra chars in code line for non-nullability check! Don't trust to implicit assertions.

Summing up, ideally, all controllers' actions' parameters should be covered with this check as well as all constructor injections, and property setters. And it's very paranoidal, and verbose, and hard to use. And we should really stop somewhere as there is no silver bullet. Only thing I'm asking is to be more defensive and use more checks where it is possible and takes small price.

Regarding Asp.Net MVC model-binding (this paragraph is bit irrelevant to main comment, but you've mentioned it so I have to react ;) ). Do we really assure that all our models received from front are checked to be correct by model validators? I guess no, at least ids are not verified to be existed in db. Also I presume that not all non-db related checks are performed also (in fresh PR I've seen that only restriction for EDRPOU Id is to contain 8 or 10 digits - I know that there is control sum included in number - quick googling provides this [site](https://1cinfo.com.ua/Article/Detail/Proverka_koda_po_EDRPOU/) and I'm sure there are more related things). So our models are semi-valid, and no one cares. Also, you could register custom model binder that even could return nulls implicitly. And no one will check it. So please don't use model-binding / validation provided by framework as an argument. They are good to have something to start with, but really far from being ideal.
Ah there is also a discussion showing that there is "no silver bullet" and model binders don't like C#8.0 non-nullability https://stackoverflow.com/questions/71811227/asp-net-mvc-model-binding-vs-nullable-reference-types 

N.B.
There is such thing as "defensive programming" with opposite approach like "offensive programming", there is also such technique as using of "barriers" (that split the code that works with non-validated input and should contain lot of checks from the code that works with pre-validated input, that is trusted and not checked). 
I'm not insisting to use any approach solely or solemnly, I do insist to make usage of any explicit ,shared with team, and enforced by your static analyzer. Or just grow simple habits for the whole team.

provicevko: Thanks for your detailed explanation, really apriciate it :)

PR #769, opened 2022-07-25
Romanchuk/Add child for parent entity

alrom:


OutOfSchool/OutOfSchool.WebApi/Services/ChildService.cs
```diff
@@ -246,6 +251,11 @@ public async Task<ChildDto> UpdateChildCheckingItsUserIdProperty(ChildDto childD
             throw new UnauthorizedAccessException($"User: {userId} is trying to update not his own child. Child Id = {childDto.Id}");
         }
 
+        if (child.IsParent || childDto.IsParent)
```
RBiyovskiy, line 254: Why should we check both conditions? Are the child and childDto is not the same object?

RBiyovskiy (commented):
Please check tests. They are broken.

OutOfSchool/OutOfSchool.WebApi/Services/ChildService.cs
alrom: Because we can't set child with IsParent manually and can't change excisting entity with same property.

alrom: Test were fixed )) Thx

RBiyovskiy (approved):

Issue #214, opened 2021-07-09
Service provider creates workshop (with Competitive selection/without)


Nikita-Babenko:
As a user (service provider) I want to create Workshop with Competitive selection so that to select applied candidates individually  

**Description**
Add a field for selection competitive selection yes / no to the workshop registration page

**Role**
user (service provider)

**Precondition**
service provider creates workshop with Competitive selection
service provider creates admin  without Competitive selection

### Acceptance Criteria
SCENARIO #1 service provider creates workshop with Competitive selection
GIVEN Service provider wants to create workshop with Competitive selection
AND opens cabinet
AND opens tab “Гуртки”
![2021-07-09 17_09_50-MoE _ Dev Team – Figma](https://user-images.githubusercontent.com/12609814/125093591-1df53180-e0db-11eb-8b84-55f392ee2533.png)
AND clicks the button “Додати гурток”
AND workshop creation page opens
![йус](https://user-images.githubusercontent.com/12609814/125093643-2d747a80-e0db-11eb-83f4-c35b78b38bdf.png)
AND clicks the option “Так” next to the text  {{Чи є конкурсний відбір}}
AND fills in all the data 
WHEN clicks “Зберегти”
![2021-07-09 17_13_55-MoE _ Dev Team – Figma](https://user-images.githubusercontent.com/12609814/125093723-441ad180-e0db-11eb-97db-fc22197dbde9.png)
THEN workshop with Competitive selection is created
AND it is displayed on the workshop
![фісфісфі](https://user-images.githubusercontent.com/12609814/125093760-4e3cd000-e0db-11eb-8e6a-44d8611d4d50.png)
AND displayed in tab “Про гурток”
![svsvsrfdbvsrvs](https://user-images.githubusercontent.com/12609814/125093790-56950b00-e0db-11eb-8346-df32f332c6db.png)

SCENARIO #2 service provider creates admin  without Competitive selection
GIVEN Service provider wants to create workshop without Competitive selection
AND opens cabinet
AND opens tab “Гуртки”
![2021-07-09 17_09_50-MoE _ Dev Team – Figma](https://user-images.githubusercontent.com/12609814/125093814-5eed4600-e0db-11eb-8a4c-e7db504a768d.png)
AND clicks the button “Додати гурток”
AND workshop creation page opens
![виевиуевв](https://user-images.githubusercontent.com/12609814/125093847-690f4480-e0db-11eb-995e-31393fb8ab23.png)
[...]

PR #780, opened 2022-07-28
#214 provider competitive selection

trdmitry:
Added new field CompetitiveSelection to Workshop, WorkshopDto, WorkshopCardDto


i-rakityanskiy:
There is a condition in the story that if workshop has no competitive selection, then applications cannot have `AcceptedForSelection` status. I think that a validation should be added for this in scope of this story.

OutOfSchool/OutOfSchool.DataAccess/Models/Workshop.cs
```diff
@@ -56,6 +56,9 @@ public class Workshop : IKeyedEntity<Guid>, IImageDependentEntity<Workshop>
     [Range(0, 18, ErrorMessage = "Max age should be a number from 0 to 18")]
     public int MaxAge { get; set; }
 
+    [Required(ErrorMessage = "CompetitiveSelection is required")]
+    public bool CompetitiveSelection { get; set; } = default;
```
i-rakityanskiy, line 60: What is the purpose of setting `default` as the initial value here?
```suggestion
    [Required(ErrorMessage = "CompetitiveSelection is required")]
    public bool CompetitiveSelection { get; set; }
```

OutOfSchool/OutOfSchool.WebApi/Models/Workshop/WorkshopCard.cs
```diff
@@ -38,6 +38,9 @@ public class WorkshopCard
     [Range(0, 18, ErrorMessage = "Max age should be a number from 0 to 18")]
     public int MaxAge { get; set; }
 
+    [Required(ErrorMessage = "CompetitiveSelection is required")]
+    public bool CompetitiveSelection { get; set; } = default;
```
i-rakityanskiy, line 42: Same here.

OutOfSchool/OutOfSchool.WebApi/Models/Workshop/WorkshopDTO.cs
```diff
@@ -60,6 +60,9 @@ public class WorkshopDTO : IValidatableObject
     [Range(0, 18, ErrorMessage = "Max age should be a number from 0 to 18")]
     public int MaxAge { get; set; }
 
+    [Required(ErrorMessage = "CompetitiveSelection is required")]
+    public bool CompetitiveSelection { get; set; } = default;
```
i-rakityanskiy, line 64: Same here.

i-rakityanskiy (changes requested):

OutOfSchool/OutOfSchool.DataAccess/Models/Workshop.cs
trdmitry: fixed.

OutOfSchool/OutOfSchool.WebApi/Models/Workshop/WorkshopCard.cs
trdmitry: fixed.

OutOfSchool/OutOfSchool.WebApi/Models/Workshop/WorkshopDTO.cs
trdmitry: fixed.

trdmitry:
> There is a condition in the story that if workshop has no competitive selection, then applications cannot have `AcceptedForSelection` status. I think that a validation should be added for this in scope of this story.

Added check status of applications with workshop that has flag CompetetiveSelection

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -453,6 +453,12 @@ private void UpdateStatus(ShortApplicationDto applicationDto, ApplicationDto app
             }
         }
 
+        if (application.Workshop.CompetitiveSelection && application.Status == ApplicationStatus.Pending &&
+            applicationDto.Status != ApplicationStatus.AcceptedForSelection)
```
i-rakityanskiy, line 457: What if we want to reject a pending application and our workshop has competitive selection? It looks like a valid action but it will cause an exception.
On the other hand there is still missing validation for workshops without competitive selection. It is still possible for them to set `ApplicationStatus` to `AcceptedForSelection`.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -453,6 +453,12 @@ private void UpdateStatus(ShortApplicationDto applicationDto, ApplicationDto app
             }
         }
 
+        if (application.Workshop.CompetitiveSelection && application.Status == ApplicationStatus.Pending &&
```
i-rakityanskiy, line 456: Please add tests for different scenarios.

i-rakityanskiy (changes requested):

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
trdmitry: > What if we want to reject a pending application and our workshop has competitive selection? It looks like a valid action but it will cause an exception.

I thought about this case and discussed with Mykita Babenko.  Provider can reject or accept the application after AcceptedForSelection status only.

We can discuss this case again with Mykita
Imho provider should be able to reject the application without competitive selection.

trdmitry: > On the other hand there is still missing validation for workshops without competitive selection. It is still possible for them to set `ApplicationStatus` to `AcceptedForSelection`.

Added validation and testcase for workshops without competitive selection.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
trdmitry: Tests added

OutOfSchool/OutOfSchool.WebApi/Common/StatusPermissions/StatusPermissions.cs
```diff
@@ -0,0 +1,53 @@
...
+            && (Convert.ToInt32(p.ToStatus) == Convert.ToInt32(to) || Convert.ToInt32(p.ToStatus) == 0)
+            && p.Allowed == false
+        );
+
+        if (denyResult)
+            return false;
+
+        var allowResult = AllowedStatuses.Any(p =>
+            (p.Role == role || p.Role == "all")
+            && (Convert.ToInt32(p.FromStatus) == Convert.ToInt32(from) || Convert.ToInt32(p.FromStatus) == 0)
```
RBiyovskiy, line 46: You use generic method and than the convert method in body of them. Maybe will be better convert value to int before give it to method and do not use generic method

RBiyovskiy (commented):
Please, fix tests.

OutOfSchool/OutOfSchool.WebApi/Common/StatusPermissions/StatusPermissions.cs
```diff
@@ -0,0 +1,53 @@
...
+            Role = role,
+            FromStatus = fromStatus,
+            ToStatus = toStatus,
+            Allowed = false,
+        });
+    }
+
+    public bool CanChangeStatus(string role, T from, T to)
+    {
+        var denyResult = AllowedStatuses.Any(p =>
```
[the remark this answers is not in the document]
trdmitry: Renamed to statusPermissionsList

trdmitry:
> Please, fix tests.

fixed

OutOfSchool/OutOfSchool.WebApi/Common/StatusPermissions/StatusPermissions.cs
trdmitry: if I change type to int I should convert enum to int when call the method CanChangeStatus 
image

With geneneric this class can be used for any enums

RBiyovskiy (approved):

DmyMi (approved):

DmyMi:
Implemented by another PR

PR #781, opened 2022-07-28
Provicevko/move automapper mappings

provicevko:
- Moved mapping extensions into MappingProfile, so they are in one place.
- `public static User ToDomain(this ShortUserDto shortUserDto, User user)` needs a separate review to be safety moved that's why it wasn't touched in this PR.

_**Better to review changes sequentially by commits because they're separated by domain models.**_


OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/PermissionsForRoleControllerTests.cs
```diff
@@ -27,13 +28,14 @@ public class PermissionsForRoleControllerTests
...
-
-
+    private IMapper mapper;
 
     [SetUp]
     public void Setup()
     {
         service = new Mock<IPermissionsForRoleService>();
+        var config = new MapperConfiguration(cfg => cfg.AddProfile<Util.MappingProfile>());
+        mapper = config.CreateMapper();
```
h4wk13, line 38: ` var config = new MapperConfiguration(cfg => cfg.AddProfile<Util.MappingProfile>());
        mapper = config.CreateMapper();`
looks like a method in TestsHelper

OutOfSchool/OutOfSchool.WebApi/Services/ParentService.cs
```diff
@@ -140,7 +149,7 @@ public async Task<ParentDTO> GetByUserId(string id)
 
         logger.LogInformation($"Successfully got a Parent with UserId = {id}.");
 
-        return parents.FirstOrDefault().ToModel();
+        return mapper.Map<ParentDTO>(parents.FirstOrDefault());
```
h4wk13, line 152: Looks like a possible NRE in the mapper, or check for `null` is redundant in the `ChatRoomWorkshopervice` ?

provicevko: I saw that but I didn't want to change any logic, just replace extensions with IMapper alternative. For now `mapper.Map<ParentDTO>` will return `null`, not NRE but I also agree that it'd be better to check for nulls like in the `ChatRoomWorkshopervice `.

DmyMi (approved):

PR #784, opened 2022-08-01
Baditsa/improve email templates

valerabad:
1) GetHtmlPlainStringAsync() method was added for return couple of html and plain text strings
2) Some css fixes
3) Update link to https://mms.gov.ua/


DmyMi (approved):

DmyMi (commented):
needs to be re-opened and merged with a new PR

Issue #811, opened 2022-08-18
GetChildrenList/GetWorkshopList by Id


litvinets:
Create endpoints:
1) GetChildrenListByParentId (probably it is better do not to pass id, and do the same way as on the screen)
![image](https://user-images.githubusercontent.com/27493959/185417869-ee56b288-adab-48c8-a474-179827ac3af5.png)
list child model: id, fullname
2) GetWorkshopListByProviderId - return the whole list of provider workshops 
list workshop model: id: fullTitle
![image](https://user-images.githubusercontent.com/27493959/185418409-8a705de3-1f60-46ab-8ab8-bf5e29cd5d05.png)
![image](https://user-images.githubusercontent.com/27493959/185418369-9638733a-932d-4310-b797-edba7b5a51f9.png)

PR #813, opened 2022-08-18
Add endpoints

alrom:
#811
GetWorkshopListByProviderId
GetChildrenListByParentId

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Services/IChildService.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Services/IChildService.cs
@@ -38,6 +38,15 @@ public interface IChildService
     Task<SearchResult<ChildDto>> GetByFilter(ChildSearchFilter filter);
 
     /// <summary>
+    /// Get all children (Id, FullName) from the database.
+    /// </summary>
+    /// <param name="parentId">ParentId.</param>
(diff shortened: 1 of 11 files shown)


OutOfSchool/OutOfSchool.WebApi/Services/IChildService.cs
```diff
@@ -37,6 +37,15 @@ public interface IChildService
     /// <exception cref="SqlException">If the database cannot execute the query.</exception>
     Task<SearchResult<ChildDto>> GetByFilter(ChildSearchFilter filter);
 
+    /// <summary>
+    /// Get all children (Id, FullName) from the database.
+    /// </summary>
+    /// <param name="parentId">ParentId.</param>
+    /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.
+    /// The result is a <see cref="List{ShortEntityDto}"/> that contains the count of all found children and a list of children that were received.</returns>
```
provicevko, line 45: It doesn't return count of children.

provicevko (approved):

OutOfSchool/OutOfSchool.WebApi/Services/IChildService.cs
alrom: ))))) Done

Issue #812, opened 2022-08-18
Separate Parent and User interface


litvinets:
Requirements:
only parents should have the required date of birth and gender, the rest of the users with all the others roles should not (among them ministryAdmin, providerDeputy, providerAdmin/ provider)

**note:**
remove from the Update MinistryAdmin gender

PR #817, opened 2022-08-22
Provicevko/birh day gender changes

provicevko:
#812 
- Removed Gender from User
- Added Gender, DateOfBirth for Parent
- Added endpoints for personal info according to new requirements
- Added essential methods in parent service for new endpoints
- Removed MappingExtensions
- Fixed tests
- Minimal changes way


SergeyNovitsky (approved):

OutOfSchool/OutOfSchool.IdentityServer/ViewModels/RegisterViewModel.cs
```diff
@@ -31,6 +31,10 @@ public class RegisterViewModel
     [EnumDataType(typeof(Gender), ErrorMessage = Constants.EnumErrorMessage)]
     public Gender Gender { get; set; }
 
+    [Required]
```
DmyMi, line 34: need to think something about this, because the same viewmodel is used  for both parent & provider. And for provider this is not required as well as gender

DmyMi (approved):

PR #818, opened 2022-08-22
Required fields

ZdorovenkoKyrylo:


OutOfSchool/OutOfSchool.WebApi/Models/Providers/ProviderDto.cs
```diff
@@ -54,9 +54,11 @@ public class ProviderDto
     public string EdrpouIpn { get; set; }
 
     [MaxLength(50)]
+    [Required(ErrorMessage = "The name of the director is required")]
     public string Director { get; set; } = string.Empty;
 
     [DataType(DataType.Date)]
+    [Required(ErrorMessage = "The date of birth is required")]
```
provicevko, line 61: I'm not sure if Required attribute works fine with non-nullable types like DateTime structure.

provicevko (approved):

DmyMi (approved):

VadymLevkovskyi (approved):

PR #828, opened 2022-08-30
fix issue with list of teachers is null

ZdorovenkoKyrylo:


OutOfSchool/OutOfSchool.WebApi/Models/Workshop/WorkshopDTO.cs
```diff
@@ -116,7 +116,7 @@ public class WorkshopDTO : IValidatableObject
     [ModelBinder(BinderType = typeof(JsonModelBinder))]
     public AddressDto Address { get; set; }
 
-    public List<TeacherDTO> Teachers { get; set; }
+    public List<TeacherDTO> Teachers { get; set; } = new ();
```
Vizary, line 119: I think you should check this property for null value in place where it uses and throw an exeption if it's null

provicevko: If I remember right, actually we allow to create workshops without teachers, so an exception if null may be inappropriate behaviour

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
```diff
@@ -86,7 +92,8 @@ public async Task<WorkshopCreationResultDto> CreateV2(WorkshopDTO dto)
         _ = dto ?? throw new ArgumentNullException(nameof(dto));
         logger.LogInformation("Workshop creating was started.");
 
-        async Task<(Workshop createdWorkshop, MultipleImageUploadingResult imagesUploadResult, Result<string> coverImageUploadResult)> CreateWorkshopAndDependencies()
+        async Task<(Workshop createdWorkshop, MultipleImageUploadingResult imagesUploadResult, Result<string>
```
Vizary, line 95: Try to avoid extra changes where you don't change logic if it wasn't approved before. It will confuse reviewers less while review process

Vizary (approved):

VadymLevkovskyi (approved):

PR #839, opened 2022-09-02
Main information

ZdorovenkoKyrylo:


OutOfSchool/OutOfSchool.WebApi/Services/CompanyInformationService.cs
```diff
@@ -56,13 +56,13 @@ public async Task<CompanyInformationDto> GetByType(CompanyInformationType type)
...
         logger.LogDebug("Updating CompanyInformation is started.");
-
         if (companyInformationDto == null)
         {
             throw new ArgumentNullException(nameof(companyInformationDto));
         }
 
-        if (companyInformationDto.CompanyInformationItems.Count() > LimitOfItems)
+        if (companyInformationDto.CompanyInformationItems.Count() > LimitOfItems
+            || (type == CompanyInformationType.Main && companyInformationDto.CompanyInformationItems.Count() > 1))
```
Vizary, line 65: You can carry out conditionals into separate variables to avoid dublication your code and do code more readable.
Don't use magic number like "1" here, as well, create some constant value for it.

Vizary (approved):

alrom (approved):

PR #840, opened 2022-09-04
Romanchuk/Achievements improvement

alrom:
Constraints to achievement's controller were added

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AchievementController.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AchievementController.cs
@@ -1,17 +1,10 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Net.Mime;
-using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AchievementController.cs
```diff
@@ -161,15 +182,61 @@ public async Task<ActionResult> Update(AchievementCreateDTO achievementDto)
...
             return NoContent();
         }
-
         catch (ArgumentException ex)
         {
             return BadRequest(ex.Message);
         }
     }
+
+    private async Task<bool> IsUserProvidersOwnerOrAdmin(Guid workshopId)
```
DmyMi, line 214: we have same method in some other controllers. Maybe create a base class so we don't ctrl+c/ctrl+v? :)

alrom: that method is different

DmyMi (approved):

PR #843, opened 2022-09-07
Provicevk/unified responses

ghost:
**Incoming:**

- Added _ApiError_ model
- Changed _ProviderController_ responses for V1

Note that _ApiError_ model should be used for 4xx, 5xx Http Codes.

All Api errors are introduced with a group and code.  Therefore we use a dictionary of dictionaries in Frontend.

**For future discussions:**

- How does Frontend handle bad responses? Do we need a lot of codes for simillar things like incorrect model for different DTOs?
- Overriding default model validation (_[ApiController]_) behaviour using custom filter.
- Do we need to provide auxiliary object in case if we need some data else (example: ModelState is not valid, hence we return it with a list of fields where an error is occured)?
- Do we need to describe errors as Problem Details according to [RFC](https://www.rfc-editor.org/rfc/rfc7807)?
- We can create overloadings for methods like: `BadRequest(ApiError error)`, `NotFound(ApiError error)` that takes `ApiError` as parameter.

**For thinking:**

- We can try to create some type like Err in builder behavior:
`return Err.WithMessage(ProviderApiError.ProviderNotCreated).WithCode(HttpStatusCode.BadRequest);`

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Common/Responses/ApiError.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Common/Responses/ApiError.cs
@@ -1,0 +1,23 @@
+﻿using Ardalis.SmartEnum;
+
+namespace OutOfSchool.WebApi.Common.Responses;
+
+public abstract class ApiError
+{
+    protected ApiError(string code, string message)
+    {
+        if (string.IsNullOrEmpty(code))
+        {
+            throw new ArgumentException(@"Code must be non-empty value", nameof(code));
+        }
+
+        Code = code;
+        Message = message;
+    }
+
+    public string Code { get; }
+
+    public string Message { get; }
+
+    public abstract string GroupName { get; }
+}
--- a/OutOfSchool/OutOfSchool.WebApi/Common/Responses/ApiErrorResponse.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Common/Responses/ApiErrorResponse.cs
@@ -1,0 +1,22 @@
+﻿using Ardalis.SmartEnum;
+
+namespace OutOfSchool.WebApi.Common.Responses;
+
+public record ApiErrorResponse
+{
+    private ApiErrorResponse(string groupName, string code, string message)
+    {
+        GroupName = groupName ?? throw new ArgumentNullException(nameof(groupName));
+        Code = code ?? throw new ArgumentNullException(nameof(code));
+        Message = message;
+    }
+
+    public static ApiErrorResponse Create(string groupName, string code, string message)
+        => new (groupName, code, message);
+
+    public string GroupName { get; init; }
+
+    public string Code { get; init; }
+
+    public string Message { get; init; }
+}
--- a/OutOfSchool/OutOfSchool.WebApi/Common/Responses/InputApiError.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Common/Responses/InputApiError.cs
@@ -1,0 +1,14 @@
+﻿namespace OutOfSchool.WebApi.Common.Responses;
+
+public class InputApiError : ApiError
+{
+    public static readonly InputApiError InputDataIncorrect
+        = new (nameof(InputDataIncorrect), "Provided data input is incorrect");
+
+    private InputApiError(string code, string message)
+        : base(code, message)
+    {
+    }
+
+    public override string GroupName => nameof(InputApiError);
+}
--- a/OutOfSchool/OutOfSchool.WebApi/Common/Responses/ProviderApiError.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Common/Responses/ProviderApiError.cs
@@ -1,0 +1,23 @@
+﻿namespace OutOfSchool.WebApi.Common.Responses;
+
+public sealed class ProviderApiError : ApiError
+{
+    public static readonly ProviderApiError ProviderNotCreated
+        = new (nameof(ProviderNotCreated), "Provider was not created");
+
+    public static readonly ProviderApiError ProviderNotUpdated
+        = new (nameof(ProviderNotUpdated), "Provider was not updated");
+
+    public static readonly ProviderApiError ProviderNotDeleted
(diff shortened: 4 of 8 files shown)


DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Common/Responses/ApiError.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Ardalis.SmartEnum;
```
VadymLevkovskyi, line 1: (Excessive code) Is this using necessary?

OutOfSchool/OutOfSchool.WebApi/Common/Responses/ApiErrorResponse.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Ardalis.SmartEnum;
```
VadymLevkovskyi, line 1: (Excessive code) Is this using necessary?

OutOfSchool/OutOfSchool.WebApi/Common/Responses/ApiErrorResponse.cs
```diff
@@ -0,0 +1,22 @@
...
+public record ApiErrorResponse
+{
+    private ApiErrorResponse(string groupName, string code, string message)
+    {
+        GroupName = groupName ?? throw new ArgumentNullException(nameof(groupName));
+        Code = code ?? throw new ArgumentNullException(nameof(code));
+        Message = message;
+    }
+
+    public static ApiErrorResponse Create(string groupName, string code, string message)
```
VadymLevkovskyi, line 14: Please explain why instead of public ctor you have private ctor and public static? Statics are worser from testability point of view. As it does the same as ctor - what is the main reason to have it?

OutOfSchool/OutOfSchool.WebApi/Common/Responses/InputApiError.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace OutOfSchool.WebApi.Common.Responses;
+
+public class InputApiError : ApiError
+{
+    public static readonly InputApiError InputDataIncorrect
```
VadymLevkovskyi, line 5: Same question about static vs ctor.

OutOfSchool/OutOfSchool.WebApi/Common/Responses/InputApiError.cs
```diff
@@ -0,0 +1,14 @@
...
+        = new (nameof(InputDataIncorrect), "Provided data input is incorrect");
```
VadymLevkovskyi, line 6: (Non-uniqueness) `nameof` is a cool idea, but it will give you only method/prop name. Thus, in other class (i.e. child of this class named `InputAdminApiError`) method/prop with same name (`InputDataIncorrect`) could exist and you'll get eror-code collision if you forget to redefine `GroupName`. Please use simple enum containing all error codes, or more complex schema ot generate them - at least 
`$"{nameof(class)}.{nameof(method or property)}"`
Or even more complex approach could be used with [CallerMemberNameAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callermembernameattribute?view=net-7.0) and it's friends, if you like it

OutOfSchool/OutOfSchool.WebApi/Common/Responses/InputApiError.cs
VadymLevkovskyi: I was wrong, it's not a method, it's prop. Please ignore my comment

OutOfSchool/OutOfSchool.WebApi/Common/Responses/ProviderApiError.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace OutOfSchool.WebApi.Common.Responses;
+
+public sealed class ProviderApiError : ApiError
+{
+    public static readonly ProviderApiError ProviderNotCreated
```
VadymLevkovskyi, line 5: See previous comment re `nameof` and error codes

OutOfSchool/OutOfSchool.WebApi/Common/Responses/InputApiError.cs
```diff
@@ -0,0 +1,14 @@
...
+
+    private InputApiError(string code, string message)
+        : base(code, message)
+    {
+    }
+
+    public override string GroupName => nameof(InputApiError);
```
VadymLevkovskyi, line 13: (Performance) Do you need this override? Why not set `GroupName` in ctor?

VadymLevkovskyi: (Related to all overiden `GroupName`s)

OutOfSchool/OutOfSchool.WebApi/Extensions/ApiErrorExtensions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Ardalis.SmartEnum;
+using OutOfSchool.WebApi.Common.Responses;
+
+namespace OutOfSchool.WebApi.Extensions;
+
+public static class ApiErrorExtensions
+{
+    public static ApiErrorResponse CreateApiErrorResponse(this ApiError apiError)
```
VadymLevkovskyi, line 8: (Code style) Consider necessity of this extension class.
Ususally extensions are good for code decoupling. Mapping extensions - when you have 2 classses not related to each other and you don't want introduce any hard dependency between them. 

It is not the case for `ApiError` and `ApiErrorResponse`, as latter is direct child of former. So why not put this method in `ApiErrorResponse` class?

VadymLevkovskyi (commented):
Probably I don't get what problem you are fixing. Please explain why we need lot of inherited error types with no changes in data. 
Are there any plans for extending logic or properties in future children? 

Maybe it could be simpler? Like 1 class with lot of static properties for each api error code case, or simple enum, list of tuples (if you need that GroupName as separate thing), whatever?

PR #869, opened 2022-09-27
rights checking in logic proposal

DmyMi:
TODO:
 - Tests
 - Refactor usage of `IProviderService.GetByUserId` to new signature (`Task<ProviderDto?> GetByUserId(string id, bool isDeputyOrAdmin = false);`)


OutOfSchool/OutOfSchool.WebApi/Services/CurrentUserService.cs
```diff
@@ -0,0 +1,134 @@
...
+
+    // TODO: in code we check multiple permissions (for example parent and provider can edit application status)
+    // TODO: either separate method calls with IsInRole check, or array like here?
+    public async Task UserHasRights(params IUserType[] userTypes)
+    {
+        // TODO: double check role?
+        var userHasRights = false;
+        if (user?.Identity?.IsAuthenticated ?? false)
+        {
+            var parent = userTypes.Select(t => t as ParentType).FirstOrDefault(p => p is not null);
```
VadymLevkovskyi, line 51: Just small code style note: you could use `Enumerable.OfType<TResult>()` for code pieces `Select(_ => _ as TResult).Where(_ => _ is not null)` - it provides a bit neater code

OutOfSchool/OutOfSchool.WebApi/Services/CurrentUserService.cs
```diff
@@ -0,0 +1,134 @@
...
+
+            var result = await Task.WhenAll(
+                new List<Task<bool>>
+                    {
+                        UserHasRights(parent),
+                        UserHasRights(provider),
+                        UserHasRights(providerAdmin),
+                    }
+                    .Select(Execute));
+            userHasRights = result.Aggregate(false, (acc, next) => acc || next);
```
VadymLevkovskyi, line 63: Another code style comment, if I get it correctly all you do here is just checking for any `true` value, right? Why not then use `result.Any(hasRight => hasRight)` ?

OutOfSchool/OutOfSchool.WebApi/Services/CurrentUserService.cs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        if (!userHasRights)
+        {
+            // TODO: copied from existing. Need custom exception
+            throw new ArgumentException("User has no rights to perform operation");
```
VadymLevkovskyi, line 69: You could use built-in [UnauthorizedAccessException](https://learn.microsoft.com/en-us/dotnet/api/system.unauthorizedaccessexception?redirectedfrom=MSDN&view=netframework-4.7.2)
or [SecurityException](https://learn.microsoft.com/en-us/dotnet/api/system.security.securityexception?view=net-7.0)

OutOfSchool/OutOfSchool.WebApi/Services/CurrentUserService.cs
```diff
@@ -0,0 +1,134 @@
...
+
+    private static async Task<bool> Execute(Task<bool> original)
+    {
+        try
+        {
+            return await original;
+        }
+        catch (Exception)
+        {
+            // TODO: Logging?
```
VadymLevkovskyi, line 81: I would like to add logging not only here, but for all cases when security check is failed (by "failed" I mean all places when check for rights return `false`). Will call it security audit logs, let it switch on/off separately, and make it switch on by default. 
I know, that it looks like a bit overkill for this project, but still it's a good practice for checking security incidents.

OutOfSchool/OutOfSchool.WebApi/Services/CurrentUserService.cs
```diff
@@ -0,0 +1,134 @@
...
+    private Task<bool> UserHasRights<T>(T? userType)
+        where T : IUserType
+        => userType switch
+        {
+            ParentType parent => ParentHasRights(parent.parentId),
+            ProviderType provider => ProviderHasRights(provider.providerAdminId),
+            ProviderAdminType providerAdmin => ProviderAdminHasRights(
+                providerAdmin.providerId,
+                providerAdmin.workshopId),
+            _ => Task.FromResult(false)
```
VadymLevkovskyi, line 95: Shouldn't default case throw a `NotImplementedException`? To let it easier distinguish "no rights" case, and "missed implementation" case

OutOfSchool/OutOfSchool.WebApi/Services/CurrentUserService.cs
```diff
@@ -0,0 +1,134 @@
...
+
+    // TODO: initial logic is probably messed up
+    private async Task<bool> ProviderAdminHasRights(Guid providerId, Guid workshopId)
+    {
+        try
+        {
+            var provider = await providerService.GetByUserId(UserId).ConfigureAwait(false);
+            return provider.Id == providerId;
+        }
+        catch (ArgumentException)
```
VadymLevkovskyi, line 121:
```suggestion
        catch (ArgumentException)
        {
            var isUserRelatedAdmin = await providerAdminService
                .CheckUserIsRelatedProviderAdmin(UserId, providerId, workshopId).ConfigureAwait(false);
            return isUserRelatedAdmin;
        }
```

OutOfSchool/OutOfSchool.WebApi/Services/CurrentUserService.cs
```diff
@@ -0,0 +1,134 @@
...
+            var isUserRelatedAdmin = await providerAdminService
+                .CheckUserIsRelatedProviderAdmin(UserId, providerId, workshopId).ConfigureAwait(false);
+            if (!isUserRelatedAdmin)
+            {
+                return false;
+            }
+        }
+
+        // TODO: probably messed up logic, was just ctrl+c/ctrl+v
+        return true;
```
VadymLevkovskyi, line 132:
```suggestion
        catch (ArgumentException)
        {
            var isUserRelatedAdmin = await providerAdminService
                .CheckUserIsRelatedProviderAdmin(UserId, providerId, workshopId).ConfigureAwait(false);
            return isUserRelatedAdmin;
        }
```

OutOfSchool/OutOfSchool.WebApi/Services/CurrentUserService.cs
```diff
@@ -0,0 +1,134 @@
...
+            ProviderType provider => ProviderHasRights(provider.providerAdminId),
+            ProviderAdminType providerAdmin => ProviderAdminHasRights(
+                providerAdmin.providerId,
+                providerAdmin.workshopId),
+            _ => Task.FromResult(false)
+        };
+
+    private async Task<bool> ParentHasRights(Guid parentId)
+    {
+        var parent = await parentService.GetByUserId(UserId).ConfigureAwait(false);
```
VadymLevkovskyi, line 100: Perf consideration: Here we take and map the whole object just to check if id is same. Isn't it better to move this check on repo/SQL level? 

I know that it slightly breaks used abstractions, but as check is pretty simple and we could gain perf boost, probably, why not do it?

DmyMi: I was thinking about having some of user information cached in Redis, because:
1. These checks gonna happen with every user request to protected operations. 
2. Most Id's (e.g., userid, parent or provider id) will never change, so safe to cache.

Was thinking about `~5 min` cache time for this info as according to some statistic i found in google - average session duration on a website in around 3 minutes.

andriy-korchynskyy (approved):

andriy-korchynskyy (approved):

PR #872, opened 2022-09-29
Update has been modified via request to IdentityServer

valerabad:


OutOfSchool/OutOfSchool.Common/Models/UpdateMinistryAdminDto.cs
```diff
@@ -0,0 +1,6 @@
+namespace OutOfSchool.Common.Models;
+
+public class UpdateMinistryAdminDto : CreateMinistryAdminDto
+{
```
andriy-korchynskyy, line 4: Why this class is empty?

andriy-korchynskyy (approved):

PR #923, opened 2022-12-08
Tsylia/922 achievement types of localization

andriitsylia:


OutOfSchool/OutOfSchool.WebApi/Enums/AchievementTypeLocalization.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace OutOfSchool.WebApi.Enums;
+
+public enum AchievementTypeLocalization
```
alrom, line 3: Maybe, it will be better to name it like "LocalizationType" and use everywhere if we need.

Issue #917, opened 2022-12-03
[GetProviderAdminById] - add new API request to get provider admin by its id


litvinets:
Because of teh added pagination, wee need to have an opportunity to get provider admin by its id, even after reloading the page (after reload we loose our data from state, so wee need an API to get data by provider admin id)

![image](https://user-images.githubusercontent.com/27493959/205463653-ffb57eda-3d7f-4dd5-8d93-f183dab73712.png)

PR #924, opened 2022-12-08
Get provider admin by Id

oleksandr-vovk:
#917

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ProviderAdminController.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ProviderAdminController.cs
@@ -1,7 +1,6 @@
 ﻿using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Filters;
-using Nest;
 using OutOfSchool.Common.Models;
 using OutOfSchool.Services.Enums;
 using OutOfSchool.Services.Models;
--- a/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ProviderAdminController.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ProviderAdminController.cs
@@ -267,6 +267,14 @@ public class ProviderAdminController : Controller
     [HttpGet("{providerAdminId}")]
     public async Task<IActionResult> GetProviderAdminById(string providerAdminId)
     {
-        throw new NotImplementedException();
+        var providerAdmin = await providerAdminService.GetProviderAdminById(userId, providerAdminId)
+            .ConfigureAwait(false);
+
+        if (providerAdmin == null)
+        {
+            return NotFound();
+        }
+
+        return Ok(providerAdmin);
     }
 }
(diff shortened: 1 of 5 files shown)


alrom (approved):

andriy-korchynskyy (approved):

DmyMi:
`[Merge branch 'develop' into get-provider-admin-by-id](https://github.com/ita-social-projects/OoS-Backend/pull/924/commits/efb526dadc333d6076bb84523f1654086ce8d59b)` need to remove this commit and do a rebase. It may lead to unnecessary conflicts for other branches.

OutOfSchool/OutOfSchool.WebApi/Services/CurrentUserService.cs
```diff
@@ -186,6 +189,28 @@ private async Task<bool> ProviderAdminHasRights(string providerAdminId)
         return result;
     }
 
+    private async Task<bool> ProviderDeputyHasRights(Guid providerId)
```
DmyMi, line 192: probably can be merged with provideradmin has rights like parent/child check is working. I will see if there's a good solution to it.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ProviderAdminController.cs
```diff
@@ -253,4 +256,41 @@ public async Task<IActionResult> ManagedWorkshops()
...
+        var providerAdmin = await providerAdminService.GetById(providerAdminId).ConfigureAwait(false);
+
+        if (providerAdmin == null)
+        {
+            return NoContent();
+        }
+
+        if (providerAdmin.IsDeputy)
+        {
+            await currentUserService.UserHasRights(new ProviderRights(providerAdmin.ProviderId))
```
DmyMi, line 282: this check should be done in service layer methods

OutOfSchool/OutOfSchool.WebApi/Services/ProviderAdminService.cs
```diff
@@ -498,6 +498,35 @@ public async Task<IEnumerable<string>> GetProviderDeputiesIds(Guid providerId)
         return providersDeputies.Select(d => d.UserId);
     }
 
+    /// <inheritdoc/>
+    public async Task<ProviderAdminDto> GetFullProviderAdmin(ProviderAdminProviderRelationDto providerAdmin)
```
DmyMi, line 502: why do you need to call get by id in controller then pass this object down here and call this method?:) can't it all be do inside here? Also, `GetByFilter(Expression<Func<TEntity, bool>> predicate, string includeProperties = "")` has `includeProperties` so you can use this method to both get by id and include other entities via join

DmyMi (changes requested):

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ProviderAdminController.cs
oleksandr-vovk: We can't use CurrentUserService inside ProviderAdminService because we have an instance of ProviderAdminService in CurrentUserService, so I used currentUserService here.

OutOfSchool/OutOfSchool.WebApi/Services/ProviderAdminService.cs
oleksandr-vovk: I described a reason in a comment above.

DmyMi (approved):

PR #925, opened 2022-12-08
Get SearchChat by UserId

### LINKED ISSUE #897: [Chat] Add an endpoint(s) to search a chat by user ID
### Task description
Now we can get a chat room only by chatroom id.

image

However, we do not always have a chatroom ID

### To do 
* Add endpoint(s) to get a chatroom by a user ID (be it a workshop or a parent)

alrom:
GetParentsRoomByWorkshop and GetProvidersRoomsByParent were added to ChatWorkshopController.
Closed #897

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChatWorkshopController.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChatWorkshopController.cs
@@ -1,21 +1,10 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Security.Authentication;
-using System.Threading.Tasks;
-using Microsoft.AspNetCore.Authorization;
-using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.JsonPatch.Operations;
+﻿using System.Security.Authentication;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Localization;
-using Microsoft.Extensions.Logging;
-using OutOfSchool.Common;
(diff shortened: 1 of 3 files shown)


andriy-korchynskyy (approved):

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChatWorkshopController.cs
```diff
@@ -175,6 +164,36 @@ public Task<IActionResult> GetParentsRoomsAsync()
     public Task<IActionResult> GetProvidersRoomsAsync()
         => this.GetUsersRoomsAsync(providerId => roomService.GetByProviderIdAsync(providerId));
 
+    /// <summary>
+    /// Get a chat room for current parent and workshopId.
+    /// </summary>
+    /// <param name="workshopId">WorkShop's Id.</param>
+    /// <returns>ChatRoom that was found.</returns>
+    [HttpGet("parent/chatroomforworkshop/{workshopId}")]
+    [Authorize(Roles = "parent")]
```
DmyMi, line 173: we're using permissions, so maybe use one of the permissions instead of role, like ParentRead/ProviderRead if applicable?

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChatWorkshopController.cs
alrom: The user, who call this endpiont, must be "parent" in our logic.

PR #949, opened 2022-12-23
update provider subrole

DmyMi:


andriy-korchynskyy (approved):

OutOfSchool/OutOfSchool.IdentityServer/Services/ProfileService.cs
```diff
@@ -75,25 +75,25 @@ private async Task<string> GetPermissionsForUser(User userFromLogin, string role
     }
 
     // Get subrole for user
-    private async Task<Subrole> GetSubroleByUserName(User userFromLogin)
+    private async Task<ProviderSubRole> GetProviderSubRoleByUserName(User userFromLogin)
     {
         if (userFromLogin.Role == nameof(Role.Provider).ToLower() && userFromLogin.IsDerived)
```
h4wk13, line 80: please update to use `StringComparison.IgnoreCase` instead of using `ToLower()`

OutOfSchool/OutOfSchool.IdentityServer/Services/ProfileService.cs
```diff
@@ -75,25 +75,25 @@ private async Task<string> GetPermissionsForUser(User userFromLogin, string role
...
         {
-            var userDeputyOrAdmin = await providerAdminRepository
-                .GetByFilter(p => p.UserId == userFromLogin.Id)
-                .ConfigureAwait(false);
+            var userDeputyOrAdmin = (await providerAdminRepository
+                .GetByFilter(p => p.UserId == userFromLogin.Id))
+                .ToList();
 
             if (userDeputyOrAdmin.Any(u => u.IsDeputy))
```
h4wk13, line 86: I believe we should use `.Count(u => u.IsDeputy)` instead of `Any(...)` wo we can test the number instead of 2 collection enumeration.

PR #950, opened 2022-12-23
filtration and pagination for chatrooms done

potapchuk-anna:


andriy-korchynskyy (approved):

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChatWorkshopController.cs
```diff
@@ -409,6 +417,10 @@ async Task<IActionResult> Operation()
         return await HandleOperationAsync(Operation);
     }
 
+    private bool IsFilterEmpty(ChatWorkshopFilter filter)
```
DmyMi, line 420: if filter is null won't this method throw a nullreference here `filter.WorkshopIds.Count()`?

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChatWorkshopController.cs
potapchuk-anna: Thank you. I've fixed this bug now

PR #966, opened 2023-01-11
Update application statuses by schedule

oleksandr-vovk:


alrom (approved):

OutOfSchool/OutOfSchool.DataAccess/Repository/ApplicationRepository.cs
```diff
@@ -67,4 +68,19 @@ public Task<int> GetCountByWorkshop(Guid workshopId)
 
         return applications.CountAsync();
     }
+
+    /// <summary>
+    /// Changes status to StudyingForYears of all applications where status is Approved.
+    /// </summary>
+    /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
+    public async Task<int> UpdateAllApprovedApplications()
```
alrom, line 76: Why we have returning type is Task<int>? Is we really need this int?

DmyMi (approved):

PR #974, opened 2023-01-14
Nazar yavor26/add english localization to social group

NazarYavor26:


OutOfSchool/OutOfSchool.WebApi/Services/SocialGroupService.cs
```diff
@@ -44,23 +45,29 @@ public class SocialGroupService : ISocialGroupService
...
+        {
+            Id = x.Id,
+            Name = localization == LocalizationType.En ? x.NameEn : x.Name,
+        });
         logger.LogInformation(!socialGroups.Any()
             ? "SocialGroup table is empty."
             : $"All {socialGroups.Count()} records were successfully received from the SocialGroup table");
 
-        return socialGroups.Select(socialGroup => mapper.Map<SocialGroupDto>(socialGroup)).ToList();
+        return mapper.Map<List<SocialGroupDto>>(socialGroupsLocalized);
```
alrom, line 63: You can call method ToList when you doing selection. And after it use previeus result in logging and returning.

DmyMi (approved):

PR #984, opened 2023-01-31
Mapped MiddleName to string.Empty in Dtos if it's null.

andriitsylia:
Mapped MiddleName to string.Empty in Dtos if it's null.


h4wk13:
Just wondering:
What is the purpose of such changes ?
Do we have some convention to return empty fields in such cases ?

andriitsylia:
It's a request of the frontend team to escape the check of the MiddleName field on null on their side because this field isn't required.

h4wk13:
> It's a request of the frontend team to escape the check of the MiddleName field on null on their side because this field isn't required.

Does it mean we have to update all other such fields ?
If no - why ?

DmyMi (approved):

h4wk13 (approved):

PR #997, opened 2023-02-08
hide private workshops

DmyMi:


VadymLevkovskyi (commented):
Could you name the unit test having assertion that private providers are filtered out? Lot of your code changes are related to this filtering and I cannot find any related test

PR #1003, opened 2023-02-09
Fix numbers of get workshops

ZdorovenkoKyrylo:


OutOfSchool/OutOfSchool.IdentityServer/Program.cs
```diff
@@ -47,7 +47,7 @@
     configService.GetSection(identityOptions.Name).Bind(identityOptions);
 
     context.Database.Migrate();
-    identityContext.Database.Migrate();
+    //identityContext.Database.Migrate();
```
alrom, line 50: ????
We really need it?

ZdorovenkoKyrylo: It was necessary to build the project (to avoid forbidden status and error of 403) it seems to me, I think I forget to uncomment it and now it should be uncommented.

alrom (dismissed):

h4wk13 (approved):

PR #1009, opened 2023-02-15
Tsylia/ Fixed the retrieving children for current parent (with or without him).

andriitsylia:


OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChildController.cs
```diff
@@ -103,11 +104,11 @@ public async Task<IActionResult> GetByParentIdForAdmin(Guid parentId, [FromQuery
     [ProducesResponseType(StatusCodes.Status403Forbidden)]
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     [HttpGet]
-    public async Task<IActionResult> GetUsersChildren([FromQuery] OffsetFilter offsetFilter)
+    public async Task<IActionResult> GetUsersChildren([FromQuery] bool? isGetParent, [FromQuery] OffsetFilter offsetFilter)
     {
```
alrom, line 108: We can use this optional parameter in controller like: 
public async Task<IActionResult> GetUsersChildren([FromQuery] OffsetFilter offsetFilter, [FromQuery] bool isGetParent = false)

OutOfSchool/OutOfSchool.WebApi/Services/ChildService.cs
```diff
@@ -219,22 +219,29 @@ public async Task<SearchResult<ChildDto>> GetByParentIdOrderedByFirstName(Guid p
...
 
         logger.LogDebug($"Getting Child's for User started. Looking UserId = {userId}.");
 
-        var totalAmount = await childRepository.Count(x => x.Parent.UserId == userId).ConfigureAwait(false);
+        Expression<Func<Child, bool>> predicate = x => x.Parent.UserId == userId;
+
+        if (!isGetParent)
+        {
+            predicate = predicate.And(x => x.IsParent == isGetParent);
+        }
```
alrom, line 234: We can write it shorter ))))
predicate = predicate.And(x => !x.IsParent);

andriitsylia: I agree. I'm doing it.

alrom (approved):

DmyMi (approved):

PR #1024, opened 2023-03-16
Add OperationWithObject

alrom:


OutOfSchool/OutOfSchool.DataAccess/Enums/OperationWithObjectEntityType.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace OutOfSchool.Services.Enums;
+
+[JsonConverter(typeof(StringEnumConverter))]
```
DmyMi, line 6: as this is part of internal job management and not exposed to outside world do we need to have string conversion here?:)

OutOfSchool/OutOfSchool.DataAccess/Enums/OperationWithObjectOperationType.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace OutOfSchool.Services.Enums;
+
+[JsonConverter(typeof(StringEnumConverter))]
```
DmyMi, line 6: same as above

OutOfSchool/OutOfSchool.WebApi/Models/OperationWithObjectDto.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.ComponentModel.DataAnnotations;
+using OutOfSchool.Services.Enums;
+
+namespace OutOfSchool.WebApi.Models;
+
+public class OperationWithObjectDto
```
DmyMi, line 6: does it really need to have all fields mapped from data access at the moment? i've only seen the usage of 2-3 so far

OutOfSchool/OutOfSchool.WebApi/Services/IOperationWithObjectService.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// <param name="filter">Filter.</param>
+    /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
+    Task<IEnumerable<OperationWithObjectDto>> GetAll(OperationWithObjectFilter filter);
+
+    /// <summary>
+    /// Check if entity exists in DB.
+    /// </summary>
+    /// <param name="filter">Filter.</param>
+    /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
+    Task<bool> IsExists(OperationWithObjectFilter filter);
```
DmyMi, line 50: `IsExists` does not sound good :) can change to just `Exists` or some other word with `Is` prefix :)

OutOfSchool/OutOfSchool.WebApi/Services/AverageRatings/AverageRatingService.cs
```diff
@@ -134,14 +141,25 @@ private async Task<IEnumerable<RatingDto>> GetNewRatings(DateTimeOffset lastLaun
...
+        workshopsIds = workshopsIds.Union(workshopsIdsFromOperations);
+
         var providersIds = GetUniqueProviderIdsByWorkshopIdsAsync(workshopsIds);
 
         Dictionary<Guid, Tuple<float, int>> averageRatings = new Dictionary<Guid, Tuple<float, int>>();
 
         averageRatings.AddRange(CalculateWorkshopsRatingsByEntityIdsAsync(workshopsIds));
         averageRatings.AddRange(CalculateProvidersRatingsByProvidersIdsAsync(providersIds));
 
+        foreach (var operation in operations)
```
DmyMi, line 158: maybe run this in transaction with what ever logic it runs in?

or it does not matter if we fail to delete a rating operation and just try it later some time?

DmyMi (approved):

OutOfSchool/OutOfSchool.DataAccess/Enums/OperationWithObjectEntityType.cs
alrom: I thought in the future we'll make a controller for ours front's tasks )))) But, OK.

OutOfSchool/OutOfSchool.WebApi/Services/AverageRatings/AverageRatingService.cs
alrom: The last one ))))

PR #1025, opened 2023-03-16
Added security HTTP headers.

andriitsylia:


alrom (approved):

OutOfSchool/OutOfSchool.Common/Extensions/Startup/SecurityHttpHeadersExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+                        .From("fonts.gstatic.com");
+                    builder.AddScriptSrc()
+                        .Self();
+                    builder.AddBaseUri()
+                        .Self();
+                    builder.AddImgSrc()
+                        .From("https://*")
+                        .Self()
+                        .Data();
+                    builder.AddCustomDirective("script-src-elem", "self");
```
DmyMi, line 45: this is the config for angular :) need to see what needs to be added for WebMVC/Razor pages on identity and should it be present in web api or not

OutOfSchool/OutOfSchool.IdentityServer/OutOfSchool.IdentityServer.csproj
```diff
@@ -30,6 +30,7 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="6.0.1" />
     <PackageReference Include="Microsoft.Extensions.Identity.Core" Version="6.0.6" />
     <PackageReference Include="Microsoft.IdentityModel.Tokens" Version="6.8.0" />
+    <PackageReference Include="NetEscapades.AspNetCore.SecurityHeaders" Version="0.18.0" />
```
DmyMi, line 33: if we import it to common, we automatically have it in every project that depends on common. No need to have it here

DmyMi (changes requested):

OutOfSchool/OutOfSchool.WebApi/OutOfSchool.WebApi.csproj
```diff
@@ -15,6 +15,7 @@
   <ItemGroup>
 	  <PackageReference Include="CsvHelper" Version="29.0.0" />
 	  <PackageReference Include="Microsoft.AspNetCore.SignalR.StackExchangeRedis" Version="6.0.9" />
+	  <PackageReference Include="NetEscapades.AspNetCore.SecurityHeaders" Version="0.18.0" />
```
DmyMi, line 18: same, do not need to import here

DmyMi (changes requested):

PR #1064, opened 2023-04-24
ReSendEmailConfirmation has been added

valerabad:


OutOfSchool/OutOfSchool.IdentityServer/Controllers/AccountController.cs
```diff
@@ -196,6 +196,33 @@ public async Task<IActionResult> SendConfirmEmail()
         return Ok();
     }
 
+    [HttpGet]
+    public async Task<IActionResult> ReSendEmailConfirmation()
+    {
+        var user = await userManager.FindByEmailAsync(User.Identity.Name);
```
VadymLevkovskyi, line 202: Couldn't there be an `NullReferenceException` thrown in `User.Identity.Name` without `[Authorize]`?

OutOfSchool/OutOfSchool.IdentityServer/Controllers/AccountController.cs
```diff
@@ -196,6 +196,33 @@ public async Task<IActionResult> SendConfirmEmail()
...
+
+        var token = await userManager.GenerateEmailConfirmationTokenAsync(user);
+        var returnUrl = "Login";
```
VadymLevkovskyi, line 205: `var returnUrl = "Login"` sounds like a constant, that shoud be set on higher level (not a local method variable)

OutOfSchool/OutOfSchool.IdentityServer/Controllers/AccountController.cs
```diff
@@ -196,6 +196,33 @@ public async Task<IActionResult> SendConfirmEmail()
...
+        var callBackUrl = Url.Action("EmailConfirmation", "Account", new { token, user.Email, returnUrl }, Request.Scheme);
```
VadymLevkovskyi, line 206: `"EmailConfirmation", "Account"` - another set of constants used in more than one place. Why not use `nameof(...)` ?

OutOfSchool/OutOfSchool.IdentityServer/Controllers/AccountController.cs
```diff
@@ -196,6 +196,33 @@ public async Task<IActionResult> SendConfirmEmail()
...
+        var email = user.Email;
+        var subject = localizer["Confirm email"];
+        var userActionViewModel = new UserActionViewModel
+        {
+            FirstName = user.FirstName,
+            LastName = user.LastName,
+            ActionUrl = callBackUrl,
+        };
+        var content = await renderer.GetHtmlPlainStringAsync(RazorTemplates.ConfirmEmail, userActionViewModel);
+        await emailSender.SendAsync(email, subject, content);
```
VadymLevkovskyi, line 217: `await emailSender.SendAsync(email, subject, content);` - I see no check for unsuccessful send. Just curious - how fails are processed?

OutOfSchool/OutOfSchool.IdentityServer/Controllers/AccountController.cs
```diff
@@ -196,6 +196,33 @@ public async Task<IActionResult> SendConfirmEmail()
...
+
+        var path = $"{HttpContext.Request.Path.Value}[{HttpContext.Request.Method}]";
```
VadymLevkovskyi, line 219: `var path = $"{HttpContext.Request.Path.Value}[{HttpContext.Request.Method}]";` is used pretty everywhere. Isn't it a right time to add at least private method for it (or even controller extension method if usages span across this controller?)

VadymLevkovskyi (commented):
Probably missed Authorize, mail send failure processing and minor code style issues.

OutOfSchool/OutOfSchool.IdentityServer/Controllers/AccountController.cs
valerabad: I've rechecked: in the nested method SendAsync() there is try...catch

DmyMi (approved):

PR #1068, opened 2023-04-26
Tsylia/1066 Unique Edrpou Ipn

andriitsylia:


OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
```diff
@@ -484,6 +484,11 @@ private protected async Task<ProviderDto> UpdateProviderWithActionBeforeSavingCh
             throw new ArgumentNullException(nameof(userId));
         }
 
+        if (await ExistsAnotherProviderWithTheSameEdrpouIpn(providerUpdateDto))
+        {
+            return null;
```
VadymLevkovskyi, line 489: Shouldn't any trace log be added here?

VadymLevkovskyi (approved):
Approved, though have minor question about handling errors. It looks like we are ignoring error reason that could be potentially harmful for prod cases debugging. 

I'm not sure what is OoS policy regarding this, so, @andriitsylia better check it with @DmyMi

OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
andriitsylia: Ok, I agree with you. I'll add logger.LogInformatiom() to this snippet of the code.

DmyMi:
@VadymLevkovskyi yes the policy is to at least log the exception

VadymLevkovskyi (approved):

OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
```diff
@@ -777,4 +784,14 @@ private async Task UpdateWorkshopsProviderStatus(Guid providerId, ProviderStatus
...
+    private async Task<bool> ExistsAnotherProviderWithTheSameEdrpouIpn(ProviderUpdateDto providerUpdateDto)
+    {
+        var providersWithTheSameEdrpouIpn = await providerRepository
+            .GetByFilter(x => x.EdrpouIpn == providerUpdateDto.EdrpouIpn)
+            .ConfigureAwait(false);
+
+        return providersWithTheSameEdrpouIpn.Any()
+            && providersWithTheSameEdrpouIpn.First().Id != providerUpdateDto.Id;
+    }
 }
```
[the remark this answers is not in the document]
andriitsylia: I did it.

PR #1082, opened 2023-05-05
Romanchuk / Change email

alrom:


OutOfSchool/OutOfSchool.IdentityServer/Controllers/AccountController.cs
```diff
@@ -141,6 +141,14 @@ public async Task<IActionResult> ConfirmChangeEmail(string userId, string email,
             return NotFound($"Changing email for user with ID: '{userId}' was not allowed.");
         }
 
+        var userByEmail = await userManager.FindByEmailAsync(email);
+        if (userByEmail is not null)
+        {
+            logger.LogError("{Path} User(email): {email} was found", path, email);
+
+            return View("Email/ConfirmEmailFailed", localizer[@"Email '{0}' is already used", email]);
```
VadymLevkovskyi, line 149: I'm bit concerned with this whole method.

A hacker having no creds could enumerate user `emails` and `ids` and could find (by response message analysis) if any exist. At this point `token` is not even checked! 

Suggestions are:
- be less specific in error messages - return same BadRequest without any details specified for all unauthorized (token is null/not verified) attempts having wrong data; 
- check token as early as possible (it could also eliminate response time check attack).

OutOfSchool/OutOfSchool.IdentityServer/Controllers/AccountController.cs
```diff
@@ -421,11 +429,11 @@ public async Task<IActionResult> ChangePassword(ChangePasswordViewModel model)
 
     [HttpGet]
     [Authorize]
-    private async Task<IActionResult> SendConfirmEmailProcess(User user, string razorTemplate, object passedData)
+    private async Task<IActionResult> SendConfirmEmailProcess(string action, User user, string razorTemplate, object passedData)
```
VadymLevkovskyi, line 432: Please remove misleading attributes for this method. Framework doesn't route to private methods and thus never checks their attributes

alrom: Done

VadymLevkovskyi (approved):

PR #1124, opened 2023-06-12
Romanchuk / Add CompetitiveSelection to Workshop

alrom:


OutOfSchool/OutOfSchool.WebApi/Services/ApplicationService.cs
```diff
@@ -492,11 +493,30 @@ public async Task<int> ChangeApprovedStatusesToStudying()
         return result;
     }
 
-    private static void UpdateStatus(ApplicationUpdate applicationDto, Application application, bool isUserProvider)
+    private static void UpdateStatus(ApplicationUpdate applicationDto, Application application, ICurrentUserService currentUserService)
```
DmyMi, line 496: `currentUserService` is a class field, so maybe having this method as static is unnecessary? what do we get for passing the class field as argument vs direct access as instance member? :)

OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs
```diff
@@ -519,6 +519,7 @@ public MappingProfile()
             .ForMember(dest => dest.WithDisabilityOptions, opt => opt.Ignore())
             .ForMember(dest => dest.AvailableSeats, opt => opt.Ignore())
             .ForMember(dest => dest.TakenSeats, opt => opt.Ignore())
+            .ForMember(dest => dest.CompetitiveSelection, opt => opt.Ignore())
```
DmyMi, line 522: more of a question to BA: should we show that the workshop has competitive selection to users on 'applications' screen so he/she won't need to click every workshop to find out

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Services/ApplicationService.cs
alrom: Fix it )))

PR #1186, opened 2023-08-07
Change soft delete for InstitutionStatus

alrom:


OutOfSchool/OutOfSchool.WebApi/Services/StatusService.cs
```diff
@@ -96,19 +97,20 @@ public async Task<InstitutionStatusDTO> Update(InstitutionStatusDTO dto)
     {
         logger.LogInformation($"Updating InstitutionStatus with Id = {dto?.Id} started.");
 
-        try
-        {
-            var institutionStatus = await repository.Update(mapper.Map<InstitutionStatus>(dto)).ConfigureAwait(false);
-
-            logger.LogInformation($"InstitutionStatus with Id = {institutionStatus?.Id} updated succesfully.");
+        var institutionStatus = await repository.GetById(dto.Id);
```
VadymLevkovskyi: (Performance) Isn't `.ConfigureAwait(false)` missed here?

VadymLevkovskyi: (Functionality) I don't see here check of `IsDeleted`, and seems `Update` could work with both deleted and not deleted entities. As I don't see Update method calling places, so asking you to consider if `Update` should be used for deleted entities.

VadymLevkovskyi (commented):
Looks good, please check comments

OutOfSchool/OutOfSchool.WebApi/Services/StatusService.cs
alrom: 1. Done. I thought it isn't necessary for Core.
2. Done. I planned to use it for both options. We discussed on the daily and I fixed it.

Thank you.

OutOfSchool/OutOfSchool.WebApi/Services/StatusService.cs
```diff
@@ -6,6 +6,7 @@
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Localization;
 using Microsoft.Extensions.Logging;
+using Nest;
```
VadymLevkovskyi: This using might be redundant. Please recheck

alrom: I did it

VadymLevkovskyi: Splendid!

VadymLevkovskyi (approved):

PR #1187, opened 2023-08-08
Add NumberOfRatings to WorkshopCard

alrom:


andriitsylia (approved):

OutOfSchool/OutOfSchool.WebApi/Services/StatisticService.cs
```diff
@@ -195,10 +195,13 @@ public async Task<IEnumerable<WorkshopCard>> GetPopularWorkshopsFromDatabase(int
     private async Task<List<WorkshopCard>> GetWorkshopsWithAverageRating(List<WorkshopCard> workshopsCards)
     {
         var averageRatings = await averageRatingService.GetByEntityIdsAsync(workshopsCards.Select(p => p.WorkshopId)).ConfigureAwait(false);
+        AverageRatingDto averageRatingDto = null;
```
VadymLevkovskyi: (Code style) This row looks bit excessive. You could just say `var averageRatingDto = averageRatings?....` in `foreach` cycle body and that's it. Don't worry, no performance penalties will be provided.

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
```diff
@@ -768,9 +768,13 @@ private async Task<List<T>> GetWorkshopsWithAverageRating<T>(List<T> workshops)
     {
         var averageRatings = await averageRatingService.GetByEntityIdsAsync(workshops.Select(p => p.WorkshopId)).ConfigureAwait(false);
 
+        AverageRatingDto averageRatingDto = null;
```
VadymLevkovskyi: (Same code style comment) Variable definition should be as close as possible to variable usage. Please get rid of this line and use `var averageRatingDto = ...` in cycle body

OutOfSchool/OutOfSchool.WebApi/Services/StatisticService.cs
alrom: Ok. Done.

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
alrom: Done

VadymLevkovskyi (approved):

PR #1189, opened 2023-08-08
FixAreaAdminGetByFilter

ZdorovenkoKyrylo:


OutOfSchool/OutOfSchool.WebApi/Services/AreaAdminService.cs
```diff
@@ -7,6 +7,7 @@
 using OutOfSchool.Services.Enums;
 using OutOfSchool.WebApi.Models;
 using System.Linq.Expressions;
+using System.Runtime.InteropServices.ComTypes;
```
VadymLevkovskyi: Is this `using` necessary?

OutOfSchool/OutOfSchool.WebApi/Services/AreaAdminService.cs
```diff
@@ -147,43 +155,22 @@ public async Task<SearchResult<AreaAdminDto>> GetByFilter(AreaAdminFilter filter
...
-                Logger.LogInformation($"Filter institutionId {filter.InstitutionId} and CATOTTGId {filter.CATOTTGId} " +
-                                      $"is not equals to logined Otg admin institutionId {areaAdminDto.InstitutionId} and CATOTTGId {areaAdminDto.CATOTTGId}");
-
-                return new SearchResult<AreaAdminDto>()
-                {
-                    TotalAmount = 0,
-                    Entities = default,
-                };
-            }
+            var regionAdminDto = await regionAdminService.GetByUserId(currentUserService.UserId);
```
VadymLevkovskyi: (Performance issue) As I get this project approach - almost all async methods calls are configured with `.ConfigureAwait(false)`. Please use it (at least in the code you've added)

OutOfSchool/OutOfSchool.WebApi/Services/AreaAdminService.cs
```diff
@@ -492,7 +479,7 @@ public async Task<bool> IsAreaAdminSubordinateRegionCreateAsync(string regionAdm
         return regionAdmin.InstitutionId == institutionId && regionAdmin.CATOTTGId == catottgId;
     }
 
-    private static Expression<Func<AreaAdmin, bool>> PredicateBuild(AreaAdminFilter filter)
+    private static Expression<Func<AreaAdmin, bool>> PredicateBuild(AreaAdminFilter filter, List<long> catottgs)
```
VadymLevkovskyi, line 481: (Open question) I see that some parts of filtering predicates are passed as `AreaAdminFilter` properties, but `catottgs` is passed as separate parameter. I can't check if adding this parameter as an `AreaAdminFilter` property will break other parts of the systme, but for isolated code I see here it looks like vital change. 

Please consider and redesign if it is valid.

OutOfSchool/OutOfSchool.WebApi/Services/AreaAdminService.cs
```diff
@@ -147,43 +155,22 @@ public async Task<SearchResult<AreaAdminDto>> GetByFilter(AreaAdminFilter filter
 
         filter ??= new AreaAdminFilter();
         ModelValidationHelper.ValidateOffsetFilter(filter);
-
+        var catottgs = new List<long>();
```
VadymLevkovskyi: (Code style) Please move variable declaration closer to it's usage 
(as far as I can see it is used first time in `if (currentUserService.IsRegionAdmin())`)

OutOfSchool/OutOfSchool.WebApi/Services/AreaAdminService.cs
```diff
@@ -519,7 +506,11 @@ public async Task<bool> IsAreaAdminSubordinateRegionCreateAsync(string regionAdm
             predicate = predicate.And(a => a.Institution.Id == filter.InstitutionId);
         }
 
-        if (filter.CATOTTGId > 0)
+        if (catottgs.Count > 0 && catottgs.Contains(filter.CATOTTGId))
+        {
+            predicate = predicate.And(a => a.Institution.Id == filter.InstitutionId && a.CATOTTG.Id == filter.CATOTTGId);
```
VadymLevkovskyi: Please refactor filtering logic.

What if `filter.InstitutionId == Guid.Empty` but `catottgs.Count > 0 && catottgs.Contains(filter.CATOTTGId)` is true? 
I see that `GetByFilter()` prevents this now, but there are no obstacles for it to be changed later. And then you'll get unnecessary request to db to return nothing.

What if `filter.InstitutionId != Guid.Empty` and `catottgs.Count > 0 && catottgs.Contains(filter.CATOTTGId)` is true ?
Then condition `a.Institution.Id == filter.InstitutionId` would be applied twice, once in the previous `if`, and second time here. 

As I can't say if double applying of same condition would be coalesced or not, I'd suggest to avoid doing it.

OutOfSchool/OutOfSchool.DataAccess/Models/Application.cs
```diff
@@ -30,4 +32,7 @@ public class Application : IKeyedEntity<Guid>
     public virtual Child Child { get; set; }
 
     public virtual Parent Parent { get; set; }
+
+    [NotMapped]
+    public static readonly ApplicationStatus[] ValidApplicationStatuses = { ApplicationStatus.Approved, ApplicationStatus.StudyingForYears, ApplicationStatus.Completed};
```
VadymLevkovskyi: I see no usage for this field, might be redundant.

VadymLevkovskyi (commented):
Plesae check comments

OutOfSchool/OutOfSchool.WebApi/Services/AreaAdminService.cs
ZdorovenkoKyrylo: I see, in AreaAdminFilter all the properties are public and they can be changed and shown practically everywhere. I think there is no necessary to add a list with `catottgs` in this filter because it fills up just in `GetByFilter` method at once and nowhere else. Method `PredicateBuilder` was created mostly to decrease the amound of code in `GetByFilter` and has just 1 usage

VadymLevkovskyi (approved):

PR #1195, opened 2023-08-15
#1179 added sorting by status

nazarstig:


alrom:
In the previous PR sorting had a reverse ordering or not? Maybe, change indexes to reversing and start them from 10. Like this: NeverLogged = 30, Accepted = 20, Blocked = 10. And don't forget about migration for updating existing data in DB.

DmyMi (approved):

PR #1196, opened 2023-08-15
Disable soft delete for Directions

alrom:


OutOfSchool/OutOfSchool.DataAccess/Models/Direction.cs
```diff
@@ -17,5 +17,7 @@ public class Direction : IKeyedEntity<long>
     [MaxLength(500)]
     public string Description { get; set; } = string.Empty;
 
+    public bool IsDeleted { get; set; }
```
VadymLevkovskyi, line 20: (Design considerations) As I can see it's common pattern to have `IsDeleted` property as a flag. Also there are lot of code checking it for this and other similar DTOs. Isn't it a reason to separate and use `ISoftDeleted` (or suggest your name) interface?

OutOfSchool/OutOfSchool.DataAccess/OutOfSchoolDbContext.SoftDelete.cs
```diff
@@ -58,7 +58,7 @@ private void ApplySoftDelete(ModelBuilder builder)
             .ApplySoftDelete<CompanyInformation>()
             .ApplySoftDelete<CompanyInformationItem>()
             .ApplySoftDelete<DateTimeRange>()
-            .ApplySoftDelete<Direction>()
+            //.ApplySoftDelete<Direction>()
```
VadymLevkovskyi, line 61: I'm not insisting, but maybe it's right time to remove this commented out code?

alrom: We planned to clear it after finishing changing all entities. Commented lines show ours progress )))

VadymLevkovskyi: Code as a backlog? I'll steal this idea but would never ever use it )
And thanks for explanation

OutOfSchool/OutOfSchool.DataAccess/Models/Direction.cs
alrom: Ok. We can use it in the EntityRepositoryBase to add predicate for queries. If you have another view how we can use this interface, please, write it. Thanks.

VadymLevkovskyi: Sounds promising, Let's start with it and them see if it is beneficial

VadymLevkovskyi (approved):

PR #1201, opened 2023-08-18
Add SoftDeleted repository and change Directions

alrom:


OutOfSchool/OutOfSchool.DataAccess/Repository/EntityRepositorySoftDeleted.cs
```diff
@@ -0,0 +1,134 @@
...
+    {
+        predicate = GetWhereExpression(predicate);
+        return base.GetByFilterNoTracking(predicate, includeProperties);
+    }
+
+    /// <inheritdoc/>
+    public override Task<TEntity> GetById(TKey id) => dbSet.FirstOrDefaultAsync(x => !x.IsDeleted && x.Id.Equals(id));
+
+    /// <inheritdoc/>
+    public override Task<int> Count(Expression<Func<TEntity, bool>> where = null)
```
VadymLevkovskyi: (Code style - not insisting, just general recommendation) Please try to avoid usage of keywords as identifiers. 
Argument `where` here and in subsequent methods could be renamed to `condition` / `whereExpression` / other variant.

OutOfSchool/OutOfSchool.DataAccess/Repository/EntityRepositorySoftDeleted.cs
```diff
@@ -0,0 +1,134 @@
...
+        private IDictionary<ParameterExpression, ParameterExpression> ParameterReplacements { get; set; }
+
+        public ExpressionParameterReplacer(
+            IList<ParameterExpression> fromParameters,
+            IList<ParameterExpression> toParameters)
+        {
+            ParameterReplacements = new Dictionary<ParameterExpression, ParameterExpression>();
+
+            for (int i = 0; i != fromParameters.Count && i != toParameters.Count; i++)
+            { ParameterReplacements.Add(fromParameters[i], toParameters[i]); }
```
VadymLevkovskyi: (Fundamental code design) Please check suggested code. Maybe it's less verbose way to fill that dictionary

```suggestion
            ParameterReplacements = fromParameters.Zip(toParameters, (k, v) => new { k, v })
                .ToDictionary(x => x.k, x => x.v);
```

OutOfSchool/OutOfSchool.DataAccess/Repository/SensitiveEntityRepositorySoftDeleted.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using OutOfSchool.Services.Models;
+
+namespace OutOfSchool.Services.Repository;
+
+public class SensitiveEntityRepositorySoftDeleted<T> : EntityRepositorySoftDeleted<Guid, T>, ISensitiveEntityRepositorySoftDeleted<T>
```
VadymLevkovskyi, line 6: (Just wondering) Is there any necessity for this specific class? 
It adds no logic, and reuses `EntityRepositorySoftDeleted` in whole, and the only tiny thing is fixation of generic type of entity key as Guid.

VadymLevkovskyi (approved):
See no issues with logic. Have few comments (that might be ignored) about code style/design

OutOfSchool/OutOfSchool.DataAccess/Repository/EntityRepositorySoftDeleted.cs
alrom: Ok. Done.

OutOfSchool/OutOfSchool.DataAccess/Repository/SensitiveEntityRepositorySoftDeleted.cs
alrom: I agree. But we have the same approach with the SensitiveEntityRepository and I used it.

OutOfSchool/OutOfSchool.DataAccess/Repository/EntityRepositorySoftDeleted.cs
alrom: Perfect. I checked it. I left code from the internet sample without refactoring. I forgot about zip. Thank you.

OutOfSchool/OutOfSchool.DataAccess/Repository/SensitiveEntityRepositorySoftDeleted.cs
VadymLevkovskyi: Closing this thread as possible changes go out of the task scope. 
At same time I'm not convinced that usage of inheritance is beneficial here.

OutOfSchool/OutOfSchool.DataAccess/Repository/EntityRepositorySoftDeleted.cs
VadymLevkovskyi: Looks like it was present in too many places. Thanks for thorough work!

OutOfSchool/OutOfSchool.DataAccess/Repository/EntityRepositorySoftDeleted.cs
VadymLevkovskyi: Thanks for taking this variant, and NP, my code is not ideal. 
It is worse from performance perspective, but better (I hope) from cognitive perspective. 
I usually choose second option, and only in rare cases (when it is proven that we have a real bottleneck) write more performant code. Your mileage might vary, that's why I wasn't insisting, just asking for your opinion.

PR #1202, opened 2023-08-21
Tsylia / Disable C(R)UD and Block operations at the blocked providers

andriitsylia:
Also automatically block corresponding deputies and provider admins.

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
@@ -1,4 +1,5 @@
 ﻿using Microsoft.AspNetCore.Mvc;
+using OutOfSchool.Services.Models;
 using OutOfSchool.WebApi.Common;
 using OutOfSchool.WebApi.Models;
 using OutOfSchool.WebApi.Models.Application;
@@ -258,11 +259,6 @@ public class ApplicationController : ControllerBase
     [HttpPost]
     public async Task<IActionResult> Create(ApplicationCreate applicationDto)
     {
-        if (await IsCurrentUserBlocked())
-        {
-            return StatusCode(403, "Forbidden to create the application by the blocked provider.");
-        }
-
         if (applicationDto == null)
         {
             return BadRequest("Application is null.");
@@ -271,6 +267,16 @@ public class ApplicationController : ControllerBase
         if (!ModelState.IsValid)
         {
             return BadRequest(ModelState);
+        }
+
+        if (await IsWorkshopOrProviderBlocked(applicationDto.WorkshopId).ConfigureAwait(false))
+        {
+            return StatusCode(403, "Forbidden to create the application at the blocked workshop/provider");
+        }
+
+        if (await IsCurrentUserBlocked())
+        {
+            return StatusCode(403, "Forbidden to create the application by the blocked provider.");
         }
 
         try
@@ -314,6 +320,11 @@ public class ApplicationController : ControllerBase
     [HttpPut]
     public async Task<IActionResult> Update(ApplicationUpdate applicationDto)
     {
+        if (await IsWorkshopOrProviderBlocked(applicationDto.WorkshopId).ConfigureAwait(false))
+        {
+            return StatusCode(403, "Forbidden to update the application at the blocked workshop/provider");
+        }
+
         if (await IsCurrentUserBlocked())
         {
             return StatusCode(403, "Forbidden to update the application by the blocked provider.");
@@ -382,5 +393,16 @@ public class ApplicationController : ControllerBase
         var userId = GettingUserProperties.GetUserId(User);
 
         return await userService.IsBlocked(userId);
+    }
+
+    private async Task<bool> IsWorkshopOrProviderBlocked(Guid workshopId)
+    {
+        var isWorkshopBlocked = await workshopService.isBlocked(workshopId).ConfigureAwait(false);
+
+        var providerId = await providerService.GetProviderIdForWorkshopById(workshopId).ConfigureAwait(false);
+
+        var isProviderBlocked = await providerService.IsBlocked(providerId).ConfigureAwait(false);
+
+        return isWorkshopBlocked || isProviderBlocked;
     }
 }
--- a/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
@@ -264,19 +264,19 @@ public class ApplicationController : ControllerBase
             return BadRequest("Application is null.");
         }
 
-        if (!ModelState.IsValid)
+        if (await IsWorkshopBlocked(applicationDto.WorkshopId).ConfigureAwait(false))
         {
-            return BadRequest(ModelState);
-        }
-
-        if (await IsWorkshopOrProviderBlocked(applicationDto.WorkshopId).ConfigureAwait(false))
-        {
-            return StatusCode(403, "Forbidden to create the application at the blocked workshop/provider");
+            return StatusCode(403, "Forbidden to create the application at the blocked workshop.");
         }
 
         if (await IsCurrentUserBlocked())
         {
-            return StatusCode(403, "Forbidden to create the application by the blocked provider.");
+            return StatusCode(403, "Forbidden to create the application by the blocked user.");
+        }
+
+        if (!ModelState.IsValid)
+        {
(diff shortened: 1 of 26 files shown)


alrom (approved):

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -303,6 +320,16 @@ public async Task<IActionResult> Create(ApplicationCreate applicationDto)
     [HttpPut]
     public async Task<IActionResult> Update(ApplicationUpdate applicationDto)
     {
+        if (await IsWorkshopOrProviderBlocked(applicationDto.WorkshopId).ConfigureAwait(false))
```
VadymLevkovskyi: (NRE) `applicationDto` might be null here. Check for null is already present in this method but it is performed later. Please change order of checks

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -360,4 +387,22 @@ public async Task<IActionResult> AllowedToReview(Guid parentId, Guid workshopId)
...
+        return await userService.IsBlocked(userId);
+    }
+
+    private async Task<bool> IsWorkshopOrProviderBlocked(Guid workshopId)
+    {
+        var isWorkshopBlocked = await workshopService.isBlocked(workshopId).ConfigureAwait(false);
+
+        var providerId = await providerService.GetProviderIdForWorkshopById(workshopId).ConfigureAwait(false);
+
+        var isProviderBlocked = await providerService.IsBlocked(providerId).ConfigureAwait(false);
```
VadymLevkovskyi: (Probably performance) There are 3 calls to services that might be potentially transformed to 3 calls to db. As I can't check now if these calls are cached, I'm asking you to consider related performance issues. 

I do not insist on massive rewriting, just asking if it could be easily fixed, or should be registered as tech debt, or could be ignored.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/BlockedProviderParentController.cs
```diff
@@ -44,8 +56,17 @@ public class BlockedProviderParentController : ControllerBase
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     public async Task<IActionResult> Block(BlockedProviderParentBlockDto blockedProviderParentBlockDto)
     {
-        var userId = GettingUserProperties.GetUserId(HttpContext);
-        var result = await blockedProviderParentService.Block(blockedProviderParentBlockDto, userId).ConfigureAwait(false);
+        if (await IsProviderBlocked(blockedProviderParentBlockDto.ProviderId).ConfigureAwait(false))
```
VadymLevkovskyi, line 65: (NRE) `blockedProviderParentBlockDto` might be `null` here. Please check it beforehand

Same in `Unblock()` method

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ProviderAdminController.cs
```diff
@@ -50,6 +57,16 @@ public async Task<ActionResult> Create(CreateProviderAdminDto providerAdmin)
     {
         logger.LogDebug($"{path} started. User(id): {userId}.");
 
+        if (await IsProviderBlocked(providerAdmin.ProviderId).ConfigureAwait(false))
```
VadymLevkovskyi, line 66: (NRE) `providerAdmin` might be `null`. Please ensure it is preliminary checked.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopController.cs
```diff
@@ -247,9 +251,14 @@ public async Task<IActionResult> GetByFilter([FromQuery] WorkshopFilter filter,
     [HttpPost]
     public async Task<IActionResult> Create(WorkshopDTO dto)
     {
-        if (await providerService.IsProviderBlocked(dto.ProviderId).ConfigureAwait(false))
+        if (await IsProviderBlocked(dto.ProviderId).ConfigureAwait(false))
```
VadymLevkovskyi, line 260: (NRE) `dto` might be `null`, please check it before using.

Same for `Update()` and `UpdateStatus()`

OutOfSchool/OutOfSchool.WebApi/Services/Database/IWorkshopService.cs
```diff
@@ -145,4 +145,11 @@ Task<SearchResult<T>> GetByProviderId<T>(Guid id, ExcludeIdFilter filter)
     /// <param name="workshopId">WorkshopId for which we need to get provider owner entity.</param>
     /// <returns>A <see cref="Task"/> representing the result of the asynchronous operation.</returns>
     Task<Guid> GetWorkshopProviderOwnerIdAsync(Guid workshopId);
+
+    /// <summary>
+    ///  Returns isBlocked status of the workshop.
+    /// </summary>
+    /// <param name="workshopId">WorkshopId for which we need to get status.</param>
+    /// <returns>A <see cref="Task"/> representing the result of the asynchronous operation.</returns>
+    Task<bool> isBlocked(Guid workshopId);
```
VadymLevkovskyi: (Naming) Please use `PascalCase` while naming methods. Thus your `isBlocked` should be `IsBlocked` (starting with uppercase **I**)

OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
```diff
@@ -365,7 +368,15 @@ public async Task<ResponseDto> Block(ProviderBlockDto providerBlockDto)
             false,
             false);
 
+        logger.LogInformation("Block/Unblock the particular provider admins and deputy providers belonging to the Provider starts.");
+
+        _ = await providerAdminService
```
VadymLevkovskyi, line 374: (Data consistency) Why you are ignoring method result? As I see from inplementation there might be a case, when db data are changed but identityServer request is failed. Is it ok to continue silently?

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
andriitsylia: Done.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/BlockedProviderParentController.cs
andriitsylia: Done.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ProviderAdminController.cs
andriitsylia: Done.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopController.cs
andriitsylia: Done.

OutOfSchool/OutOfSchool.WebApi/Services/Database/IWorkshopService.cs
andriitsylia: :) I have to go back to the basics. Shame on me for that. :(
Fixed it.

OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
andriitsylia: We've discussed this code at our meeting. But we didn't have a common point of view on what we had to do in case some of the provider admins or deputies wouldn't be blocked. We decided to do additional checks on other controllers. 
Create TODO.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
andriitsylia: I have performed stricter checks in this case. Made changes. We only need to check workshops because blocking the provider also blocks all his workshops.  Thus, there is no need to check the provider.

OutOfSchool/OutOfSchool.WebApi/Services/ProviderService.cs
VadymLevkovskyi: Thanks for the explanation.

VadymLevkovskyi (approved):

PR #1212, opened 2023-08-30
Kymachynskyi / Version 2 of Add number of unread messages on workshop card

DKymachynskyi:


OutOfSchool/OutOfSchool.WebApi/Services/WorkshopServicesCombiner.cs
```diff
@@ -228,10 +228,9 @@ public async Task<List<ShortEntityDto>> GetWorkshopListByProviderAdminId(string
     }
 
     /// <inheritdoc/>
-    public async Task<SearchResult<T>> GetByProviderId<T>(Guid id, ExcludeIdFilter filter)
-        where T : WorkshopBaseCard
+    public async Task<SearchResult<WorkshopProviderViewCard>> GetByProviderId(Guid id, ExcludeIdFilter filter)
```
VadymLevkovskyi: (Code style) 5 lines of method body could be rewritten nto 1 line, with removing excessive `async`/`await` keywords and `ConfigureAwait` that (I hope) resides on controller level
```
public Task<SearchResult<WorkshopProviderViewCard>> GetByProviderId(Guid id, ExcludeIdFilter filter)
    => workshopService.GetByProviderId(id, filter);
```

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
```diff
@@ -6,6 +6,7 @@
 using H3Lib.Extensions;
 using Nest;
 using OutOfSchool.Common.Enums;
+using OutOfSchool.ElasticsearchData.Models;
```
VadymLevkovskyi: Please confirm that this `using OutOfSchool.ElasticsearchData.Models;` is not excessive

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
```diff
@@ -266,22 +269,100 @@ public async Task<SearchResult<T>> GetByProviderId<T>(Guid id, ExcludeIdFilter f
...
-                    : (x.ProviderId == id && x.Id != filter.ExcludedId))
-            .ToListAsync()
-            .ConfigureAwait(false);
-
-        logger.LogInformation(!workshops.Any()
+                    : (x.ProviderId == id && x.Id != filter.ExcludedId));
+
+        var chatrooms = roomRepository.Get(
+            includeProperties: $"{nameof(ChatRoomWorkshop.ChatMessages)}",
+            where: x => x.ChatMessages.Where(x => x.ReadDateTime == null && !x.SenderRoleIsProvider).Any());
```
VadymLevkovskyi: GitHub Action says `The best overload for 'Get' does not have a parameter named 'where'`. Probably it should be `whereExpression`?

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
```diff
@@ -266,22 +269,100 @@ public async Task<SearchResult<T>> GetByProviderId<T>(Guid id, ExcludeIdFilter f
...
+
+        var query = from w in workshops
```
VadymLevkovskyi: (Code style) This method demonstrates both LINQ syntax/style - chained (see `workshopsWithUnreadMessages.Select` below) and query one. 
General recommendation is to use only one, my own recommendation (taking code maintainability into account) is to prefer chained.
Please redesign your code.

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
```diff
@@ -266,22 +269,100 @@ public async Task<SearchResult<T>> GetByProviderId<T>(Guid id, ExcludeIdFilter f
...
+                    join c in chatrooms
+                    on w.Id equals c.WorkshopId
+                    into g
+                    from cg in g.DefaultIfEmpty()
+                    select new
+                    {
+                        w.Id,
+                        w.Title,
+                        w.Phone,
```
VadymLevkovskyi: (Performance and readability) 
1. Excessive fields
You are asking `query` to return lot of fields that are not used anywhere (I won'tname all, but `Phone` and `DateTimeRanges` aren't used). 

This is bad from **performance** perspective - because they are transfered from db to the app.
This is bad from **readability** and **code maintainability** perspective - your anonymous object creation takes 30 lines of your code

2. Memory pressure, excessive objects and GroupJoin.
Theory: `GroupJoin` (see [complex query operators](https://learn.microsoft.com/en-us/ef/core/querying/complex-query-operators#groupjoin)) as well as your `into` operator doesn't execute on db side (`EF Core doesn't translate GroupJoin` - from mentioned article). So initially you obtain all records from db and then on C# side EF does the group join. 
In your case: for the each grouping (done on C# side) you have intermediate anonymous object (`select new { w.Id ....`), that stores necessary data. And immediately after that (smaller) part of that data is mapped into `WorkshopProviderViewCard`, and then all that anon objects are no longer used.

This is bad from **performance** perspective - you have to copy data twice (to anon object and then to `WorkshopProviderViewCard`).
This is bad from **memory usage** perspective - you are bothering GC with ephemeral allocations (all that anon objects).
This is bad from **readability** and **code maintainability** perspective - extra 25 lines of code are needed to copy data second time, and it's hard to see in one place if previously mentioned excessive fields could be removed.

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
```diff
@@ -266,22 +269,100 @@ public async Task<SearchResult<T>> GetByProviderId<T>(Guid id, ExcludeIdFilter f
...
+                        w.Status,
+                        w.AvailableSeats,
+                        w.Provider,
+                        w.InstitutionHierarchy,
+                        w.Address,
+                        w.ProviderAdmins,
+                        w.Teachers,
+                        w.Applications,
+                        w.DateTimeRanges,
+                        ChatRooms = cg,
```
VadymLevkovskyi: (Performance, readability, ...) The query is "irrelevant". 

`ChatRooms` are never used. 

From all this grouping story you need only count! of cg.ChatMessages. Please redesign query using [GroupBy](https://learn.microsoft.com/en-us/ef/core/querying/complex-query-operators#groupby) that is expected to be executed on the server side. Check the generated sql. If it is not doing all grouping/counting on server side (I'm bit suspicious about `chatrooms` filter) - find your LINQ way or finally write that piece of sql code.

VadymLevkovskyi (commented):
In short - query should be redesigned. It's working now, though it is neither fast, nor cute.

OutOfSchool/OutOfSchool.WebApi/Services/WorkshopServicesCombiner.cs
DKymachynskyi: It resides on controller level, so I removed them. Thank you!

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
DKymachynskyi: Thank you for your patience and detailed description of my bad😔 code. I tried to fix it.

VadymLevkovskyi: Np and thanks for fixes. Have you checked generated SQL?

DKymachynskyi: Yes, on Saturday I found it large with a duplicate subquery, but I thought it's ok🤔 (at 12 am). Today, after your question, I double-checked and understood where the duplicate subquery comes from - "whereExpression" in roomRepository.Get. 
I remove "whereExpression" and got SQL:
SELECT `c`.`WorkshopId`, COUNT(*) AS `UnreadMessageCount`
FROM `ChatRoomWorkshops` AS `c`
INNER JOIN (
    SELECT `c0`.`ChatRoomId`, `c0`.`ReadDateTime`, `c0`.`SenderRoleIsProvider`
    FROM `ChatMessageWorkshops` AS `c0`
    WHERE NOT (`c0`.`IsDeleted`)
) AS `t` ON `c`.`Id` = `t`.`ChatRoomId`
WHERE NOT (`c`.`IsDeleted`) AND (`t`.`ReadDateTime` IS NULL AND NOT (`t`.`SenderRoleIsProvider`))
GROUP BY `c`.`WorkshopId`
So, I will commit removing unnecessary whereExpression.

VadymLevkovskyi: "Midnight coding" ))
That's cool that you've found it! Hope it wasn't too boring

VadymLevkovskyi (approved):
At the moment see no issues, thanks for polishing the PR!

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
DKymachynskyi: I'm interested) Thanks for your mentorship!

DKymachynskyi:
@VadymLevkovskyi I made one more commit, where I added a condition to the query that restricts the count of unread messages only for workshops obtained from the workshops table in this method. Please review it.

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
```diff
@@ -279,9 +279,13 @@ public async Task<SearchResult<WorkshopProviderViewCard>> GetByProviderId(Guid i
...
         var workshopProviderViewCards = mapper.Map<List<WorkshopProviderViewCard>>(workshops);
 
+        var workshopsIds = workshops.Select(x => x.Id).ToList();
+
         var query = chatrooms
             .SelectMany(room => room.ChatMessages, (room, message) => new { room, message })
-            .Where(entry => entry.message.ReadDateTime == null && !entry.message.SenderRoleIsProvider)
+            .Where(entry => entry.message.ReadDateTime == null
+                        && !entry.message.SenderRoleIsProvider
+                        && workshopsIds.Any(x => x == entry.room.WorkshopId))
```
VadymLevkovskyi: (Coding) Isn't `workshopsIds.Contains(entry.room.WorkshopId)` simpler?

DKymachynskyi: Yes, it is simpler. Thank you!

VadymLevkovskyi (approved):

PR #1217, opened 2023-09-03
Change soft deleted

alrom:


VadymLevkovskyi (approved):
Changed parts of code are approved though I don't see any changes in tests. 
Was it "magically" fixed or there are no tests?

PR #1227, opened 2023-09-06
Add additional fields

ZdorovenkoKyrylo:


VadymLevkovskyi (approved):
Approved, though I'd recommend to work on your naming skills :) (PR name)

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs
```diff
@@ -468,6 +468,8 @@ public MappingProfile()
                    .ForMember(dest => dest.Email, opt => opt.MapFrom(src => src.User.Email))
                    .ForMember(dest => dest.CATOTTGCategory, opt => opt.MapFrom(src => src.CATOTTG.Category))
                    .ForMember(dest => dest.CATOTTGName, opt => opt.MapFrom(src => src.CATOTTG.Name))
+                   .ForMember(dest => dest.RegionId, opt => opt.MapFrom(src => src.CATOTTG != null ? src.CATOTTG.Parent != null ? src.CATOTTG.Parent.Parent != null ? src.CATOTTG.Parent.Parent.Id :0 :0 :0))
```
VadymLevkovskyi, line 471: (Code style) I might be wrong but could you try instead of lot of ternary operators use elvis and null-coalescing operators, like
```suggestion
                   .ForMember(dest => dest.RegionId, opt => opt.MapFrom(src => src.CATOTTG?.Parent?.Parent ?? 0))
```

ZdorovenkoKyrylo: Using such case I`ve caught the following error:
"An expression tree lambda cannot contain conditional access expressions".
Developers on StackOverflow suggest using something similar to what I tried to realize

VadymLevkovskyi: Thanks for sharing! (Was kinda expecting this but wanted to clarify)
(Probably we need to suggest PR for StackOverflow mapper? ;) )

VadymLevkovskyi: Funny enough is that I've checked it on a small test project ([see on SharpLab](https://sharplab.io/#v2:C4LglgNgPgAgTARgLAChUwAwAICCBXYAewFkBDAB3IFMAnAblVQDtSBbKgZ3NIGMrcCJCtRoBRAB4iw7JsA6oA3qiwqsy1fCwwEAdnUqlKVca1wsAZTwAjS1YAiRfScMnXWgMxYwsrADk8rFa0WApYAOZUwHRYHJHRAL5OxoloRm6atg6EIWppbhqetplEAPwW1sXZoRFRMXFYKfkpSRpmWTktrjCF1llllTk10bG1jW7NeSaaxIQAJlQQnSFLxt1ePv6BwdX1IwlLE/naAGxaACxYZN4AFNoYANoAulikNGEcAJRLLvmqAG6vLByYBYAC8WCYVAA7lgstcPgxUr9/oCeIQmAAzMBhMEQ6GXYS0ADC6KxYTwNFIwDA6OuKzcPAxONBAD4sIywgA6Ik0KhUqhkcgAHiyABpLnMFiz4fTkZyAGKEGjEKhbGh0ybI5HzDgg1lYHXATmbII0UWyrUqQjkPVs61GwXymiEVjXDg0Hhgtnunicyp+iq9IjGgKmr6ay2qcOWhEWgE0LCsQkJ8FozHY7m8/mCkTwxFxwGsSUQXFJyi0TmCoUzeYQaXA2NIrXaACc1wAJAAiABqpAgeH4YA4ISLtZDavinawCKwAHpZ1goYQ8BBZlodFhsFCABZUJgvJgATywndsU6HEJXJdITDXPoD9mDRSDhHHpq8GIh6MHw6YV4OTgpPEQA===)) 
and it seems you even don't need to check that nulls at all. 
(Unfortunately it wouldn't compile on SharpLab, b/c you need AutoMapper pckg)

PR #1228, opened 2023-09-06
Remove custom request id and use header propagation

DmyMi:
Propagate request tracing from inbound proxy to webapi to authserver.
Current http context enricher extracts `Request-Id` header automatically. Some other implementations use `X-Request-Id` so we propagate both to http client.

### DIFF
--- a/OutOfSchool/OutOfSchool.AuthCommon/Services/CommonMinistryAdminService.cs
+++ b/OutOfSchool/OutOfSchool.AuthCommon/Services/CommonMinistryAdminService.cs
@@ -1,6 +1,7 @@
 ﻿using AutoMapper;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Localization;
+using OutOfSchool.AuthCommon.Extensions;
 using OutOfSchool.AuthCommon.Services.Password;
 using OutOfSchool.Common.Models;
 using OutOfSchool.RazorTemplatesData.Models.Emails;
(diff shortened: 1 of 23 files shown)


OutOfSchool/OutOfSchool.AuthCommon/Services/CommonMinistryAdminService.cs
```diff
@@ -90,9 +88,9 @@ async Task<ResponseDto> CreateMinistryAdminOperation()
                     await transaction.RollbackAsync();
 
                     logger.LogError(
-                        $"Error happened while creation MinistryAdmin. Request(id): {requestId}" +
-                        $"User(id): {userId}" +
-                        $"{string.Join(Environment.NewLine, result.Errors.Select(e => e.Description))}");
+                        "Error happened while creation MinistryAdmin. User(id): {UserId}. {Error}",
+                        userId,
+                        string.Join(Environment.NewLine, result.Errors.Select(e => e.Description)));
```
VadymLevkovskyi: (Code style) The next code pattern 
```
string.Join(Environment.NewLine, result.Errors.Select(e => e.Description)))
```
seems to be duplicated in subsequent calls. Couldn't it be seprated as an extension method?

VadymLevkovskyi: Also same thing found in `ProviderAdminService`

OutOfSchool/OutOfSchool.WebApi/Startup.cs
```diff
@@ -392,5 +395,11 @@ public static void AddApplicationServices(this WebApplicationBuilder builder)
             .AddDbContextCheck<OutOfSchoolDbContext>(
                 "Database",
                 tags: new[] { "readiness" });
+
+        services.AddHeaderPropagation(options =>
+        {
+            options.Headers.Add("Request-Id");
```
VadymLevkovskyi: (Logic) Don't you want to add some default value builder? (As a result for case when `Request-Id` is empty on one step you could rehidrate it by some GUID and track all subsequent steps)

VadymLevkovskyi (approved):
Aproved with minor suggestions (though they may be rejected or done in separate PR)

DmyMi:
@VadymLevkovskyi thank you, fixed comments in new commit.

VadymLevkovskyi:
> @VadymLevkovskyi thank you, fixed comments in new commit.

Not at all @DmyMi, changes look nice!

PR #1233, opened 2023-09-11
Change soft delete for acievements, acievement types and acievement teachers

alrom:


OutOfSchool/OutOfSchool.WebApi/Services/AchievementService.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using AutoMapper;
 using Microsoft.Extensions.Localization;
+using NuGet.Protocol.Core.Types;
```
VadymLevkovskyi: Is this `using` necessary?

VadymLevkovskyi (approved):
Logic is approved, there is only minor question about (probably excessive) using

OutOfSchool/OutOfSchool.WebApi/Services/AchievementService.cs
alrom: AAAAAAAAAAAAAAA. I didn't include it manually. I didn't see it before making this PR ))) Deleted. Thanks.

PR #1241, opened 2023-09-13
Change SoftDelete for AverageRatings, Ratings and QuartzJobs.

alrom:


OutOfSchool/OutOfSchool.WebApi/Services/RatingService.cs
```diff
@@ -209,7 +199,9 @@ public async Task<RatingDto> Update(RatingDto dto)
 
         if (await CheckRatingUpdate(dto).ConfigureAwait(false))
         {
-            var rating = await ratingRepository.Update(mapper.Map<Rating>(dto)).ConfigureAwait(false);
+            var rating = await ratingRepository.GetById(dto.Id).ConfigureAwait(false);
+            mapper.Map(dto, rating);
+            rating = await ratingRepository.Update(rating).ConfigureAwait(false);
```
VadymLevkovskyi: (DRY - I'm not insisting, please consider if you like it) 
I see such or similar code in many places. What it does - read from db by `dto.Id` into `entity`, take all changes from `dto` into `entity`, save it back to db. Isn't it a time to extract some helper? This looks to possible to be put on `IEntityRepositoryBase` level:

```
public async Task<TEntity> ReadAndUpdateWith<T>(TKey id, T dto, Func<T, TEntity, TEntity> map) 
{
	ArgumentNullException.ThrowIfNull(dto);

	var entity = await GetById(id).ConfigureAwait(false);
	if (entity is null)
	{
		var name = typeof(TEntity).Name;
		throw new DbUpdateConcurrencyException($"Updating failed. {name} with Id = {id} doesn't exist in the system.");
	}

	return await Update(map(dto, entity)).ConfigureAwait(false);
}
```
---
BTL, it worth mentioning from models design perspective that having twofold representation of data as `entity` and `dto` we have only `entity`-level "key" constraint (`IKeyedEntity`), but nothing for the `dto`. Thus having simple `IKeyed<TKey>` introduced and implemented by dtos could simplify beforementioned method to:
```public async Task<TEntity> ReadAndUpdateWith<T>(T dto, Func<T, TEntity, TEntity> map) where T: IKeyed<TKey>```
where you could avoid passing id as separate argument.

And in thinking deeper even in more freaky style - as dtos are usually bound to entities you could constraint them by
`IDto<TEntity> where TEntity: IKeyedEntity<TKey>`
and thus have a power to check in generic code like written above that `T` and `TEntity` are "friends":
```public async Task<TEntity> ReadAndUpdateWith<TDto>(TDto dto, Func<TDto, TEntity, TEntity> map) where TDto: IDto<TEntity>```

And additionally, `CreateMap`s in automapper profile could be simplified. And who knows what else )

File: OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs [hunk 2]
```diff
@@ -571,7 +571,9 @@ public MappingProfile()
         CreateMap<Rating, RatingDto>()
             .ForMember(dest => dest.FirstName, opt => opt.Ignore())
             .ForMember(dest => dest.LastName, opt => opt.Ignore());
+
         CreateMap<RatingDto, Rating>()
+            .ForMember(dest => dest.IsDeleted, opt => opt.Ignore())
```
VadymLevkovskyi: (DRY - again, I'm not insisting on changes) 
Please consider whether such helper method could ease repetitions
```
    public IMappingExpression<TSource, TDestination> CreateSoftDeletedMap<TSource, TDestination>()
        where TSource : class
        where TDestination : class, ISoftDeleted
        => CreateMap<TSource, TDestination>()
            .ForMember(dest => dest.IsDeleted, opt => opt.Ignore());
```
usage:
```
    CreateSoftDeletedMap<RatingDto, Rating>()
        .ForMember(dest => dest.Parent, opt => opt.Ignore());
```

VadymLevkovskyi (approved):
See no issues in logic. Few code suggestion were mentioned in comments, though they might be out of scope

OutOfSchool/OutOfSchool.WebApi/Services/RatingService.cs
alrom: I added it to the base repository class. I'll change other places, where the same code exists, in the nearest future.

OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs
alrom: Done

DmyMi:
@VadymLevkovskyi I like the `IDto` interface idea as it will do magic to simplify our code in some places. But I have a question for future:
Currently we do not really follow REST convention and do not have `id`s in request path for most updates as we put it in `dto` body.
But what if we finally decide to refactor to be more RESTy, is there maybe some recommendation we can do to both use the interface magic and follow convention?

VadymLevkovskyi:
> @VadymLevkovskyi I like the `IDto` interface idea as it will do magic to simplify our code in some places. But I have a question for future: Currently we do not really follow REST convention and do not have `id`s in request path for most updates as we put it in `dto` body. But what if we finally decide to refactor to be more RESTy, is there maybe some recommendation we can do to both use the interface magic and follow convention?

@DmyMi - it's a good point I hadn't thought before. Let's investigate.
You'd like to have "id-less" dto's passed from front-end to presentation layer and (as I get it) there are "id-full" dto's taken from persistence layer (I'm talking about EF side, not ES, mainly because ES part is not analized deep enough). 

So main question (as for me) - do we really need that "id-less" dto's?

If **yes** - we need 2 sets of dto's (with/without id field), 
if **not** - have dto with id field that is sometimes ignored. 

So _naive_ solution for **1st variant**: 
`NotificationUIDto : IDtoNoId<Notification>` and `NotificationDbDto : IDto<Notification>` classes with duplicated fields.
**Pro** - DB and UI dto's could evolve independently, **contra** - lot of fields duplication.

_Mechanical "improvement"_ of **1st variant**:
`NotificationUIDto : IDtoNoId<Notification>` and `NotificationDbDto : NotificationUIDto, IDto<Notification>` classes where duplication is solved thanks to **inheritance**. As for me, this approach introduces **too high coupling** between UI and DB dto's, so better to avoid it.

Solution for **2nd variant**: 
single `NotificationDto : IDto<Notification>` where `.Id` field could be taken into account or not. **Same issue** with DB and UI dto's independency and **additional** one - you would never trust you `.Id` field - as you don't know whether it was populated or not.

_Slightly improved_ solution for **2nd variant** - rehydrate UI `dto.Id` from query string using **custom model binder** (one generic implementation could be enough). Have to check, but expect it to be simple enough - first use default model binder, and then add id from ModelBindingContext data. Registering such binders (all generics for concrete models) might be a tricky point, though reflection/code generation could help. **Contra** - it requires writing extra code. **Pro** - you're assured that `.Id` field is always correct. 
(Moreover, I've even saw in internet approach where `svc.GetById()` - that we have as initial code in some actions - was performed directly in model binder. I have mixed feelings about this way - as it moves logic on UI level, but in kinda "elegant" way, "pretty ugly", I'd say )) )

These're possible solutions I can see, and, pay attention, that "magical Dto" interfaces (mentioned initially) aren't really the obstacle.

--- 

And here are (finally) my recommendations:
- I'd prefer to have **duplicated dto's** for UI and DB, even if they looks similar. DB is not UI, right?
- I'd prefer to have `.Id` **encapsulated into dto** via automatic handling of `.Id` on UI level - **custom model binders**, mainly because they introduce other neat possibilities such as "full-fledged" (cross-fields - i.e. first name + last name, or complex fields - i.e. email) validation (that could be reused on code level - again emails), and I don't like all that metaprogramming (attribute-powered) stuff (not reusable on C# code level).

Unfortunately both my suggestions requires extra coding, so please "choose your poison" )

--
Feel free to ping me in Teams, if you'd like to discuss it in voice

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/QuartzJobConfiguration.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace OutOfSchool.Services.Models.Configurations;
+
+internal class QuartzJobConfiguration : IEntityTypeConfiguration<QuartzJob>
+{
+    public void Configure(EntityTypeBuilder<QuartzJob> builder)
+    {
+        builder.HasKey(x => x.Id);
+
+        builder.HasIndex(x => x.IsDeleted);
```
DmyMi: do we ready need to track deleted quartz jobs? They they are either useful and exist in db or not usefull and we can safely delete it completely.

DmyMi (approved):

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/QuartzJobConfiguration.cs
alrom: Ok. I'll fix it in the next PR.

VadymLevkovskyi (approved):

PR #1248, opened 2023-09-26
Fix mapping for WorkshopProviderViewCard

alrom:


VadymLevkovskyi (approved):
Change seems to be innocent.
Could it be unit-tested somewhere, to make it hard to be changed back? What is our current policy re mapping tests? 
(I'm not asking about testing of automappers mapping profile, I'm more into testing of some module where address changes are asserted and dto/model mapping is performed on some lower levels, unfortunately can't suggest the right class at the moment)

alrom:
> Change seems to be innocent. Could it be unit-tested somewhere, to make it hard to be changed back? What is our current policy re mapping tests? (I'm not asking about testing of automappers mapping profile, I'm more into testing of some module where address changes are asserted and dto/model mapping is performed on some lower levels, unfortunately can't suggest the right class at the moment)

We don't have tests for mapping policy. Only general test for mapper configuration.

VadymLevkovskyi:
> > Change seems to be innocent. Could it be unit-tested somewhere, to make it hard to be changed back? What is our current policy re mapping tests? (I'm not asking about testing of automappers mapping profile, I'm more into testing of some module where address changes are asserted and dto/model mapping is performed on some lower levels, unfortunately can't suggest the right class at the moment)
> 
> We don't have tests for mapping policy. Only general test for mapper configuration.

Thanks, now I know that you have this policy. Though I was asking another thing. Can't agree with your answer, but probably it's lack of capacity.

PR #1256, opened 2023-10-03
Kymachynskyi / Add history of changes for provider admin

DKymachynskyi:


OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminService.cs
```diff
@@ -50,6 +51,7 @@ public class ProviderAdminService : IProviderAdminService
         this.gPRCConfig = gRPCConfig?.Value ?? throw new ArgumentNullException(nameof(gRPCConfig));
         this.externalUrisConfig =
             externalUrisConfig?.Value ?? throw new ArgumentNullException(nameof(externalUrisConfig));
+        this.changesLogConfig = changesLogConfig?.Value ?? throw new ArgumentNullException(nameof(changesLogConfig));
     }
 
     public async Task<ResponseDto> CreateProviderAdminAsync(
```
VadymLevkovskyi, line 62: (Code style, but I understand that it's not yours, so not insisting on fix) 
`CreateProviderAdminAsync` method is "smelly code". It spans over lines 57-210 and as for me is too long. Also it contains nested lambda, where main logic resides, Also there is repeated pattern for setting `ResponseDto` fields (that might be separated as method of that class). Probably there are more changes to approve readability.
Please consider it as an example of not good code, and (if you brave enaugh) rewrite it, or (at least) avoid adding more code in such methods. 
For now, I'm not insisting on changes, because it is out of this PR scope

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminService.cs
```diff
@@ -237,12 +262,16 @@ async Task<ResponseDto> UpdateProviderAdminOperation(UpdateProviderAdminDto upda
             await using var transaction = await context.Database.BeginTransactionAsync().ConfigureAwait(false);
```
VadymLevkovskyi, line 267: (Same code style issue, and I understand that it's not yours, so not insisting on fix) 
`UpdateProviderAdminOperation` takes lines 260-387 and as for me it's just too big. At least it has same `ResponseDto()` fileds setting recurrent pattern.
Again, I'm not insisting on changes right now, but please, don't make such code yourself

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminService.cs
```diff
@@ -277,14 +306,56 @@ async Task<ResponseDto> UpdateProviderAdminOperation(UpdateProviderAdminDto upda
...
+                        addedId.ToString())
+                        .ConfigureAwait(false);
+                }
+
+                foreach (var removedId in removedWorkshopsIds)
+                {
+                    await providerAdminChangesLogService.SaveChangesLogAsync(
+                        providerAdmin,
+                        userId,
+                        OperationType.Update,
```
VadymLevkovskyi, line 342: (Logic) Should it be `OperationType.Update` ?
As for now I can't distinguish if some workshopId was added or removed.
Probably it could be coded not via `OperationType` but by id itself? (I.e. adding minus sign before removed ids - `"-" + removedId.ToString()`?)

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminService.cs
```diff
@@ -612,6 +709,32 @@ await providerAdminChangesLogService.SaveChangesLogAsync(providerAdmin, userId,
         return result;
     }
 
+    private List<(string Name, string Value)> GetTrackedUserProperties(User user)
+    {
+        if (user != null && changesLogConfig.TrackedProperties.TryGetValue("ProviderAdmin", out var trackedProperties))
```
VadymLevkovskyi: 1. (Code style) You could invert your `if` condition and `return` empty list immediately when `user is null` or there is no `"ProviderAdmin"` key or `trackedProperties` is empty. 
For such code construction (`if - return`) there is no need to write `else` keyword explicitely, thus you get less code nesting in your method body and introduce so called "fail fast" approach.

2. (Logic) As I can see `changesLogConfig.TrackedProperties["ProviderAdmin"]` is read each time when `GetTrackedUserProperties` is called, and it might me redundant or even dangerous for multiple calls (i.e. in `UpdateProviderAdminOperation`). For that case you either do same work few times or get different set of properties to be tracked. 
Ofc, you might state that `changesLogConfig.TrackedProperties["ProviderAdmin"]` is never changed for class `ProviderServiceAdmin` lifetime - then why reread it?
Suggestion: read `"ProviderAdmin"` from `changesLogConfig.TrackedProperties` in class ctor.

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminService.cs
```diff
@@ -612,6 +709,32 @@ await providerAdminChangesLogService.SaveChangesLogAsync(providerAdmin, userId,
         return result;
     }
 
+    private List<(string Name, string Value)> GetTrackedUserProperties(User user)
```
VadymLevkovskyi: Why return type `List<(string Name, string Value)>` is used instead of `Dictionary<string, string>`? 
As `Name` is a `User` property name, it is unique and you'll never get key collisions. 
Dictionary is easier for data retrieving (instead of `.FirstOrDefault(x => x.Name == propertyName).Value` simple `TryGetValue` could be used).

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminService.cs
VadymLevkovskyi: 3. (Code style) Additional work to eliminate nested code could be done for nested `if`s inside `for`. It's up to you to decide if it is worth changing the code.

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminService.cs
DKymachynskyi: @VadymLevkovskyi my logic is next. We have properties OperationType, PropertyName, OldValue, NewValue. 
If we update ProviderAdmin with by adding only one workshop - OperationType is "Update", PropertyName is "WorkshopId", OldValue is "string.Empy", NewValue is GUID of added workshop.
If we update ProviderAdmin with by removing only one workshop - OperationType is "Update", PropertyName is "WorkshopId", but OldValue is GUID of removed workshop, NewValue is "string.Empty".

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminChangesLogService.cs
```diff
@@ -11,50 +11,31 @@ public ProviderAdminChangesLogService(IEntityRepository<long, ProviderAdminChang
         this.providerAdminChangesLogRepository = providerAdminChangesLogRepository;
```
VadymLevkovskyi, line 11: (Code style, probably tech debt)
`providerAdminChangesLogRepository` is of type `IEntityRepository<...>` that let you change/delete logs, not only add or read them. Is it intended? Shouldn't it be something like `IEntityAddOnlyRepository<>` ?

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminService.cs
VadymLevkovskyi: I see, thank you. Then there is other question. 
Few lines below you are logging changes from `newPropertiesValues`. 
Let's imagine that `user.MiddleName` was changed from some value to empty/null string. Am I right that such change will never come to the log?

DKymachynskyi: You're right, it's because of the null check in the GetTrackedUserProperties method for propertyValue - changes with null value never come to log.

DKymachynskyi: I fixed it. Thank you!

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminService.cs
DKymachynskyi: I changed the List to the Dictionary. Thank you!

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminService.cs
DKymachynskyi: Fixed all points. Thank you!

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminChangesLogService.cs
DKymachynskyi: @VadymLevkovskyi I added `IEntityAddOnlyRepository<>` and moved the add/read methods from `IEntityRepositoryBase` into it.

VadymLevkovskyi: Thank you. 
Though `RunInTransaction()` is still vulnerable, I can find reasons to put it in either interface, so let it be in new one.

VadymLevkovskyi (approved):
Approved. There are still few "smelly" methods, but it's not part of this PR, I guess

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminChangesLogService.cs
DKymachynskyi: Thank you!

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminService.cs
DmyMi: Hello Vadym, extracted the comments to separate [issue](https://github.com/ita-social-projects/OoS-Backend/issues/1258) so we can deal with it in new PR.

VadymLevkovskyi: Thanks @DmyMi!

PR #1267, opened 2023-10-17
Separation Get Methods

ZdorovenkoKyrylo:


VadymLevkovskyi (commented):
Honestly I don't get the purpose of this change, could you explain why existing Direction/Provider/Institution/MinistryAdmin controllers are wrong?

ZdorovenkoKyrylo:
> Honestly I don't get the purpose of this change, could you explain why existing Direction/Provider/Institution/MinistryAdmin controllers are wrong?

@VadymLevkovskyi As I understand the purpose in separation methods which can use only ```Admins``` from other methods which are created for average users using. Now I just copied some methods for separation, then it will be necessary to delete these methods from other controllers. The initial goal is to decrease the complexity of present Services and separate some checks.

VadymLevkovskyi:
> @VadymLevkovskyi As I understand the purpose in separation methods which can use only `Admins` from other methods which are created for average users using. Now I just copied some methods for separation, then it will be necessary to delete these methods from other controllers. The initial goal is to decrease the complexity of present Services and separate some checks.

@ZdorovenkoKyrylo thank you for the explanation. 
Though I can't buy this decision (as it makes code maintenance, reuse worser, you are already able see it in amount of dependencies injected into constructor), I will not ask for any changes, and will approve the PR when it comes from draft mode.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
```diff
@@ -14,13 +20,48 @@ namespace OutOfSchool.WebApi.Controllers.V1;
...
+
+    private readonly ILogger<DirectionService> loggerDirection;
+    private readonly ILogger<MinistryAdminService> loggerMinistryAdmin;
+    private readonly ILogger<ProviderService> loggerProvider;
+    private readonly ILogger<InstitutionService> loggerInstitution;
+
     private readonly IMapper mapper;
+    private readonly ICurrentUserService currentUserService;
+    private readonly IMinistryAdminService ministryAdminService;
+    private readonly ICacheService cache;
```
alrom: You don't use it. (ICacheService cache) Look at code smells.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -43,30 +43,30 @@ public class ApplicationController : ControllerBase
...
-        var applications = await applicationService.GetAll(filter).ConfigureAwait(false);
-
-        if (!applications.Entities.Any())
-        {
-            return NoContent();
-        }
-
-        return Ok(applications);
-    }
+    // /// <summary>
```
VadymLevkovskyi: (Code maintenance) There is lot of commented out code. What are expected next steps with this code? Isn't it enough to just drop it?

ZdorovenkoKyrylo: As I understand all the commented code will be removed in the future. Now it is commented to be possible to return easier previous state before separation. I can move it out of course, if it is a big problem.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
```diff
@@ -9,36 +12,181 @@ namespace OutOfSchool.WebApi.Controllers.V1;
...
+    private readonly ISensitiveDirectionService directionService;
+    private readonly ISensitiveProviderService providerService;
+    private readonly ISensitiveApplicationService applicationService;
 
-    public AdminController(IUserService userService, IMapper mapper)
+    public AdminController(
+        ILogger<DirectionService> loggerDirection,
+        ILogger<MinistryAdminService> loggerMinistryAdmin,
+        ILogger<ProviderService> loggerProvider,
```
VadymLevkovskyi: Only one `ILogger<AdminController> logger` is necessary. Please use it instead of separate "service" loggers.

ZdorovenkoKyrylo:
> > @VadymLevkovskyi As I understand the purpose in separation methods which can use only `Admins` from other methods which are created for average users using. Now I just copied some methods for separation, then it will be necessary to delete these methods from other controllers. The initial goal is to decrease the complexity of present Services and separate some checks.
> 
> @ZdorovenkoKyrylo thank you for the explanation. Though I can't buy this decision (as it makes code maintenance, reuse worser, you are already able see it in amount of dependencies injected into constructor), I will not ask for any changes, and will approve the PR when it comes from draft mode.

>@VadymLevkovskyi can you check please if everything looks properly? Thanks in advance.

VadymLevkovskyi:
> > @VadymLevkovskyi can you check please if everything looks properly? Thanks in advance.

@ZdorovenkoKyrylo I wish I could but I can't )) and there2 reasons:
- [69.1% Coverage](https://sonarcloud.io/component_measures?id=ita-social-projects-oos-backend&pullRequest=1267&metric=new_coverage&view=list) (36.7% Estimated after merge)
- "This branch has conflicts that must be resolved"

ZdorovenkoKyrylo:
> > > @VadymLevkovskyi can you check please if everything looks properly? Thanks in advance.
> 
> @ZdorovenkoKyrylo I wish I could but I can't )) and there2 reasons:
> 
> * [69.1% Coverage](https://sonarcloud.io/component_measures?id=ita-social-projects-oos-backend&pullRequest=1267&metric=new_coverage&view=list) (36.7% Estimated after merge)
> * "This branch has conflicts that must be resolved"

@VadymLevkovskyi I wish I fixed everything)

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
```diff
@@ -9,36 +12,248 @@ namespace OutOfSchool.WebApi.Controllers.V1;
...
+    /// <response code="401">If the user is not authorized.</response>
+    /// <response code="403">If the user has no rights to use this method.</response>
+    /// <response code="500">If any server error occures.</response>
+    [HttpDelete("{id}")]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
     [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    //TODO: Need to think about implementation. Temporary it is hidden.
```
VadymLevkovskyi: (TODO is confusing) I don't see that this method is hidden. Could you rewrite TODO comment or drop it?

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
```diff
@@ -9,36 +12,248 @@ namespace OutOfSchool.WebApi.Controllers.V1;
...
+     /// <param name="filter">Filter to get a part of all providers that were found.</param>
+     /// <returns>The result is a <see cref="SearchResult{ProviderDto}"/> that contains the count of all found providers and a list of providers that were received.</returns>
+    [HasPermission(Permissions.ProviderRead)]
+    [HttpGet]
+    [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(SearchResult<ProviderDto>))]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    public async Task<IActionResult> GetByFilterProvider([FromQuery] ProviderFilter filter)
```
VadymLevkovskyi: (Logic) Other methods - `GetProviderByFilter` and `GetProvider` do almost same thing as this one but have no check for admin. As initial idea was to extract controller for admins - shouldn't they be dropped?

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
```diff
@@ -9,36 +12,248 @@ namespace OutOfSchool.WebApi.Controllers.V1;
...
+    /// <response code="200">All entities were found.</response>
+    /// <response code="204">No entity was found.</response>
+    /// <response code="500">If any server error occurs.</response>
+    [HasPermission(Permissions.ApplicationRead)]
+    [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(SearchResult<ApplicationDto>))]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     [HttpGet]
-    [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(AdminDto))]
+    public async Task<IActionResult> GetApplications([FromQuery] ApplicationFilter filter)
```
VadymLevkovskyi, line 93: (Logic) This action and subsequent are not checking that user is admin. Is it intended? (I see that ApplicationService has this check for some methods inside, but, as for me we should have consistent checking schema, even if it results in some duplication. Otherwise - there could be risk, that for some method no checks be done.)
Same question about the action having `[AllowAnonymous]` attribute.

And more generic question - shouldn't check for `user.isAdmin()` be put for all action using `OnActionExecuting` override?

OutOfSchool/OutOfSchool.WebApi.Tests/Services/DirectionServiceTests.cs
```diff
@@ -157,67 +162,15 @@ public void GetById_WhenIdIsInvalid_ThrowsArgumentOutOfRangeException(long id)
...
-        };
-        mapper.Setup(m => m.Map<DirectionDto>(It.IsAny<Direction>())).Returns(expected);
-
         // Act
         var countBeforeDeleting = (await service.GetAll().ConfigureAwait(false)).Count();
 
-        context.Entry<Direction>(await repo.GetById(id).ConfigureAwait(false)).State = EntityState.Detached;
-
-        await service.Delete(id).ConfigureAwait(false);
+        await ((ISensitiveDirectionService)service).Delete(id);
```
VadymLevkovskyi, line 168: (Maybe excessive code?) You could change type of `service` field to `DirectionService` and get rid of all these casts.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AdminControllerTests.cs
```diff
@@ -0,0 +1,780 @@
...
+        provider = ProviderDtoGenerator.Generate();
+        provider.UserId = userId;
+        provider.Id = providerId;
+        applications = ApplicationDTOsGenerator.Generate(2).WithWorkshopCard(workshops.First()).WithParent(parent);
+        fakeHttpContext = GetFakeHttpContext();
+        controller.ControllerContext.HttpContext = fakeHttpContext;
+    }
+
+    [Test]
+    public async Task GetApplications_WhenCalledByAdmin_ShouldReturnOkResultObject()
```
VadymLevkovskyi: (Test naming) Test states `...WhenCalledByAdmin...` but not sets this prerequisite. Either wrong name or wrong test (or wrong implementation of `GetApplications` method in controller - see my other comment)

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AdminControllerTests.cs
```diff
@@ -0,0 +1,780 @@
...
+        // Arrange
+        sensitiveDirectionService.Setup(x => x.Delete(id));
+
+        // Act and Assert
+        Assert.ThrowsAsync<ArgumentOutOfRangeException>(
+            async () => await controller.DeleteDirectionById(id).ConfigureAwait(false));
+    }
+
+    [Test]
+    public void GetApplications_WhenCalledParentOrProvider_ShouldThrowUnauthorizedAccess()
```
VadymLevkovskyi: (Code readability) Could you rearrange all tests to have them be grouped by same method tests (now it's bit uncomfortable to analyze what tests were written for `GetApplication`, `Delete`, etc. when they are unordered)

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AdminControllerTests.cs
```diff
@@ -0,0 +1,780 @@
...
+        // Act
+        var result = await controller.GetApplications(new ApplicationFilter()).ConfigureAwait(false) as NoContentResult;
+
+        // Assert
+        result.Should().NotBeNull();
+        result.StatusCode.Should().Be(StatusCodes.Status204NoContent);
+    }
+
+    [Test]
+    public async Task GetMinistryAdmins_WhenCalled_ReturnsOkResultObject_WithExpectedCollectionDtos()
```
VadymLevkovskyi: (Test naming) Test name starts with `GetMinistryAdmins_` while method it tests is `GetByFilterMinistryAdmin()`. Please avoid such ambiguity in this and subsequent tests.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AdminControllerTests.cs
```diff
@@ -0,0 +1,780 @@
...
+                    WorkshopId = new Guid("5e519d63-0cdd-48a8-81da-6365aa5ad8c3"),
+                },
+            },
+        };
+    }
+    private List<WorkshopV2Dto> FakeWorkshops()
+    {
+        return new List<WorkshopV2Dto>()
+        {
+            new WorkshopV2Dto()
```
VadymLevkovskyi: (Test helpers and code reuse) Please use existing (if any) or write new faker to generate `WorkshopV2Dto`s.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationSensitiveServiceTests.cs
```diff
@@ -0,0 +1,359 @@
...
+            areaAdminServiceMock.Object,
+            codeficatorServiceMock.Object,
+            rendererMock.Object,
+            emailSenderMock.Object,
+            localizerMock.Object,
+            hostsConfigMock.Object);
+    }
+
+    [Test]
+    public async Task GetApplications_WhenCalled_ShouldReturnApplications()
```
VadymLevkovskyi: (Test naming) `GetApplications_...` tests `GetAll()` method. Please fix it

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationSensitiveServiceTests.cs
```diff
@@ -0,0 +1,359 @@
...
+
+        // Act
+        var result = await service.GetAll(new ApplicationFilter());
+
+        // Assert
+        Assert.That(result.Entities.Count, Is.EqualTo(1));
+        Assert.That(result.Entities.FirstOrDefault().Workshop.InstitutionId, Is.EqualTo(institutionId));
+    }
+
+    private List<Application> WithApplicationsList()
```
VadymLevkovskyi: (Tests readability, fragility) This approach - to provide test data having specific entities for every test is suboptimal:
1. It's unclear which constants could be changed (b/c we don't check them in tests) and which have dependencies in test code.
2. (Similar to previous) It makes tests obscure (i.e. there is line `Assert.That(result.Entities.Count, Is.EqualTo(1));` where 1 is a magic number and noone knows why it is there)
3. It makes hard to add new tests, because while adding more data entities or changing existing you could break other tests.
4. It doesn't use all power of fake test data generators.

Suggested change
You could still have `WithApplicationsList()` method that generates test-neutral data (though it could be easily replaced by `faker.Generate()`). 
In arrange phase - you take this data for your test, change N (and N could even be random, no magical 1s) entities to match test criteria, perform act phase as usual and assert that expected N is obtained in results.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/DirectionSensitiveServiceTests.cs
```diff
@@ -0,0 +1,171 @@
...
+    private void SeedDatabase()
+    {
+        using var ctx = new OutOfSchoolDbContext(options);
+        {
+            ctx.Database.EnsureDeleted();
+            ctx.Database.EnsureCreated();
+
+            ctx.Institutions.Add(new Institution()
+            {
+                Id = new Guid("af475193-6a1e-4a75-9ba3-439c4300f771"),
```
VadymLevkovskyi: (Readablity) Please make `"af475193-6a1e-4a75-9ba3-439c4300f771"` guid a local variable and use it for `InstitutionId`, as it's not fun to check if all symbols are same everywhere
Also please prepare different local variable with guid for `InstitutionHierarchyId` (At least I believe they should be different for real db).

VadymLevkovskyi (commented):
Please check my other comments. 
In short I see no big problems, but
- user status checking is not uniform;
- there is some duplicated code in controller;
- tests are bit not readable.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
VadymLevkovskyi: Resolving this one, as code seems to be fixed

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
ZdorovenkoKyrylo: I tried to put into practice your comments. I think now this is more or less done. Can you check please?)

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
VadymLevkovskyi: You've set this comment resolved, but performed no changes. Please clarify reasons to resolve it.

Technical note - as it's hard for me to distinguish what was really done and what wasn't - please don't resolve my comments. I'll do that, when I agree with your changes/arguments given in discussion. If you need some way to mark processed things, then the best way is to response in text to my comment - even simple "fixed" is enough - or to react using emoji functionality. 
In short - please don't resolve comments created by other person.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
```diff
@@ -1,44 +1,274 @@
...
+       ISensitiveDirectionService directionService,
+       ISensitiveProviderService providerService,
+       IStringLocalizer<SharedResource> localizer)
+    {
+        this.localizer = localizer;
+        this.logger = logger;
+        this.applicationService = applicationService;
+        this.directionService = directionService;
+        this.providerService = providerService;
+        this.currentUserService = currentUserService ?? throw new ArgumentNullException(nameof(currentUserService));
```
VadymLevkovskyi: (Redundant code) This field `currentUserService` - seems to be never used. Please use or remove

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
```diff
@@ -1,44 +1,274 @@
-using AutoMapper;
+using Google.Apis.Discovery;
```
VadymLevkovskyi: (Redundant code) `using Google.Apis.Discovery;` - do we really need it here? Please remove this and other redundant `using`s

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
```diff
@@ -1,44 +1,274 @@
...
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    public async Task<IActionResult> GetByFilterMinistryAdmin([FromQuery] MinistryAdminFilter filter)
+     {
+         if (!User.IsInRole(nameof(Role.TechAdmin).ToLower()))
+         {
+             logger.LogError("You have no rights because you are not an admin");
+             return StatusCode(403, "Forbidden to update another user if you don't have TechAdmin role.");
+         }
```
VadymLevkovskyi: (Code duplication) Please consider better way to enforce user rights checking. 
Pay attention, that your copy-paste approach is really bad. I.e., check the next:
```
    public async Task<ActionResult> DeleteDirectionById(long id)
    {
        if (!User.IsInRole(nameof(Role.TechAdmin).ToLower()))
        {
            ...
            return StatusCode(403, "Forbidden to update another user if you don't have TechAdmin role.");
        }
```
Method is `DeleteDirectionById` - log text says `Forbidden to update another user`. 
I believe you could do it better.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
```diff
@@ -1,44 +1,274 @@
...
+    /// </summary>
+    /// <param name="filter">Application filter.</param>
+    /// <returns>List of all applications.</returns>
+    /// <response code="200">All entities were found.</response>
+    /// <response code="204">No entity was found.</response>
+    /// <response code="500">If any server error occurs.</response>
+    [HasPermission(Permissions.ApplicationRead)]
+    [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(SearchResult<ApplicationDto>))]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
```
VadymLevkovskyi: (Irrelevant comments and API doc) Method could return `403`, but it is not described here.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
```diff
@@ -1,44 +1,274 @@
...
+
+    /// <summary>
+    /// Block/unblock Provider.
+    /// </summary>
+    /// <param name="providerBlockDto">Entity to update.</param>
+    /// <returns>Block Provider.</returns>
+    [HasPermission(Permissions.ProviderBlock)]
+    [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(ProviderBlockDto))]
     [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
```
VadymLevkovskyi: (Irrelevant API doc) Method could return `403`, but here `401` is stated.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
VadymLevkovskyi: Thank you for your efforts, I've checked it. Please check my other comments.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/DirectionServiceTests.cs
```diff
@@ -1,17 +1,22 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Linq.Expressions;
```
VadymLevkovskyi: (redundant code) Please check and remove unnecessary usings (I see 5 of them at the moment)

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AdminControllerTests.cs
```diff
@@ -0,0 +1,956 @@
...
+using System.Linq;
+using System.Net;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using AutoMapper;
+using FluentAssertions;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Filters;
```
VadymLevkovskyi: (Redundant usings) This one seems to be redundant

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AdminControllerTests.cs
```diff
@@ -0,0 +1,956 @@
...
+    private Mock<ISensitiveApplicationService> sensitiveApplicationService;
+    private Mock<ILogger<AdminController>> logger;
+    private Mock<IStringLocalizer<SharedResource>> localizer;
+
+    private string userId;
+    private Guid providerId;
+    private Mock<HttpContext> httpContext;
+
+    private List<ApplicationDto> applications;
+    private IEnumerable<ChildDto> children;
```
VadymLevkovskyi: (Redundant code) Fields `children` and `userRole` aren't used

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AdminControllerTests.cs
```diff
@@ -0,0 +1,956 @@
...
+    [Test]
+    public async Task GetByFilterMinistryAdmin_WhenCalled_ReturnsOkResultObject_WithExpectedCollectionDtos()
+    {
+        // Arrange
+        var user = new ClaimsPrincipal(new ClaimsIdentity(
+            new Claim[]
+            {
+                new Claim(ClaimTypes.Role, nameof(Role.TechAdmin).ToLower()),
+            }));
+        controller.ControllerContext.HttpContext.User = user;
```
VadymLevkovskyi: (Code duplication) These lines are repeating (with variations of user role) everywhere. Could small helper method be used?

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AdminControllerTests.cs
```diff
@@ -0,0 +1,956 @@
...
+        var result = await controller.DeleteDirectionById(id) as NoContentResult;
+
+        // Assert
+        Assert.That(result, Is.Not.Null);
+        Assert.AreEqual(204, result.StatusCode);
+    }
+
+    [Test]
+    [TestCase(0)]
+    public void DeleteDirectionById_WhenIdIsInvalid_ReturnsBadRequestObjectResult(long id)
```
VadymLevkovskyi: (Naming issue) Test name promises that test `_ReturnsBadRequestObjectResult`, but test code is asserting thrown exception. Please fix

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AdminControllerTests.cs
VadymLevkovskyi: You've changed nothing here and not responded. But set the comment to resolved. I'm sad

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationSensitiveServiceTests.cs
```diff
@@ -0,0 +1,360 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Threading.Tasks;
+using AutoMapper;
+using FluentAssertions;
+using Microsoft.Extensions.Localization;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using MockQueryable.Moq;
```
VadymLevkovskyi: (Redundant using) There are 2 excessive `using`s here

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationSensitiveServiceTests.cs
```diff
@@ -0,0 +1,360 @@
...
+using OutOfSchool.WebApi.Models.Workshops;
+using OutOfSchool.WebApi.Services;
+namespace OutOfSchool.WebApi.Tests.Services;
+
+[TestFixture]
+public class ApplicationSensitiveServiceTests
+{
+    private const int N = 1;
+    private ISensitiveApplicationService service;
+    private ISensitiveApplicationService sensitiveApplicationService;
```
VadymLevkovskyi: (Redundant code) `sensitiveApplicationService` field is not used

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationSensitiveServiceTests.cs
VadymLevkovskyi: Resolved but nothing done.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/DirectionSensitiveServiceTests.cs
VadymLevkovskyi: Resolved but nothing done.

VadymLevkovskyi (changes requested):
Few more changes need to be implemented. 
Few previous comments weren't addressed, just set as resolved with no explanation - not a best way to speed up things

ZdorovenkoKyrylo:
> Few more changes need to be implemented. Few previous comments weren't addressed, just set as resolved with no explanation - not a best way to speed up things

Thank you for the comments. I`ve understood about mistake with resolving comments. As far as I remember there was 2 big problems for me we discussed which during daily. 
The first one was with TODO comments for the method which is not used, but actually for this method tests were written so we thought how to hide it and decided to make it using FeatureGate.
The second was with ```OnActionExecuting``` and I tried to realise similar to what we had already in our project.

Sorry for the situation I think a part of comments can be unsolved because there are so many merge conflicts to resolve with this branch. And things are happen. I`ll try to fix everyting as fast as possible and request a review. Thank you for all the comments.)

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
ZdorovenkoKyrylo: Done

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
ZdorovenkoKyrylo: > (Redundant code) `using Google.Apis.Discovery;` - do we really need it here? Please remove this and other redundant `using`s

It is not used. Done

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationSensitiveServiceTests.cs
ZdorovenkoKyrylo: > (Redundant using) There are 2 excessive `using`s here

Done

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationSensitiveServiceTests.cs
ZdorovenkoKyrylo: > (Redundant code) `sensitiveApplicationService` field is not used

Done

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
ZdorovenkoKyrylo: Thank you for the comment. We discussed this issue and found that these 2 methods really do practically the same logic, so we had a talk with front-end team and asked about their using these methods. They didn`t use ```GetProvider``` so it was removed

VadymLevkovskyi: Ok, I'll be more specific: there are `GetProviderByFilter` and `GetByFilterProvider`. Please check them as well

OutOfSchool/OutOfSchool.WebApi.Tests/Services/DirectionSensitiveServiceTests.cs
VadymLevkovskyi: Sorry Kyrylo, it is my fault. I was wrong (was looking at the `ApplicationSensitiveServiceTests`, but commented here). 
The initial comment is resolved. 
Sorry again.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationSensitiveServiceTests.cs
ZdorovenkoKyrylo: > Resolved but nothing done.

@VadymLevkovskyi Can you please a little bit deeper about this comment. I do not full understand how it should be look like. I tried to introduce a field which was like a constant parament ```N``` and inside this test I replaced hardcode number with these value of ```N```. But as I understand now it should be done in another way. And I do not understand what should be done with this paramert. Is it should be used like an attribute. Because when I mock smth I do not quite understand how to return exactly N element and how it should be written there.
Thank you in advance!)

VadymLevkovskyi: Little tiny bit deeper ;) 

Few comments about code below:
1. Added test for not admin case 
2. All data are taken from generators
3. Wrong data are explicitely constructed in related test, no magic numbers used, no generic "wrong" data created for all tests - thus tests are more independent.
4. As many as possible replacements of `It.IsAny` is done - thus we see what ids are expected in what methods
5. `Assert.AreEqual(result.Entities.Count, application.Count);` is dropped everywhere, because it doesn't check the SUT implementation, instead it tests part of the same test code (specifically `mapper.Setup`), so there is no reason to reverify it. Rephrasing, that assert is schematically same to 
```
// somewhere in test code
var isTrue = true;
...
Assert.That(isTrue. Is.True());
```
6. Pay attention to the real assertion - powered by `.It.Is<Expression<...>>(...)` and related `.VerifyAll()`

Sorry for lot of text. Hope it will be helpful.

```
using AutoMapper;
using Microsoft.Extensions.Logging;
using Microsoft.Extensions.Options;
using Moq;
using NUnit.Framework;
using OutOfSchool.EmailSender;
using OutOfSchool.RazorTemplatesData.Services;
using OutOfSchool.Services.Models;
using OutOfSchool.Services.Repository;
using OutOfSchool.Tests.Common;
using OutOfSchool.Tests.Common.TestDataGenerators;
using OutOfSchool.WebApi.Config;
using OutOfSchool.WebApi.Models.Application;
using OutOfSchool.WebApi.Services;
using System;
using System.Collections.Generic;
using System.Linq;
using System.Linq.Expressions;
using System.Threading.Tasks;
namespace OutOfSchool.WebApi.Tests.Services;

[TestFixture]
public class ApplicationSensitiveServiceTests
{
    private ISensitiveApplicationService service;
    private string currentUserId;
    private Mock<IApplicationRepository> applicationRepositoryMock;
    private Mock<IMapper> mapper;
    private Mock<ICurrentUserService> currentUserServiceMock;
    private Mock<IMinistryAdminService> ministryAdminServiceMock;
    private Mock<IRegionAdminService> regionAdminServiceMock;
    private Mock<IAreaAdminService> areaAdminServiceMock;
    private Mock<ICodeficatorService> codeficatorServiceMock;

    [SetUp]
    public void SetUp()
    {
        applicationRepositoryMock = new Mock<IApplicationRepository>();
        mapper = new Mock<IMapper>();
        currentUserServiceMock = new Mock<ICurrentUserService>();
        ministryAdminServiceMock = new Mock<IMinistryAdminService>();
        regionAdminServiceMock = new Mock<IRegionAdminService>();
        areaAdminServiceMock = new Mock<IAreaAdminService>();
        codeficatorServiceMock = new Mock<ICodeficatorService>();

        service = new ApplicationService(
            applicationRepositoryMock.Object,
            new Mock<ILogger<ApplicationService>>().Object,
            new Mock<IWorkshopRepository>().Object,
            mapper.Object,
            new Mock<IOptions<ApplicationsConstraintsConfig>>().Object,
            new Mock<INotificationService>().Object,
            new Mock<IProviderAdminService>().Object,
            new Mock<IChangesLogService>().Object,
            new Mock<IWorkshopServicesCombiner>().Object,
            currentUserServiceMock.Object,
            ministryAdminServiceMock.Object,
            regionAdminServiceMock.Object,
            areaAdminServiceMock.Object,
            codeficatorServiceMock.Object,
            new Mock<IRazorViewToStringRenderer>().Object,
            new Mock<IEmailSender>().Object,
            new Mock<IStringLocalizer<SharedResource>>().Object,
            new Mock<IOptions<HostsConfig>>().Object);

        currentUserId = Guid.NewGuid().ToString();
        currentUserServiceMock.SetupGet(c => c.UserId).Returns(currentUserId);
    }

    [Test]
    public void GetAll_WhenNotAdmin_Throws()
    {
        // Arrange
        SetupAdminRights(isAdmin: false);

        // Act, Assert
        Assert.ThrowsAsync<UnauthorizedAccessException>(async () => await service.GetAll(new ApplicationFilter()));
    }

    [Test]
    public async Task GetAll_WhenMinistryAdmin_ReturnsApplications()
    {
        // Arrange
        var applications = GenerateApplicationsForSameWorkshop();

        SetupAdminRights(isAdmin: true, isMinistry: true);

        var appFilter = new ApplicationFilter();
        SetupRepositoryWithVerification(appFilter, applications);

        ministryAdminServiceMock.Setup(m => m.GetByIdAsync(currentUserId))
            .Returns(Task.FromResult(new MinistryAdminDto() { InstitutionId = applications[0].Workshop.InstitutionHierarchy.InstitutionId }));

        // Act
        var result = await service.GetAll(appFilter);

        // Assert
        mapper.VerifyAll();
        currentUserServiceMock.VerifyAll();
        applicationRepositoryMock.VerifyAll();
    }

    [Test]
    public async Task GetAll_WhenMinistryAdmin_ReturnsRelatedApplications()
    {
        // Arrange
        var applications = GenerateApplicationsForSameWorkshop();
        var institutionId = applications[0].Workshop.InstitutionHierarchy.InstitutionId;

        var mnistryAdmin = AdminGenerator.GenerateMinistryAdminDto();
        mnistryAdmin.InstitutionId = institutionId;

        var otherInstitutionHierarchy = InstitutionHierarchyGenerator.Generate();
        var otherWorkshop = WorkshopGenerator.Generate().WithInstitutionHierarchy(otherInstitutionHierarchy);

        var wrongApplications = applications.Take(1).ToList().WithWorkshop(otherWorkshop);
        var correctApplications = applications.Skip(wrongApplications.Count).ToList();

        SetupAdminRights(isAdmin: true, isMinistry: true);

        ministryAdminServiceMock.Setup(m => m.GetByIdAsync(currentUserId)).Returns(Task.FromResult(mnistryAdmin));

        var appFilter = new ApplicationFilter();
        SetupRepositoryWithVerification(appFilter, correctApplications, wrongApplications);

        // Act
        var result = await service.GetAll(appFilter);

        // Assert
        mapper.VerifyAll();
        currentUserServiceMock.VerifyAll();
        ministryAdminServiceMock.VerifyAll();
        applicationRepositoryMock.VerifyAll();
    }

    [Test]
    public async Task GetAll_WhenRegionAdmin_ReturnsRelatedApplications()
    {
        // Arrange
        var applications = GenerateApplicationsForSameWorkshop();
        var workshop = applications[0].Workshop;
        var institutionId = workshop.InstitutionHierarchy.InstitutionId;
        var catottgId = workshop.Provider.LegalAddress.CATOTTGId;

        var regionAdmin = AdminGenerator.GenerateRegionAdminDto();
        regionAdmin.InstitutionId = institutionId;

        var childrenCatottgIds = new List<long> { catottgId } as IEnumerable<long>;

        var otherInstitutionHierarchyWorkshop = WorkshopGenerator.Generate().WithInstitutionHierarchy(InstitutionHierarchyGenerator.Generate()).WithProvider(workshop.Provider);
        var otherProviderWorkshop = WorkshopGenerator.Generate().WithInstitutionHierarchy(workshop.InstitutionHierarchy).WithProvider(ProvidersGenerator.Generate());

        var wrongApplications = new List<Application> 
        {
            applications[0].WithWorkshop(otherInstitutionHierarchyWorkshop),
            applications[1].WithWorkshop(otherProviderWorkshop),
        };
        var correctApplications = applications.Skip(wrongApplications.Count).ToList();

        SetupAdminRights(isAdmin: true, isRegion: true);

        var appFilter = new ApplicationFilter();
        SetupRepositoryWithVerification(appFilter, correctApplications, wrongApplications);

        regionAdminServiceMock.Setup(m => m.GetByUserId(currentUserId)).Returns(Task.FromResult(regionAdmin));

        codeficatorServiceMock
            .Setup(x => x.GetAllChildrenIdsByParentIdAsync(regionAdmin.CATOTTGId))
            .Returns(Task.FromResult(childrenCatottgIds));

        // Act
        var result = await service.GetAll(appFilter);

        // Assert
        mapper.VerifyAll();
        currentUserServiceMock.VerifyAll();
        regionAdminServiceMock.VerifyAll();
        codeficatorServiceMock.VerifyAll();
        applicationRepositoryMock.VerifyAll();
    }

    [Test]
    public async Task GetAll_WhenAreaAdmin_ReturnsRelatedApplications()
    {
        // Arrange
        var applications = GenerateApplicationsForSameWorkshop();
        var workshop = applications[0].Workshop;
        var institutionId = workshop.InstitutionHierarchy.InstitutionId;
        var catottgId = workshop.Provider.LegalAddress.CATOTTGId;

        var areaAdmin = AdminGenerator.GenerateAreaAdminDto();
        areaAdmin.InstitutionId = institutionId;

        var childrenCatottgIds = new List<long> { catottgId } as IEnumerable<long>;

        var otherInstitutionHierarchyWorkshop = WorkshopGenerator.Generate().WithInstitutionHierarchy(InstitutionHierarchyGenerator.Generate()).WithProvider(workshop.Provider);
        var otherProviderWorkshop = WorkshopGenerator.Generate().WithInstitutionHierarchy(workshop.InstitutionHierarchy).WithProvider(ProvidersGenerator.Generate());

        var wrongApplications = new List<Application> 
        {
            applications[0].WithWorkshop(otherInstitutionHierarchyWorkshop),
            applications[1].WithWorkshop(otherProviderWorkshop),
        };
        var correctApplications = applications.Skip(wrongApplications.Count).ToList();

        SetupAdminRights(isAdmin: true, isArea: true);

        var appFilter = new ApplicationFilter();
        SetupRepositoryWithVerification(appFilter, correctApplications, wrongApplications);

        areaAdminServiceMock.Setup(m => m.GetByUserId(currentUserId)).Returns(Task.FromResult(areaAdmin));

        codeficatorServiceMock
            .Setup(x => x.GetAllChildrenIdsByParentIdAsync(areaAdmin.CATOTTGId))
            .Returns(Task.FromResult(childrenCatottgIds));

        // Act
        var result = await service.GetAll(appFilter);

        // Assert
        mapper.VerifyAll();
        currentUserServiceMock.VerifyAll();
        areaAdminServiceMock.VerifyAll();
        codeficatorServiceMock.VerifyAll();
        applicationRepositoryMock.VerifyAll();
    }

    private void SetupAdminRights(bool isAdmin, bool isMinistry = false, bool isRegion = false, bool isArea = false)
    {
        currentUserServiceMock.Setup(c => c.IsAdmin()).Returns(isAdmin);
        currentUserServiceMock.Setup(c => c.IsMinistryAdmin()).Returns(isMinistry);
        currentUserServiceMock.Setup(c => c.IsRegionAdmin()).Returns(isRegion);
        currentUserServiceMock.Setup(c => c.IsAreaAdmin()).Returns(isArea);
    }

    private List<Application> GenerateApplicationsForSameWorkshop()
    {
        var institutionHierarchy = InstitutionHierarchyGenerator.Generate();
        var workshop = WorkshopGenerator.Generate().WithInstitutionHierarchy(institutionHierarchy).WithProvider();
        var parent = ParentGenerator.Generate();
        var child = ChildGenerator.Generate().WithParent(parent);

        return ApplicationGenerator.Generate(6).WithWorkshop(workshop).WithParent(parent).WithChild(child);
    }

    private void SetupRepositoryWithVerification(ApplicationFilter appFilter, List<Application> correctApplications, List<Application> wrongApplications = null)
    {
        applicationRepositoryMock.Setup(
            w => w.Get(
                appFilter.From,
                appFilter.Size,
                It.IsAny<string>(),
                It.Is<Expression<Func<Application, bool>>>(
                    // here main magic goes - verification that predicate built in service and passed to applicationRepository should behave correctly on "correct" and "wrong" applications
                    expr => correctApplications.All(expr.Compile())
                        && (wrongApplications == null || !wrongApplications.Any(expr.Compile()))
                ),
                It.IsAny<Dictionary<Expression<Func<Application, object>>, SortDirection>>(),
                It.IsAny<bool>()))
            .Returns(correctApplications.AsTestAsyncEnumerableQuery())
            .Verifiable("Wrong expression was built in service and passed to applicationRepository.Get() whereExpression argument. Probably some conditions aren't tested");

        mapper.Setup(m => m.Map<List<ApplicationDto>>(It.IsAny<List<Application>>()))
            .Returns(correctApplications.Select(app => new ApplicationDto()).ToList());
    }
}
```

Also improved AddressGenerator:
```
    static AddressGenerator()
    {
        faker = new Faker<Address>()
            .RuleFor(x => x.Id, f => f.IndexFaker)
            .RuleFor(x => x.CATOTTG, _ => CATOTTGGenerator.Generate())
            .RuleFor(x => x.CATOTTGId, (f, address) => address.CATOTTG.Id) // <-- this line was changed
            .RuleFor(x => x.Street, f => f.Address.StreetName())
            .RuleFor(x => x.BuildingNumber, f => f.Address.BuildingNumber())
            .RuleFor(x => x.Latitude, f => f.Address.Latitude())
            .RuleFor(x => x.Longitude, f => f.Address.Longitude());

        // Increment initial value of IndexFaker to have first created entity with Id=1
        // and prevent System.InvalidOperationException when it is added to the DbContext
        (faker as IFakerTInternal).FakerHub.IndexFaker++;
    }
```

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
VadymLevkovskyi: @ZdorovenkoKyrylo Is it pushed to server? Can't find the change

ZdorovenkoKyrylo: > @ZdorovenkoKyrylo Is it pushed to server? Can't find the change

I hope it will be pushed today together with implementation your comment about test below. 
Thank you for it)
I just highlight for myself to find comments which I should do. Because there so many of them. I used practice resolving before. Now I follow your tips and just write the comment "Done"

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
ZdorovenkoKyrylo: > Ok, I'll be more specific: there are `GetProviderByFilter` and `GetByFilterProvider`. Please check them as well

I asked frontend team, they use as I understand- ```GetProviderByFilter```, so I remove the second one. Because they do the same job)

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
ZdorovenkoKyrylo: > (Code duplication) Please consider better way to enforce user rights checking. Pay attention, that your copy-paste approach is really bad. I.e., check the next:
> 
> ```
>     public async Task<ActionResult> DeleteDirectionById(long id)
>     {
>         if (!User.IsInRole(nameof(Role.TechAdmin).ToLower()))
>         {
>             ...
>             return StatusCode(403, "Forbidden to update another user if you don't have TechAdmin role.");
>         }
> ```
> 
> Method is `DeleteDirectionById` - log text says `Forbidden to update another user`. I believe you could do it better.

I`ve created a private method ```IsTechAdmin``` to decrease duplication)
I hope you like this approach)

ZdorovenkoKyrylo:
> > Few more changes need to be implemented. Few previous comments weren't addressed, just set as resolved with no explanation - not a best way to speed up things
> 
> Thank you for the comments. I`ve understood about mistake with resolving comments. As far as I remember there was 2 big problems for me we discussed which during daily. The first one was with TODO comments for the method which is not used, but actually for this method tests were written so we thought how to hide it and decided to make it using FeatureGate. The second was with `OnActionExecuting` and I tried to realise similar to what we had already in our project.
> 
> Sorry for the situation I think a part of comments can be unsolved because there are so many merge conflicts to resolve with this branch. And things are happen. I`ll try to fix everyting as fast as possible and request a review. Thank you for all the comments.)

@VadymLevkovskyi I hope I didn`t skip anything and resolve all. PR should be ready for RerereReview))
I wish everything is right)

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
VadymLevkovskyi: Though I'll treat this change from you as approved, please check other options: [action filters](https://learn.microsoft.com/en-us/aspnet/core/mvc/controllers/filters?view=aspnetcore-8.0#action-filters) and [custom authorization policy](https://learn.microsoft.com/en-us/aspnet/core/security/authorization/iard?view=aspnetcore-8.0).

Former is better because it's impossible not execute the `action filter` (in your code you could forget to call `IsTechAdmin()` in some action), and also it is already used in the project (i.e. see the `AreaAdminController`). At same time it is not reusable outside of the controller.

Latter is better, b/c it is reusable outside of one controller, but is worse, b/c you need to write the attribute it in every action (and could forget to do it).

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AdminControllerTests.cs
VadymLevkovskyi: Field `providers` isn't used as well

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AdminControllerTests.cs
```diff
@@ -0,0 +1,643 @@
...
+using FluentAssertions;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Localization;
+using Microsoft.Extensions.Logging;
+using Moq;
+using NUnit.Framework;
+using OutOfSchool.Common;
+using OutOfSchool.Common.Enums;
```
VadymLevkovskyi: (Excessive using) `using OutOfSchool.Common.Enums;` - this one could be deleted

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AdminControllerTests.cs
VadymLevkovskyi: Not addressed in last commit

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AdminControllerTests.cs
VadymLevkovskyi: Not addressed in last commit

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationServiceTests.cs
```diff
@@ -33,6 +33,7 @@ namespace OutOfSchool.WebApi.Tests.Services;
 public class ApplicationServiceTests
 {
     private IApplicationService service;
+    private ISensitiveApplicationService sensitiveApplicationService;
```
VadymLevkovskyi: (Excessive code) Field `sensitiveApplicationService` is never used

OutOfSchool/OutOfSchool.WebApi.Tests/Services/DirectionSensitiveServiceTests.cs
```diff
@@ -0,0 +1,175 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
```
VadymLevkovskyi: (Excessive code) 3 usings are redundant

OutOfSchool/OutOfSchool.WebApi.Tests/Services/DirectionSensitiveServiceTests.cs
```diff
@@ -0,0 +1,175 @@
...
+    private OutOfSchoolDbContext context;
+    private IEntityRepositorySoftDeleted<long, Direction> repo;
+    private IWorkshopRepository repositoryWorkshop;
+    private ISensitiveDirectionService service;
+    private Mock<IStringLocalizer<SharedResource>> localizer;
+    private Mock<ILogger<DirectionService>> logger;
+    private Mock<IMapper> mapper;
+    private Mock<ICurrentUserService> currentUserServiceMock;
+    private Mock<IMinistryAdminService> ministryAdminServiceMock;
+    private Mock<IRegionAdminService> regionAdminServiceMock;
```
VadymLevkovskyi: (Excessive code) From all fields listed here only next (less than a half) are necessary. 
```
    private static Guid institutionId;
    private static Guid institutionHierarchyId;
    private IEntityRepositorySoftDeleted<long, Direction> repo;
    private ISensitiveDirectionService service;
    private Mock<IMapper> mapper;
```

Other could be made local variables/inlined/passed as a parameter.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
```diff
@@ -1,44 +1,247 @@
...
+    /// Get all applications from the database.
+    /// </summary>
+    /// <param name="filter">Application filter.</param>
+    /// <returns>List of all applications.</returns>
+    /// <response code="200">All entities were found.</response>
+    /// <response code="204">No entity was found.</response>
+    /// <response code="500">If any server error occurs.</response>
+    [HasPermission(Permissions.ApplicationRead)]
+    [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(SearchResult<ApplicationDto>))]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
```
VadymLevkovskyi, line 89: (OpenAPI documentation issue) `Status403` is not mentioned here

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AdminController.cs
VadymLevkovskyi: Not addressed in the last commit

VadymLevkovskyi (changes requested):
Some comments reported earlier weren't fixed, some new added. 
Sorry for overcommenting and such long-long PR, hope we need one last commit.

ZdorovenkoKyrylo:
> Some comments reported earlier weren't fixed, some new added. Sorry for overcommenting and such long-long PR, hope we need one last commit.

@VadymLevkovskyi Thank you, I fixed all (except for 2 comments) locally. And I hope we will find some solution here. 
It is about Code duplication in tests where we set permissions for user. And about unnecessary field in ```DirectionSensitiveServiceTests```. I`ll wait about your opinion about that. Thank you!

VadymLevkovskyi:
> > Some comments reported earlier weren't fixed, some new added. Sorry for overcommenting and such long-long PR, hope we need one last commit.
> 
> @VadymLevkovskyi Thank you, I fixed all (except for 2 comments) locally. And I hope we will find some solution here. It is about Code duplication in tests where we set permissions for user. And about unnecessary field in `DirectionSensitiveServiceTests`. I`ll wait about your opinion about that. Thank you!

I don't get what opinion I should provide. Should I write the code?

ZdorovenkoKyrylo:
> > > Some comments reported earlier weren't fixed, some new added. Sorry for overcommenting and such long-long PR, hope we need one last commit.
> > 
> > 
> > @VadymLevkovskyi Thank you, I fixed all (except for 2 comments) locally. And I hope we will find some solution here. It is about Code duplication in tests where we set permissions for user. And about unnecessary field in `DirectionSensitiveServiceTests`. I`ll wait about your opinion about that. Thank you!
> 
> I don't get what opinion I should provide. Should I write the code?
I think no. I have some idea about it. Because it should be smth similar in our code already, and I really didn`t understand what we need this change for, because we have different permissions in our test methods. Now I think it is more clear. So, I wrote it and I hope you will like it then, thank you)

ZdorovenkoKyrylo:
> Some comments reported earlier weren't fixed, some new added. Sorry for overcommenting and such long-long PR, hope we need one last commit.

@VadymLevkovskyi I tried to fix all the comments. Check please when it will be convenient for you. I want to call your attention on file ```DirectionSensitiveServiceTests```. Here I removed fields. But I stayed here all the fields you have mentioned and one more- ```options``` it is because it is used in method ```SeedDatabase``` which is located a little bit lower. So I have a question about this field. Other comments it seems I fixed but I am not so sure because there was lots of conflicts and smth could be skipped))
So, I will be waiting for your review. Thank in advance)

ZdorovenkoKyrylo:
> > Some comments reported earlier weren't fixed, some new added. Sorry for overcommenting and such long-long PR, hope we need one last commit.
> 
> @VadymLevkovskyi I tried to fix all the comments. Check please when it will be convenient for you. I want to call your attention on file `DirectionSensitiveServiceTests`. Here I removed fields. But I stayed here all the fields you have mentioned and one more- `options` it is because it is used in method `SeedDatabase` which is located a little bit lower. So I have a question about this field. Other comments it seems I fixed but I am not so sure because there were lots of conflicts and smth could be skipped)) So, I will be waiting for your review. Thanks in advance)

ZdorovenkoKyrylo:
> > Some comments reported earlier weren't fixed, some new added. Sorry for overcommenting and such long-long PR, hope we need one last commit.
> 
> @VadymLevkovskyi I tried to fix all the comments. Check please when it will be convenient for you. I want to call your attention on file `DirectionSensitiveServiceTests`. Here I removed fields. But I stayed here all the fields you have mentioned and one more- `options` it is because it is used in method `SeedDatabase` which is located a little bit lower. So I have a question about this field. Other comments it seems I fixed but I am not so sure because there was lots of conflicts and smth could be skipped)) So, I will be waiting for your review. Thanks in advance)

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AdminControllerTests.cs
VadymLevkovskyi: Closing this one (mainly because I'm tired and see no good progress), though neither `provider`, nor `userRole` fields (both redundant) are not removed. 
Kyrylo you are very inattentive (

VadymLevkovskyi (approved):

PR #1272, opened 2023-10-19
Semkiv/admin update fix

min4uk:
Fixed update error by changing incoming Dto.
Removed institutionId changing, because it`s not allowed by business rules.
Removed some code smell.


alrom (approved):

VadymLevkovskyi:
You've mentioned "Removed institutionId changing, because it`s not allowed by business rules". Are there any tests enforcing this business rule? If not - could they be written?

min4uk:
> You've mentioned "Removed institutionId changing, because it`s not allowed by business rules". Are there any tests enforcing this business rule? If not - could they be written?

I will think about how to check it correctly and do it

min4uk:
> You've mentioned "Removed institutionId changing, because it`s not allowed by business rules". Are there any tests enforcing this business rule? If not - could they be written?

I tried to, but I couldn't figure out how to write such unit test😥

VadymLevkovskyi:
> > You've mentioned "Removed institutionId changing, because it`s not allowed by business rules". Are there any tests enforcing this business rule? If not - could they be written?
> 
> I tried to, but I couldn't figure out how to write such unit test😥

@min4uk please correct me if I'm wrong. As I can see the logic of Id hiding is hidden in `CreateMap<BaseUserDto, NnnAdminBaseDto>` that is used while creating request via Http using `CommunicationService.SendRequest` method.

As `CommunicationService` is parent class, not a dependency (as it has to be!) and mentioned method cannot be mocked we could go deeper and find that real http request is sent via httpClient that is created by `httpClientFactory` - injected dependency that could be mocked.

So one possible solution - is to play with `httpClientFactory` dependency.
Other solution - redesign `NnnAdminService`s to have `CommunicationService` (or part of it related to SendRequest) as a dependency that could be mocked.

min4uk:
> > > You've mentioned "Removed institutionId changing, because it`s not allowed by business rules". Are there any tests enforcing this business rule? If not - could they be written?
> > 
> > 
> > I tried to, but I couldn't figure out how to write such unit test😥
> 
> @min4uk please correct me if I'm wrong. As I can see the logic of Id hiding is hidden in `CreateMap<BaseUserDto, NnnAdminBaseDto>`
@VadymLevkovskyi 
Our controller binds only that properties, that corresponds to the properties of `BaseUserDto`, so `institutionId` property will not be binded



Unless we add this property to `BaseUserDto` class or change Dto that we are using.
We have a test that checks that we are using right class `Update_WithInvalidModel_ReturnsRequestObjectResult()` (it has to be ...BadRequestObjectResult() i will fix it)
It seems that I need to check if `institutionId` property have been added to `BaseUserDto`. Can i do it this way?

VadymLevkovskyi:
@min4uk re suggested test

Long story short: it's more palliative care than the real treatment.

Details.
- Your test does what it states, and that's cool, but not enough.
Your test is really checking if property `institutionId` is not added to `BaseUserDto`. But that time when it fail you will have no idea why it was important to write this unit test. (There is no err message in neither `Assert` and test name is just doubling the code logic, not the real explanation why test was written)

- "Our controller binds only that properties, that corresponds to the properties of `BaseUserDto`" 
What if controller input parameter type will be changed? Suggested `Update_WithInvalidModel_ReturnsRequestObjectResult` test doesn't cover this case.

- So main issue of this test case is that you are checking not the controller action, but (only partially) details of implementation. That's good, but not sufficient.

DmyMi (approved):

min4uk:
@VadymLevkovskyi 
I added error messages to mapper tests, and added `Update_WorkOnlyWithBaseUserDto` test to avoid changing BaseUserDto, without undestanding consequences.
Is this ok, or did you mean that these tests are not enough in any case and i need to moq request sending somehow?

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/RegionAdminControllerTests.cs
```diff
@@ -210,17 +210,36 @@ public async Task Update_WithInvalidModel_ReturnsRequestObjectResult()
         Assert.That(result, Is.InstanceOf<BadRequestObjectResult>());
     }
 
+    [Test]
+    public async Task Update_WorkOnlyWithBaseUserDto()
+    {
+        // Arrange
+        var updateRegionAdminDto = new BaseUserDto();
```
VadymLevkovskyi: Sorry, but I can't agree with this test existence. Probably better delete it.

Details:

The test promises to `_WorkOnlyWithBaseUserDto` but when I change it in test to
`var updateRegionAdminDto = new RegionAdminDto();` 
it doesn't fail. 

If I change it in `UpdateRegionAdminAsync` parameter type, the compiler will error on test code and - see my previous comments - next developer wil just rewrite test to meet the new type.

As for me this checks only one thing - that controller calls service. The same thing that is already verified in 2 tests below.

The initial task - checking absence of "institutionId changing, because it`s not allowed by business rules" is not performed. 

For this check I expect to see in test something like:
```
//Arrange
SeedRepository(new RegionAdminDto() { InstitutionId = oldInstitutionId });
...

//Act
await regionAdminController.Update(new RegionAdminDto() { InstitutionId = otherInstitutionId });

//Assert
var newInstitutionId = GetFromRepository().InstitutionId;
Assert.That(oldInstitutionId, Is.EqualTo(newInstitutionId)); // thus we ensure that it is not changed to otherInstitutionId 
```
Ofc, it's pseudocode as `SeedRepository()` and `GetFromRepository()` are not written and they could be rather big redesign task. It's doable, but I given up to insist on it, and now ask only for this test deletion.

min4uk: @VadymLevkovskyi 
Sorry😓, I want to do this right, not just close PR. Maybe we have misunderstood each other.
Earlier I tried to create such a test as you showed, but in-memory database does not support transactions. We discussed this on meeting, and decided, to try something else. Then i tried to mock send method, but 
> Non-overridable members (here: HttpClient.SendAsync) may not be used in setup / verification expressions.

I can try another option, mock communication service(by changing inheritance with constructor dependency), I don't know any other options

min4uk: Sorry😓, I want to do this right, not just close PR. Maybe we have misunderstood each other.
Earlier I tried to create such a test as you showed, but in-memory database does not support transactions. We discussed this on meeting, and decided, to try something else. Then i tried to mock send method, but

> Non-overridable members (here: HttpClient.SendAsync) may not be used in setup / verification expressions.

**UPD.**
It appeared that i can simply disable that error of in-memory database...😔
Please, check if this test fits

VadymLevkovskyi: Balissimo! Thank you for being persistent.
There is few more percents of test coverage needed to reach the perfection - please check [Sonar report](https://sonarcloud.io/component_measures?id=ita-social-projects-oos-backend&pullRequest=1272&metric=new_coverage&view=list)

min4uk: TY for your patience. I will work on test coverage now

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AreaAdminControllerTests.cs
```diff
@@ -250,6 +252,43 @@ public async Task Update_WithErrorResponse_ReturnsStatusCodeResult()
...
+        Assert.That(result, Is.InstanceOf<BadRequestObjectResult>());
+    }
+
+    [Test]
+    public async Task Update_WhenUserIsNotAdmin_ReturnsForbidenResponse()
+    {
+        // Arrange
+        var updateAreaAdminDto = new AreaAdminDto { Id = string.Empty };
+
+        areaAdminServiceMock
```
VadymLevkovskyi: (Test logic?) Test says "_WhenUserIsNotAdmin_", but you're mocking db exception. It looks like previous test dublicate.

min4uk: It is dublicate...
I'll fix it

VadymLevkovskyi (approved):

Issue #1271, opened 2023-10-19
[Region Admin;Permissions] Region Admin can see Admins OTG from other regions


OlenaYakubova:
**Environment:** Windows 11, 64 bit
Chrome 117.0.5938.134
**Reproducible:** always.

**Preconditions**
1. Go to https://oos.dmytrominochkin.cloud/
2. Log in as Region Admin

**Steps to reproduce**
1. Go to 'Адміністрування' tab

**Actual result**
Administrators from other districts are displayed
![image](https://github.com/ita-social-projects/OoS-Backend/assets/73597578/971a18fb-27db-4847-8b7e-e392467050af)

**Expected result**
Relevant information is displayed to make permitted action according to the Region Admin role

**User story and test case links**
 "User story #478 

**Labels to be added**
"Bug", Priority ("high"), Severity ("High"), Type ( "Functional").

PR #1273, opened 2023-10-20
Added region admin CATOTTGId to filtration

min4uk:
#1271 fixed

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Services/AreaAdminService.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Services/AreaAdminService.cs
@@ -166,6 +166,7 @@ public class AreaAdminService : CommunicationService, IAreaAdminService
         {
             var regionAdminDto = await regionAdminService.GetByUserId(currentUserService.UserId).ConfigureAwait(false);
             filter.InstitutionId = regionAdminDto.InstitutionId;
+            filter.CATOTTGId = regionAdminDto.CATOTTGId;
             var childrenIds = await codeficatorService.GetAllChildrenIdsByParentIdAsync(filter.CATOTTGId);
             catottgs = childrenIds.ToList();
         }
--- a/OutOfSchool/OutOfSchool.WebApi/Services/AreaAdminService.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Services/AreaAdminService.cs
@@ -167,7 +167,7 @@ public class AreaAdminService : CommunicationService, IAreaAdminService
             var regionAdminDto = await regionAdminService.GetByUserId(currentUserService.UserId).ConfigureAwait(false);
             filter.InstitutionId = regionAdminDto.InstitutionId;
             var childrenIds = await codeficatorService.GetAllChildrenIdsByParentIdAsync(regionAdminDto.CATOTTGId);
-            if (filter.CATOTTGId != 0)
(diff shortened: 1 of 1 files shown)


OutOfSchool/OutOfSchool.WebApi/Services/AreaAdminService.cs
```diff
@@ -166,6 +166,7 @@ public async Task<SearchResult<AreaAdminDto>> GetByFilter(AreaAdminFilter filter
         {
             var regionAdminDto = await regionAdminService.GetByUserId(currentUserService.UserId).ConfigureAwait(false);
             filter.InstitutionId = regionAdminDto.InstitutionId;
+            filter.CATOTTGId = regionAdminDto.CATOTTGId;
```
VadymLevkovskyi: (Logic) You've fixed the bug, definitely, but I'm wondering if other bug is not reproduced? 
Consider the case when `GetByFilter()` is called with `filter.CATOTTGId` already set (it might be child of `regionAdminDto.CATOTTGId`) then your fix will return all admins from the region, instead of that one smaller OTG.

min4uk: Since this method is called without the ability to explicitly specify the CATOTTGId for the region admin, I thought it was okay to leave it that way.
But i did not think that this can be changed in the future. I will rework it, to work properly, when `filter.CATOTTGId` is not empty.

min4uk: Reworked, considering the case you mentioned

OutOfSchool/OutOfSchool.WebApi/Services/AreaAdminService.cs
```diff
@@ -166,7 +166,20 @@ public async Task<SearchResult<AreaAdminDto>> GetByFilter(AreaAdminFilter filter
         {
             var regionAdminDto = await regionAdminService.GetByUserId(currentUserService.UserId).ConfigureAwait(false);
             filter.InstitutionId = regionAdminDto.InstitutionId;
-            var childrenIds = await codeficatorService.GetAllChildrenIdsByParentIdAsync(filter.CATOTTGId);
+            var childrenIds = await codeficatorService.GetAllChildrenIdsByParentIdAsync(regionAdminDto.CATOTTGId);
+            if (filter.CATOTTGId != 0)
+            {
+                if (childrenIds.Contains(filter.CATOTTGId))
```
VadymLevkovskyi, line 172: (I'm not sure if it is bug b/c don't know the `codeficatorService.GetAllChildrenIdsByParentIdAsync()` functionality.)
Please check if case when `filter.CATOTTGId` == `regionAdminDto.CATOTTGId` does not throw `UnauthorizedAccessException`.

min4uk: This method includes parent Id to the result list, so it works right in this case

min4uk: > This method includes parent Id to the result list, so it works right in this case

I looked again on code, and thought that it would be better to add equality check, so `GetAllChildrenIdsByParentIdAsync` not called in case you mentioned

VadymLevkovskyi: Perfect! As this was my second question )

VadymLevkovskyi (approved):

PR #1275, opened 2023-10-20
Update Elasticsearch query for improved searching

VitaliyYuras:


OutOfSchool/OutOfSchool.ElasticsearchData/ESWorkshopProvider.cs
```diff
@@ -67,14 +67,13 @@ private QueryContainer CreateQueryFromFilter(WorkshopFilterES filter)
 
         if (!string.IsNullOrWhiteSpace(filter.SearchText))
         {
-            queryContainer &= new MultiMatchQuery()
+            queryContainer &= new QueryStringQuery()
```
VadymLevkovskyi, line 70: (Performance, not insisting if it breaks requirements) I'm not sure if it goes well with expected business logic, but from performance perspective why not prevent initial wildcards analysis using `QueryStringQuery.AllowLeadingWildcard = false` ? By default it's true and this works worser with indexes (see [doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#query-string-wildcard))

VadymLevkovskyi (approved):

PR #1276, opened 2023-10-23
Add memory cache to our external cache class

alrom:


OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
```diff
@@ -0,0 +1,125 @@
...
+            cacheLock.EnterWriteLock();
+
+            try
+            {
+                var cacheEntryOptions = new MemoryCacheEntryOptions()
+                {
+                    SlidingExpiration = slidingExpirationInterval ?? memoryCacheConfig.SlidingExpirationInterval,
+                    AbsoluteExpirationRelativeToNow = absoluteExpirationRelativeToNowInterval ?? memoryCacheConfig.AbsoluteExpirationRelativeToNowInterval,
+                    Priority = CacheItemPriority.Normal,
+                    Size = 1024,
```
VadymLevkovskyi: (Magic literals) What `1024` means here? 

Also, accordingly to [SO: What do the size settings for MemoryCache mean?](https://stackoverflow.com/questions/59774566/what-do-the-size-settings-for-memorycache-mean) you could use any number but this `Size` parameter is related to `SizeLimit` parameter of the whole cache. If latter is not set, the former does nothing.

Probably it could be removed at all

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+        });
+    }
+
+    public void Dispose()
+    {
+        cacheLock?.Dispose();
+    }
+
+    private Task ExecuteMultipleAsync(Action operation)
```
VadymLevkovskyi: Please provide more details why this method is named `ExecuteMultipleAsync` and why we need it at all?
As I can see from the usage it's only purpose is to make synchronous methods async. Why don't have it stated clearly like
```
public void ClearCache(string key) // <--- here the "cache logic" resides
{
    cacheLock.EnterWriteLock();
    try
    {
        _memoryCache.Remove(key);
        _cacheService.ClearCacheAsync(key);
    }
    finally
    {
        cacheLock.ExitWriteLock();
    }
}

public Task ClearCacheAsync(string key) // <--- here the "async logic" resides
  => Task.Run(() => ClearCache(key));
```  
Thus you have `cache` logic separated from `async` flavour.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
```diff
@@ -0,0 +1,125 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace OutOfSchool.Redis;
+
+public sealed class MultiLayerCache : IMultiLayerCacheService, IDisposable
+{
+    private readonly ReaderWriterLockSlim cacheLock = new ReaderWriterLockSlim();
```
VadymLevkovskyi: Could you explain what purpose this `cacheLock` serves? 

The MemoryCache is thread safe (see [MS doc](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.caching.memorycache?view=dotnet-plat-ext-7.0#thread-safety)).

Current implementation makes all writes **blocking** each other, so changing cache items with **diferent keys** is done **sequentially**. Is it expected? What is the trade to **loose performance**?

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
VadymLevkovskyi: The question is still open.
The suggested code change is no longer viable, as I found in other comment async call to `_cacheService.ClearCacheAsync(key);`.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
```diff
@@ -0,0 +1,125 @@
...
+
+    public Task ClearCacheAsync(string key)
+    {
+        return ExecuteMultipleAsync(() =>
+        {
+            cacheLock.EnterWriteLock();
+            try
+            {
+                _memoryCache.Remove(key);
+                _cacheService.ClearCacheAsync(key);
```
VadymLevkovskyi: `_cacheService.ClearCacheAsync(key);` - is never awaited. 
So your `ClearCacheAsync()` implementation could return "completed" task earlier than it is really completed.

VadymLevkovskyi: Same for `Refresh()`

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+            finally
+            {
+                cacheLock.ExitWriteLock();
+            }
+        });
+    }
+
+    public async Task<T> GetOrAddAsync<T>(string key, Func<Task<T>> newValueFactory, TimeSpan? absoluteExpirationRelativeToNowInterval = null, TimeSpan? slidingExpirationInterval = null)
```
VadymLevkovskyi, line 32: (Logic) For non-existing key I expect `GetOrAddAsync()` and `SetAsync()` to behave similarly. They don't! 
One is asking `_cacheService` to store the data, the other - does not.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
```diff
@@ -0,0 +1,125 @@
...
+        if (EqualityComparer<T>.Default.Equals(returnValue, default))
+        {
+            returnValue = await _cacheService.GetOrAddAsync(key, newValueFactory, absoluteExpirationRelativeToNowInterval, slidingExpirationInterval);
+            await SetAsync(key, returnValue, absoluteExpirationRelativeToNowInterval, slidingExpirationInterval);
+        }
+
+        return returnValue;
+    }
+
+    public Task RefreshAsync(string key)
```
VadymLevkovskyi: As a suggestion
```
public Task RefreshAsync(string key)
{
    cacheLock.EnterWriteLock();
    return _cacheService.RefreshAsync(key).ContinueWith(_ => cacheLock.ExitWriteLock());
}
```
This way it removes the `ExecuteMultipleAsync` and really awaits the `_cacheService.RefreshAsync()`

VadymLevkovskyi (changes requested):
The `MultiLayerCache` implementation concerns me. Please check other comments

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
alrom: Added Size to cache config.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
alrom: I read about it. But all samples, that I found, were with lockers. And I used the same approach.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
VadymLevkovskyi: That's cool, but I asked about possible removing it at all, as it does nothing until you set the `SizeLimit` per whole cache.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
VadymLevkovskyi: "Quotes found on the Internet are not always accurate" (c) Lao Tzu & (later) Abraham Lincoln

(More seriously) Could you share links to that samples?

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
alrom: I  meant that I added Size to config and changed it to whole cache SizeLimit. And deleted it from options of "set" operation.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
alrom: https://habr.com/ru/companies/otus/articles/502974/
https://code-maze.com/aspnetcore-in-memory-caching/
https://michaelscodingspot.com/cache-implementations-in-csharp-net/

And some words about MemoryCache in this article:
https://blog.novanet.no/asp-net-core-memory-cache-is-get-or-create-thread-safe/

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
alrom: Looks pretty cool. I removed using of lock because it uses inside _cacheService.RefreshAsync. And added await.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
alrom: I can`t get you an answer about naming. It smth with my deep night minds. I renamed it to ExecuteCacheMethod.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
alrom: Added awaiters.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
alrom: I deleted Set method from the interface. It doesn`t need in our project.

Sorry, but I can`t catch your mind.
Could you expand it?

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
VadymLevkovskyi: Thank you for links. It was interesting reading, especially the github code of LazyCache mentioned in last article.

At same time it doesn't answer any of my questions regarding your implementation. (TLDR: in articles I see Liverpool, in your code I see Manchester - in general both are about english football, both are about different things.)

---

But let's read them together, step-by-step. 
- 1st link "Habr" - suggests to use locks per each cache key (`ConcurrentDictionary<object, SemaphoreSlim>`). Not 1 lock for all. At same time comments explains why it is bad. Noone is suggesting to use 1 lock for all. 
(Btw, has to mention good example of `if-lock-if` pattern usage in last implementation of `WaitToFinishMemoryCache`. It's good to understand why it is better than simple `if-lock` and use it wherever possible. I.e. check your implementation of `SetAsync<T>()`. Ofc it doesn't mean, that the whole idea of `_locks` is good.)
- 2nd link "Code-Maze" - single `SemaphoreSlim` is used for isolated case of GetOrCreate. Not everywhere. Thus blocking issue is bit reduced, but still same - two parallel cache additions would block cache even when it is done for different keys.
- 3rd link "Michael's coding spot" - (it has slightly broken code, but I hope I got it) again `ConcurrentDictionary<object, SemaphoreSlim>` is used. And no surprise, b/c Habr article was translation of this one. So all issues still are still the same.
- 4th link "nova net" - mentioned same issue with isolated `GetOrCreate`, not all cache methods. As a good cache article suggests 3rd-party library - LazyCache and inside implementation [github: CachingService](https://github.com/alastairtree/LazyCache/blob/master/LazyCache/CachingService.cs#L103) you could see that it's not so easy as your implementation.

---

Summing all up. 
- Issues mentioned in my first comment are not solved in neither article except last one. 
- Code, suggested in all articles targets only one cache method `GetOrCreate`/`GetOrAdd`, not all cache methods, and "improved versions" of that code suggests to use as many locks as cache keys (or as count of hashes of cache key - for LazyCache [see lines 119-120](https://github.com/alastairtree/LazyCache/blob/master/LazyCache/CachingService.cs#L119)). 
- Please take all things written in Internet with a grain of salt, especially, if there is no comment section. 
- If you feel that you are not convinced - please call me in Teams, we could discuss it in voice.
- If you don't want to change anything and sure that your code is perfect - please state it here. I'll stop asking for improvements and will close my comments as "resolved".

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
VadymLevkovskyi: I'll try to explain
Imagine that you see someone implemented cache interface having two methods - `GetOrAdd` and `Set`. For case of non-existing in cache key you would expect them do the same things. Taking into account that this is `MultiLayerCache` I'd expect that both cache layers will be covered by these Add/Set - both `_cacheService` and `_memoryCache`. `Set` works with `_memoryCache` only.

Is it a good method naming - I'm not sure. 

As now it is not a public method - it's less important, but still not perfect.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
VadymLevkovskyi: > I meant that I added Size to config and changed it to whole cache SizeLimit. And deleted it from options of "set" operation.

I'm sorry, but I don't see it in a cache config. 
Please either name the file with changes or update your comments to be relevant to code changes

alrom: I deleted it and forgot to change my comment. Sorry.

OutOfSchool/OutOfSchool.Redis/CacheService.cs
```diff
@@ -74,48 +74,25 @@ IOptions<RedisConfig> redisConfig
...
             finally
             {
                 cacheLock.ExitWriteLock();
             }
         });
 
-    public Task RefreshAsync(string key)
-        => ExecuteRedisMethod(() => {
+    public async Task RefreshAsync(string key)
+        => await ExecuteRedisMethod(async () => {
```
VadymLevkovskyi: Please avoid adding excessive async/await. Your code could be simplified to
```
    public Task RefreshAsync(string key)
        => ExecuteRedisMethod(async () => {
            ...
```

VadymLevkovskyi: Same for `RemoveAsync` and `SetAsync`

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
```diff
@@ -0,0 +1,113 @@
...
+        IMemoryCache inMemoryCache,
+        IOptions<MemoryCacheConfig> memoryCacheConfig
+    )
+    {
+        this._memoryCache = inMemoryCache;
+        this._cacheService = externalCache;
+        this.memoryCacheConfig = memoryCacheConfig.Value;
+    }
+
+    public async Task RemoveAsync(string key)
```
VadymLevkovskyi: Excessive async/await here and in similar methods. Please check my comment in `CacheService.cs` for more details

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
VadymLevkovskyi: There is method that does only one action - calls `Task.Run`. 
1. Why it is named `ExecuteMultipleAsync`/`ExecuteCacheMethod`? What code inside this method is related to `Multiple` or to `Cache`? 
2. Why this method is needed at all? What if inline it?

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
VadymLevkovskyi: It would be cool if `CacheService` `RefreshAsync()` be written in this way. It is not.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
VadymLevkovskyi: The other small point to add to the conversation. I see it now, that CacheService is using the same locking mechanism with different lock. Hadn't analyzed it thoroughly but there might be potential for dead-locks.

alrom: My code is not perfect ))))
I like your comments and I always wait for them.
They help me improve my skills and understanding.
After reading your comments I can look at my code from another side.
And I am very grateful for the time that you spend reviewing and analyzing my PRs.

I try to change something.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
alrom: I deleted this method

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
alrom: Deleted locker

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
alrom: Deleted Refresh and from the interface

OutOfSchool/OutOfSchool.Redis/CacheService.cs
alrom: Deleted async/await

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
alrom: Deleted

VadymLevkovskyi: And it's still excessive. Please check the sharplab.io [snippet](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABADAAgwRgFYBuAWACgMBmPAJhwGEcBvCndnNj6vADjyQ4AggGcAngDswACgCUOALwA+PAE4cAWTnEOncroD0BnJAC2pmBIAuOAA6wAbgEsIAVxHDxUuTgCGEgBMcVykIc0sbUUkZeWAYABsIAHcAGhwrCBMACxgwAGscGAAzItybCCLChDAYEREnBxg+AwxVLnYjHgxBKO95ZU1tdr1dLsEteVZ9XW4AdgEAOgYw23iYKxgA7rJp9gBfCj2gA) to experiment what excessive async/await does with your code (you could there comment/uncomment one of two implementations of `Async()` method, pay attention to autogenerated `<Async>d__0` struct and imagine how it influences performance).

Here is simpler version with same functionality

```
    public Task RemoveAsync(string key)
    {
        _memoryCache.Remove(key);
        return _cacheService.RemoveAsync(key);
    }
```

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
VadymLevkovskyi: Long story short. Please check the implementation below

```
    public async Task<T> GetOrAddAsync<T>(string key, Func<Task<T>> newValueFactory, TimeSpan? absoluteExpirationRelativeToNowInterval = null, TimeSpan? slidingExpirationInterval = null)
    {
        if (_memoryCache.TryGetValue(key, out T value))
        {
            return value;
        }

        value = await _cacheService.GetOrAddAsync(key, newValueFactory, absoluteExpirationRelativeToNowInterval, slidingExpirationInterval);

        var cacheEntryOptions = new MemoryCacheEntryOptions()
        {
            SlidingExpiration = slidingExpirationInterval ?? _memoryCacheConfig.SlidingExpirationInterval,
            AbsoluteExpirationRelativeToNow = absoluteExpirationRelativeToNowInterval ?? _memoryCacheConfig.AbsoluteExpirationRelativeToNowInterval,
            Priority = CacheItemPriority.Normal,
        };

        _memoryCache.Set(key, value, cacheEntryOptions);        

        return value;
    }
```

1, There is no longer `SetAsync` present. Now you don't need to await something that is not async by its nature. Shorter code, faster performance.
2. There is no check like `if (EqualityComparer<T>.Default.Equals(returnValue, default))`. Now you don't need to constantly retry to add value if it is equal to default type value (i.e. `false` for `bool` type - this was the case for many calls of `CurrentUserService.ProviderAdminHasRights()` and similar)

OutOfSchool/OutOfSchool.Redis/CacheService.cs
```diff
@@ -170,4 +134,27 @@ private async Task ExecuteRedisMethod(Action redisMethod)
             }
         }
     }
+
+    private Task SetAsync<T>(
```
VadymLevkovskyi: (Performance) As I see the `SetAsync` is used in one place, thus can be inlined there with moving all code under one `ExecuteRedisMethod` lambda. The less separate lambdas, separate tasks, context switching - the faster it is.

As result, you could avoid excessive resetting cache value (when it is == default(T)) and rewrite that `if (EqualityComparer<T>.Default.Equals(returnValue, default))` part of logic in similar way as was suggested for `MultiLayerCache`.

VadymLevkovskyi: It could look like
```
    public async Task<T> GetOrAddAsync<T>(
        string key,
        Func<Task<T>> newValueFactory,
        TimeSpan? absoluteExpirationRelativeToNowInterval = null,
        TimeSpan? slidingExpirationInterval = null)
    {
        T returnValue = default;

        await ExecuteRedisMethod(async () =>
        {
            cacheLock.EnterUpgradeableReadLock();
            try
            {
                var value = await cache.GetStringAsync(key);

                if (value != null)
                {
                    returnValue = JsonConvert.DeserializeObject<T>(value);
                    return;
                }

                cacheLock.EnterWriteLock();
                try
                {
                    returnValue = await newValueFactory();
                    var options = new DistributedCacheEntryOptions
                    {
                        AbsoluteExpirationRelativeToNow = absoluteExpirationRelativeToNowInterval ?? redisConfig.AbsoluteExpirationRelativeToNowInterval,
                        SlidingExpiration = slidingExpirationInterval ?? redisConfig.SlidingExpirationInterval,
                    };

                    await cache.SetStringAsync(key, JsonConvert.SerializeObject(returnValue), options);
                }
                finally
                {
                    cacheLock.ExitWriteLock();
                }
            }
            finally
            {
                cacheLock.ExitUpgradeableReadLock();
            }
        });

        return returnValue;
    }
```

That could be further simplified having `Task<T> ExecuteRedisMethod<T>(Func<Task<T>>)` version implemented.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
alrom: Done. Deleted Set method.

OutOfSchool/OutOfSchool.Redis/MultiLayerCache.cs
alrom: Done

OutOfSchool/OutOfSchool.Redis/CacheService.cs
alrom: Deleted set.

VadymLevkovskyi:
Have no more comments from my side. Waiting for PR be in not draft mode

DmyMi (approved):

PR #1288, opened 2023-11-10
Yuras/block parent can not send message

VitaliyYuras:


OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs
```diff
@@ -207,6 +207,8 @@ public MappingProfile()
             .ForMember(dest => dest.ImageFiles, opt => opt.Ignore())
             .ForMember(dest => dest.ImageIds, opt => opt.MapFrom(src => src.Images.Select(x => x.ExternalStorageId)));
 
+        CreateMap<ProviderDto, ProviderCreateDto>();
```
DmyMi: do we really need a map from `ProviderDto` to `ProviderCreateDto`? Where is it used?

DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs
VitaliyYuras11: > do we really need a map from `ProviderDto` to `ProviderCreateDto`? Where is it used?

Fixed . Really mapping is not used

PR #1290, opened 2023-11-13
Nazarstig/providers separation

nazarstig:


OutOfSchool/OutOfSchool.WebApi/Services/ProviderServices/PrivateProviderService.cs
```diff
@@ -0,0 +1,118 @@
...
+    /// <param name="providerAdminService">Service for getting provider admins and deputies.</param>
+    /// <param name="institutionAdminRepository">Repository for getting ministry admins.</param>
+    /// <param name="currentUserService">Service for manage current user.</param>
+    /// <param name="ministryAdminService">Service for manage ministry admin.</param>
+    /// <param name="regionAdminService">Service for managing region admin rigths.</param>
+    /// <param name="codeficatorService">Codeficator service.</param>
+    /// <param name="regionAdminRepository">RegionAdminRepository</param>
+    /// <param name="averageRatingService">Average rating service.</param>
+    /// <param name="areaAdminService">Service for manage area admin.</param>
+    public PrivateProviderService(
```
VadymLevkovskyi, line 39: (Code style, design) Please explain why `PrivateProviderService` should be descendant of `ProviderService`?
As for me, if you use "composition over inheritance" principle (and make `IProviderService` as a dependency to `PrivateProviderService`) you'll get better testability, and dramatically decrease count of input params in this constructor. I'm not even sure if `PrivateProviderService` should implement `IProviderService`, as those 2 dependencies could be injected separately.

Same for `PublicProviderService`.

OutOfSchool/OutOfSchool.WebApi/Services/ProviderServices/ProviderService.cs
```diff
@@ -500,6 +421,36 @@ async Task<IEnumerable<string>> INotificationReciever.GetNotificationsRecipientI
         return recipientIds.Distinct();
     }
 
+    private protected async Task SendNotification(
```
VadymLevkovskyi, line 424: (Code style) I know, that this method was introduced earlier, but as you've touched it - please make it more tidy:
- fix input parameters indentation,
- invert `if (provider != null)` condition to have early return and less nesting.

VadymLevkovskyi (commented):
See comments

PR #1293, opened 2023-11-13
Kymachynskyi/add block/unblock parent for tech and ministry admin, add logging for these operations

DKymachynskyi:


OutOfSchool/OutOfSchool.DataAccess/Repository/EntityRepositoryBase.cs
```diff
@@ -161,6 +161,19 @@ public virtual async Task<IEnumerable<TEntity>> GetAllWithDetails(string include
     /// <inheritdoc/>
     public virtual Task<TEntity> GetById(TKey id) => dbSet.FirstOrDefaultAsync(x => x.Id.Equals(id));
 
+    /// <inheritdoc/>
+    public virtual Task<TEntity> GetByIdWithDetails(TKey id, string includeProperties = "")
+    {
+        var query = dbSet.Where(x => x.Id.Equals(id));
+        foreach (var includeProperty in includeProperties.Split(
```
VadymLevkovskyi: (Code style) I understand, that this code was copied from other implementation, thus have few questions. 
1. What if `includeProperties == null`? 
2. Why not extract this `foreach` block into own method?

VadymLevkovskyi: Same for `EntityRepositorySoftDeleted`

OutOfSchool/OutOfSchool.WebApi/Models/Parent/ParentBlockByAdminDto.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace OutOfSchool.WebApi.Models.Parent;
+
+public class ParentBlockByAdminDto
+{
+    [Required]
+    public Guid ParentId { get; set; }
+
+    [Required]
+    public bool IsBlocked { get; set; }
```
VadymLevkovskyi: (Naming) I don't get how `IsBlocked` could be `false` for the entity named `ParentBlockByAdminDto` (and controller action named `BlockParent`)?

Probably it should be `...ToggleBlock...` or something similar in name of both to show that it also could **unblock**?

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ParentController.cs
```diff
@@ -60,4 +62,34 @@ public async Task<ActionResult<ParentDTO>> GetProfile()
...
+    public async Task<ActionResult> BlockParent(ParentBlockByAdminDto parentBlockByAdmin)
+    {
+        var result = await serviceParent.BlockParent(parentBlockByAdmin).ConfigureAwait(false);
+
+        if (result.Succeeded)
+        {
+            return Ok();
+        }
+
+        if (result.OperationResult.Errors.Any(x => x.Code == "404"))
```
VadymLevkovskyi: (Code style) Please avoid usage of literals in code. Use constants instead. I'm talking about `"404"`.

OutOfSchool/OutOfSchool.WebApi/Services/ParentService.cs
```diff
@@ -122,6 +129,40 @@ public Task<ShortUserDto> Update(ShortUserDto dto)
...
+    /// <inheritdoc/>
+    public async Task<Result<bool>> BlockParent(ParentBlockByAdminDto parentBlock)
+    {
+        logger.LogInformation("Changing Block status of Parent by ParentId started. Looking ParentId is {Id}", parentBlock.ParentId);
+        var parent = await repositoryParent.GetByIdWithDetails(parentBlock.ParentId, "User").ConfigureAwait(false);
+        if (parent is null)
+        {
+            return Result<bool>.Failed(new OperationError
+            {
+                Code = "404",
```
VadymLevkovskyi: (Code style) Beforementioned literals aren't good for your code maintainability. Use constants @DKymachynskyi !

Btw I'm not sure if you should return different codes for not found/already (un)blocked parents. Was it in initial task?Providing info about parentId absence/existence you give security leakage that is not somehow logged by `parentBlockedByAdminLogService.SaveChangesLogAsync`.

VadymLevkovskyi (changes requested):
Please 
- [x] - check other comments, 
- [x] - fix SonarQube warnings (interface signature doesn't math implementation signature),
- [x] - and add unit tests - well done, coverage is 91% !!

OutOfSchool/OutOfSchool.DataAccess/Repository/EntityRepositoryBase.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.WebApi/Models/Parent/ParentBlockByAdminDto.cs
DKymachynskyi: I came up with the names BlockStatus and BlockAction, but ToggleBlock fits better, so I used it. Thank you!

OutOfSchool/OutOfSchool.DataAccess/Repository/EntityRepositoryBase.cs
VadymLevkovskyi: Thanks for changes, but there might be few improvements.
1. As `EntityRepositorySoftDeleted` inherits `EntityRepositoryBase` you could have only 1 implementation of `IncludePropertiesToQuery()` in base class.
2. As `IncludePropertiesToQuery()` method is dependent only to `IQueryable` (not the RepositorySomething) you could easily decribe this in code using extension method. 
3. Please consider split parameter (comma char - ',') as a wide-use constant, so 
4. Please use "fail-fast" approach for all checks of prerequisites or "fastlanes", at lest it reduces nesting/code complexity, at most - improves code readability
5. Please avoid regular memory reallocation (`new char[]`) using instead static value.

See suggested fix:
```
public static class QueryableExtension
{
    public const char PROPERTIES_SEPARATOR = ',';
    private static char[] propertiesSplitSymbols = new char[] { PROPERTIES_SEPARATOR };

    public static IQueryable<T> IncludeProperties<T>(this IQueryable<T> query, string properties) where T: class
    {
        ArgumentNullException.ThrowIfNull(query, nameof(query));
        if (string.IsNullOrWhiteSpace(properties))
            return query;

        foreach (var property in properties.Split(propertiesSplitSymbols, StringSplitOptions.RemoveEmptyEntries))
            query = query.Include(property);

        return query;
    }
}
```

OutOfSchool/OutOfSchool.WebApi/Models/Parent/ParentBlockByAdminDto.cs
VadymLevkovskyi: Sorry for being not specific, but I was more expecting for changes in clas/method names, not property name.
Reasoning: 
- `ToggleBlock = false` is not always clear if it means do not toggle or toggle to unblocked status
- at same time `BlockParent()` - usually is treated as block only.

Please reconsider your changes

DKymachynskyi: @VadymLevkovskyi I propose BlockUnblockParent or UpdateParentBlockStatus.

VadymLevkovskyi: As you wish, just do it for both controller method name (currently `BlockParent()`) and class name (currently `ParentBlockByAdminDto`), while keeping `IsBlocked` property as is.

OutOfSchool/OutOfSchool.DataAccess/Repository/EntityRepositoryBase.cs
DKymachynskyi: I have implemented these improvements and taken note of your recommendations. Thank you!

VadymLevkovskyi: (Code style, optional) And now you could do one last touch of "perfection" ))

In all places where `IncludeProperties()` is used you could write "oneliners". 
So instead of
```
        IQueryable<TEntity> query = dbSet;
        query = query.IncludeProperties(includeProperties);
        return await query.ToListAsync();
```
it could be
```
        => dbSet.IncludeProperties(includeProperties).ToListAsync();        
```
Consider removal of excessive `await`.

DKymachynskyi: Sonar Cloud has a new issue related to the extension method line: `ArgumentNullException.ThrowIfNull(query, nameof(query));` It proposes to remove an argument `nameof(query)`. Can I remove it?

OutOfSchool/OutOfSchool.WebApi/Models/Parent/ParentBlockByAdminDto.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.DataAccess/Repository/EntityRepositoryBase.cs
VadymLevkovskyi: Definitely! As documentation says about 2nd parameter:
"The name of the parameter with which argument corresponds. If you omit this parameter, the name of argument is used."
So that's better to have less code and make Sonar happy

DKymachynskyi: > (Code style, optional) And now you could do one last touch of "perfection" ))
> :
> Consider removal of excessive `await`.

I propose to separate this task to a new issue. To accomplish this task (removing the excessive await), it is necessary to change signatures of the methods in interface IEntityAddOnlyRepository and in the methods themselves from IEnumerable<TEntity> to List<TEntity>:
`Task<IEnumerable<TEntity>> GetAllWithDetails(string includeProperties = "") `-> 
`Task<List<TEntity>> GetAllWithDetails(string includeProperties = "")`
And then it is necessary to check all calls these methods.

VadymLevkovskyi: I agree that wording "last touch" was too optimistic. 

As suggested change induces or is related to 

- possible usage of cancellation tokens during EF queries, 
- changes of repo methods' return types to `IAsyncEnumerable<TEntity>` (and thus necessity of `System.Linq.Async` package) or `Tas<List<TEntity>>` that is more common, though not always performance friendly

it is easier to make it in separate PR. 

@DKymachynskyi , please, create a tech debt issue with information from this and beforementioned comment

DKymachynskyi: I have created the issue. Thank you!

DmyMi (approved):

DKymachynskyi:
> Please
> 
> * [x]  - check other comments,
> * [x]  - fix SonarQube warnings (interface signature doesn't math implementation signature),
> * [ ]  - and add unit tests - not done, coverage is 18%

@VadymLevkovskyi I have added unit tests so if you don't mind we will merge this PR because I need it to work on the next task. Thank you!

VadymLevkovskyi (approved):

PR #1294, opened 2023-11-13
Add new state to ShowBlocked of ApplicationFilter

alrom:


VadymLevkovskyi:
Logic seems OK. 
What about unit tests?

alrom:
We never tested filters which use for making predicates. I'll search for how to check the predicate condition.
Could you approve this PR and leave unit tests like my tech debt? Because front needs these changes.

VadymLevkovskyi:
> We never tested filters which use for making predicates. I'll search for how to check the predicate condition. Could you approve this PR and leave unit tests like my tech debt? Because front needs these changes.

1. As changes were done in private predicate building method you should check not the method, but data returned by service. I.e. results of `ApplicationService.GetAll()` called with different `filter.ShowBlocked` values on known collection.
2. I've checked the existing unit-tests and see 8 tests for mentioned `ApplicationService.GetAll()`. I belive it's really simple task to add another 3 tests for different `ShowBlocked` values. 
3. I doubt that front requires code without unit-tests. Please never use this as an excuse. 

Approved, though disappointed

VadymLevkovskyi (approved):

alrom:
I added tests

VadymLevkovskyi:
> I added tests

Thanks, man

PR #1302, opened 2023-11-17
Implement ChatWorkshop/user/unreadMessagesCount endpoint

olsavchenk:
Implemented ChatWorkshop/user/unreadMessagesCount(https://github.com/ita-social-projects/OoS-Frontend/issues/2294)
Fixed ChatWorkshop/parent/chatrooms endpoint

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi.Tests/Services/ChatRoomWorkshopServiceTests.cs
+++ b/OutOfSchool/OutOfSchool.WebApi.Tests/Services/ChatRoomWorkshopServiceTests.cs
@@ -461,6 +461,8 @@ public class ChatRoomWorkshopServiceTests
             new ChatRoomWorkshopForChatList { NotReadByCurrentUserMessagesCount = 1 },
         };
 
+        var expectedCount = chatRooms.Count(r => r.NotReadByCurrentUserMessagesCount > 0);
+
         roomWithSpecialModelRepositoryMock
             .Setup(x => x.GetByParentIdAsync(parentId, It.IsAny<bool>()))
             .ReturnsAsync(chatRooms);
@@ -469,9 +471,8 @@ public class ChatRoomWorkshopServiceTests
         var result = await roomService.GetCurrentUserUnreadMessagesCountAsync(parentId, role).ConfigureAwait(false);
 
         // Assert
-        Assert.IsInstanceOf<int>(result);
         Assert.IsNotNull(result);
-        Assert.AreEqual(2, result);
+        Assert.AreEqual(expectedCount, result);
     }
 
     [Test]
@@ -487,6 +488,8 @@ public class ChatRoomWorkshopServiceTests
             new ChatRoomWorkshopForChatList { NotReadByCurrentUserMessagesCount = 1 },
         };
 
+        var expectedCount = chatRooms.Count(r => r.NotReadByCurrentUserMessagesCount > 0);
+
         roomWithSpecialModelRepositoryMock
             .Setup(x => x.GetByProviderIdAsync(providerId, It.IsAny<bool>()))
             .ReturnsAsync(chatRooms);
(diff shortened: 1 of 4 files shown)


OutOfSchool/OutOfSchool.WebApi.Tests/Services/ChatRoomWorkshopServiceTests.cs
```diff
@@ -448,6 +448,73 @@ public async Task GetUniqueChatRoom_WhenRoomDoesNotExist_ShouldReturnNull()
...
+            .Setup(x => x.GetByParentIdAsync(parentId, It.IsAny<bool>()))
+            .ReturnsAsync(chatRooms);
+
+        // Act
+        var result = await roomService.GetCurrentUserUnreadMessagesCountAsync(parentId, role).ConfigureAwait(false);
+
+        // Assert
+        Assert.IsInstanceOf<int>(result);
+        Assert.IsNotNull(result);
+        Assert.AreEqual(2, result);
```
VadymLevkovskyi: (Code style) `2` is a magic number. Don't use them! 
For this specific case you could instead use 
`var expectedCount = chatRooms.Count(r => r.NotReadByCurrentUserMessagesCount > 0)`.
Same in test below.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ChatRoomWorkshopServiceTests.cs
```diff
@@ -448,6 +448,73 @@ public async Task GetUniqueChatRoom_WhenRoomDoesNotExist_ShouldReturnNull()
...
+
+        roomWithSpecialModelRepositoryMock
+            .Setup(x => x.GetByParentIdAsync(parentId, It.IsAny<bool>()))
+            .ReturnsAsync(chatRooms);
+
+        // Act
+        var result = await roomService.GetCurrentUserUnreadMessagesCountAsync(parentId, role).ConfigureAwait(false);
+
+        // Assert
+        Assert.IsInstanceOf<int>(result);
```
VadymLevkovskyi: (Excessive code) Please, don't test the compiler. If code changes to other return type you'll see compiler error (where it is valuable) earlier than this unit test will fail.

OutOfSchool/OutOfSchool.WebApi/Services/ChatRoomWorkshopService.cs
```diff
@@ -432,6 +434,22 @@ public async Task<ChatRoomWorkshopDto> GetUniqueChatRoomAsync(Guid workshopId, G
...
+        {
+            chatrooms = await GetByParentIdAsync(parentOrProviderId);
+        }
+
+        if (userRole == Role.Provider)
+        {
+            chatrooms = await GetByProviderIdAsync(parentOrProviderId);
+        }
+
+        return chatrooms.Count(chatroom => chatroom.NotReadByCurrentUserMessagesCount != 0);
```
VadymLevkovskyi, line 446: (Possible NRE) 
If `userRole` is neither `Role.Parent` nor `Role.Provider` - you'll get null set to `chatrooms` and thus `NullReferenceException` on `.Count()`. 
If it is a part of requirement (I'll left out of discussion downsides of the exeption-driven control flow approach), then please state it directly in code.
I.e. 
```
    public async Task<int> GetCurrentUserUnreadMessagesCountAsync(Guid parentOrProviderId, Role userRole)
    {
        var chatrooms = userRole switch {
            Role.Parent => await GetByParentIdAsync(parentOrProviderId),
            Role.Provider => await GetByProviderIdAsync(parentOrProviderId),
            _ => throw new ArgumentOutOfRangeException(nameof(userRole)),
        };

        return chatrooms.Count(chatroom => chatroom.NotReadByCurrentUserMessagesCount != 0);
    }
```

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ChatRoomWorkshopServiceTests.cs
```diff
@@ -448,6 +448,75 @@ public async Task GetUniqueChatRoom_WhenRoomDoesNotExist_ShouldReturnNull()
...
+
+        roomWithSpecialModelRepositoryMock
+            .Setup(x => x.GetByParentIdAsync(parentId, It.IsAny<bool>()))
+            .ReturnsAsync(chatRooms);
+
+        // Act
+        var result = await roomService.GetCurrentUserUnreadMessagesCountAsync(parentId, role).ConfigureAwait(false);
+
+        // Assert
+        Assert.IsNotNull(result);
```
VadymLevkovskyi: What does `Assert.IsNotNull(result);` when `result` is of `int` type? Could `int` be `null`? Why are you checking it?

olsavchenk: Understood and fixed

VadymLevkovskyi (approved):

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ChatRoomWorkshopServiceTests.cs
VadymLevkovskyi: Thank you

PR #1309, opened 2023-11-24
Kymachynskyi/ Add history of changes for Parent blocked by TechAdmin, MinistryAdmin

DKymachynskyi:


VadymLevkovskyi (approved):
Seems valid, though there is lack of tests. 
Could you write simple tests with different filters and check if it works on fake data?

OutOfSchool/OutOfSchool.WebApi/Services/ChangesLogService.cs
```diff
@@ -13,6 +13,9 @@ public class ChangesLogService : IChangesLogService
 {
     public const char WORD_SEPARATOR_SPACE = ' ';
     public const char WORD_SEPARATOR_COMMA = ',';
+
+    private static readonly TimeSpan TillEndOfDay = new(23, 59, 59);
```
VadymLevkovskyi: (Logic minor issue) Please never use such approach. 
1. (Readability) If you want to have 1 second less than 1 usual day, please specify it clearly: `TimeSpan.FromDays(1) - TimeSpan.FromSeconds(1)`
2. (Logic) But the whole idea of "1 second less is broken" 
- there are leap seconds sometimes (extra second before midnight)
- you are ignoring fractional time 23:59:59,xxx
- and there is better approach - just check for `x.UpdatedDate < filter.DateTo.Value.Date.AddDays(1)`
- and more better one - introduce and use extension method for DateTime, like 
`DateTime NextDayStart(this DateTime value) => value.Date.AddDays(1);`

DKymachynskyi: Thank you! I added extension method for DateTime, with DateTime.MaxValue (9999-12-31) it throw ArgumentOutOfRangeException. To avoid this I suggest the following:
```
DateTime NextDayStart(this DateTime value) =>
        value.Date == DateTime.MaxValue.Date ? DateTime.MaxValue : value.Date.AddDays(1);
```

VadymLevkovskyi (approved):

PR #1310, opened 2023-11-27
Nazarstig/providers separation 2

nazarstig:
Create for private and public provider separate services and controllers for functionality that belongs only for that type of provider.


OutOfSchool/OutOfSchool.WebApi.Tests/Services/ProviderServicesTests/ProviderTestsHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        bool UserExist(string userId)
+        {
+            userExistsResult = providersCollection.Any(p => p.UserId.Equals(userId));
+            return userExistsResult;
+        }
+
+        providersRepository.Setup(r => r.ExistsUserId(It.IsAny<string>()))
+            .Callback<string>(user => UserExist(user))
+            .Returns(() => userExistsResult);
```
VadymLevkovskyi: (Readablitiy) I believe this method mock could be simplified to (see [Moq/wiki/Quickstart](https://github.com/devlooped/moq/wiki/Quickstart#methods)):
```
providersRepository.Setup(r => r.ExistsUserId(It.IsAny<string>()))
            .Returns(user => UserExist(user));
```

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ProviderServicesTests/ProviderTestsHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+    public static Mock<IProviderRepository> CreateProvidersRepositoryMock(IEnumerable<Provider> providersCollection)
+    {
+        var providersRepository = new Mock<IProviderRepository>();
+        var userExistsResult = false;
+
+        bool UserExist(string userId)
+        {
+            userExistsResult = providersCollection.Any(p => p.UserId.Equals(userId));
+            return userExistsResult;
+        }
```
VadymLevkovskyi: (Readability) Taking into account comment below, you could get rid of `userExistsResult` and `UserExist()` method and inline it in mock `.Returns` lambda.

OutOfSchool/OutOfSchool.WebApi/Services/ProviderServices/ProviderService.cs
```diff
@@ -421,41 +394,43 @@ public async Task<bool> IsBlocked(Guid providerId)
         return (await providerRepository.GetById(providerId).ConfigureAwait(false)).IsBlocked;
     }
 
-    public async Task<ProviderLicenseStatusDto> UpdateLicenseStatus(ProviderLicenseStatusDto dto, string userId)
+    public async Task SendNotification(Provider provider, NotificationAction notificationAction, bool addStatusData, bool addLicenseStatusData)
```
VadymLevkovskyi: (Performance, readability) You could remove async/await pair in this method, as only awaitable operation is last one.

OutOfSchool/OutOfSchool.WebApi/Services/ProviderServices/IPrivateProviderService.cs
```diff
@@ -0,0 +1,17 @@
+﻿using OutOfSchool.WebApi.Models.Providers;
+
+namespace OutOfSchool.WebApi.Services.ProviderServices;
+
+/// <summary>
+/// Defines interface for functionality for Private Provider.
+/// </summary>
+public interface IPrivateProviderService
```
VadymLevkovskyi, line 8: (It's more question than recommendation) I see interface with one method, that are named differently. Is this i-face supposed to be extended by more methods? If not, than elaborate naming (of i-face and related class). Same for `IPublic...`

VadymLevkovskyi (commented):
See no logic issues but have few comments

OutOfSchool/OutOfSchool.WebApi/Services/ProviderServices/IPrivateProviderService.cs
nazarstig: Probably it will be extended, I believe this names can be at this stage.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ProviderServicesTests/ProviderServiceV2Tests.cs
```diff
@@ -0,0 +1,580 @@
...
+            authorizationServerConfig,
+            communicationService.Object);
+        providersRepositoryMock.Setup(w => w.UnitOfWork.CompleteAsync()).ReturnsAsync(It.IsAny<int>());
+    }
+
+    #region Create
+
+    [TestCase(null, ProviderLicenseStatus.NotProvided)]
+    [TestCase("1234567890", ProviderLicenseStatus.Pending)]
+    public async Task Create_WhenEntityIsValid_ReturnsCreatedEntity(string license, ProviderLicenseStatus expectedLicenseStatus)
```
VadymLevkovskyi, line 127: (Tests completeness) `providerImagesService` methods has to be called after provider creation (via `Create()` method) if there is cover image or image files set. This is not tested.

Same for `Delete()`.

VadymLevkovskyi (approved):

PR #1311, opened 2023-11-27
Fixed possible error for Workshop update endpoint while converting null for available seats

olsavchenk:


OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/WorkshopServiceTests.cs
```diff
@@ -347,6 +347,28 @@ public async Task Update_WhenEntityIsValid_ShouldReturnUpdatedEntity([Random(2,
...
+        changedFirstEntity.Provider = provider;
+        SetupUpdate(changedFirstEntity);
+        var expectedTeachers = teachers.Select(s => mapper.Map<TeacherDTO>(s));
+
+        // Act
+        var result = await workshopService.Update(mapper.Map<WorkshopBaseDto>(changedFirstEntity)).ConfigureAwait(false);
+
+        // Assert
+        result.Teachers.Should().BeEquivalentTo(expectedTeachers);
+        result.AvailableSeats.Should().Be(uint.MaxValue);
```
VadymLevkovskyi, line 396: (Tets) Please add same `AvailableSeats` assertion to `Create()` method tests, as you've changed that service method too

VadymLevkovskyi: Thanks for changing, but please mind test coverage. Seems more test should be written

olsavchenk: I don't really get what I have to do. Should I add more test cases?

VadymLevkovskyi (approved):

PR #1313, opened 2023-11-29
Tfront1/custom error responses

### LINKED ISSUE #1307: Create a unified way to have custom error responses.
Based on this [PR](https://github.com/ita-social-projects/OoS-Backend/pull/843) and/or maybe googling 😄  create a unified error response model that can be extended with different error types/codes to notify frontend of specific errors because sometime simple `400` http code is not enough.

Tfront1:
closes #1307 

I would like some feedback for this solution, if it is ok then I will complete it for all other cases.


OutOfSchool/OutOfSchool.Common/Responses/ApiErrorsTypes.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.Diagnostics;
+using System.Reflection;
+using OutOfSchool.Common.Responces;
+
+namespace OutOfSchool.Common.Responses;
+public abstract class ApiErrorsTypes
+{
+    private static string GetGroup()
+    {
```
strivitech, line 5: Is it necessary to dynamically determine the declaring names? As far as I understand, you want to return "ProviderAdmin/Creation" here. Maybe better use constants approach (same as with codes) not to retrieve the same info with reflection on every single call as it reduces performance and forces the use of 2 nesting classes.

OutOfSchool/OutOfSchool.Common/Responses/ApiErrorsTypes.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.Diagnostics;
+using System.Reflection;
+using OutOfSchool.Common.Responces;
+
+namespace OutOfSchool.Common.Responses;
+public abstract class ApiErrorsTypes
```
strivitech: Use static class here and in nested classes.

OutOfSchool/OutOfSchool.Common/Responses/ApiErrorResponse.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        this.apiErrors = new List<ApiError>(apiErrors);
+    }
+
+    public ApiErrorResponse()
+    {
+        this.apiErrors = new List<ApiError>();
+    }
+
+    public List<ApiError> ApiErrors => apiErrors;
```
strivitech: Just comment. This basically opens full access to `List<ApiError>` and auto property `public List<ApiError> ApiErrors {get;}` can be used instead of field + property. If your purpose to restrict operations on returned list, you could use `IReadOnlyList<ApiError>` with `.AsReadOnly()`.

OutOfSchool/OutOfSchool.Common/Responses/ApiErrorResponse.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    public void AddApiError(ApiError apiError)
```
strivitech, line 20: Just comment. As I mentioned earlier, `List<ApiError>` is public, so all this method does is "convenient" way to add a new error.

OutOfSchool/OutOfSchool.Common/Responses/ApiErrorsTypes.cs
```diff
@@ -0,0 +1,51 @@
...
+    public abstract class ProviderAdmin
+    {
+        private static int entityCode = 10;
+
+        public abstract class Creation
+        {
+            private static int actionCode = 1;
+
+            public static ApiError UserDontHavePermission(string userId) =>
+            new ApiError(
```
strivitech: Maybe I don't get the logic but why do you need both code and group if they both represent the same logical information ProviderAdmin -> Creation.

If I understand the problem, you want frontend to show some messages based on the received code.
What is enough to frontend?
Probably, they want you to directly give them unique keys, so they can search in their dictionary for the appropriate messages.
Why can't we simplify and use one of the following ideas:
1) { Code = "ProviderAdmin.UserDontHavePermissionToCreate", Message = "..." }
2) { Code = "UserDontHavePermissionToCreate", Group = "ProviderAdmin", Message = "..." }

If you are worried about writing to much strings, you are able to use `$nameof(SomeGroup).nameof(CurrentMethod)`. Although, we have to be sure we don't want to rename classes and methods as it leads to breaking changes on frontend side.
You could go with numbers approach as well to reduce response size but strings are more explicit maybe.

OutOfSchool/OutOfSchool.Common/Responses/ApiErrorsTypes.cs
```diff
@@ -0,0 +1,51 @@
...
+        string group = declaringType.DeclaringType.Name;
+
+        return $"{group}/{subGroup}";
+    }
+
+    public abstract class ProviderAdmin
+    {
+        private static int entityCode = 10;
+
+        public abstract class Creation
```
strivitech: I don't think it makes a lot of sense to create this nesting level as it adds complexity: Creation, Deletion, etc. Look at my comment above.

OutOfSchool/OutOfSchool.WebApi/Services/ApiErrors/ApiErrorService.cs
```diff
@@ -0,0 +1,18 @@
+﻿using OutOfSchool.Common.Responces;
+
+namespace OutOfSchool.WebApi.Services.ApiErrors;
+
+public class ApiErrorService : IApiErrorService
```
strivitech: Just comment. This looks like cross cutting service. To work properly, it requires either return Error state from service (from Either<>) as part of flow or throw new exception. If we return an Error or throw an exception, why don't put Code and Message in there directly? I will describe below.

OutOfSchool/OutOfSchool.WebApi/Services/ProviderAdminService.cs
```diff
@@ -61,7 +65,7 @@ public class ProviderAdminService : CommunicationService, IProviderAdminService
         if (!hasAccess)
         {
             Logger.LogError("User(id): {UserId} doesn't have permission to create provider admin", userId);
-
+            apiErrorService.AddApiError(ApiErrorsTypes.ProviderAdmin.Creation.UserDontHavePermission(userId));
             return new ErrorResponse
```
strivitech, line 65: Is it possible to extend ErrorResponse? Instead of `HttpStatusCode` and `Message`, it would return `HttpStatusCode` and `ApiError` or even better `List<ApiError>`. This way it directly contains all necessary information.

Based on the code base, we have both Either flow and Exceptions flow, so I'd like to offer some idea for the second one.
We could create some custom exception (let's say called "OperationException") that contain the same `HttpStatusCode` and `List<ApiError>`. This would allow us to catch it on the middleware side and return appropriate response to user. If needed, you could catch it earlier on controller side (don't know why).
This suggestion based on the already existing services throwing `InvalidOperationException`, `ArgumentOutOfRange`, ... with messages inside caught by controller level.

@DmyMi

strivitech:
Please don't do any changes before team review as I am not up to date with the codebase.

By _provicevko_ :smile:.

OutOfSchool/OutOfSchool.Common/Responses/ApiErrorsTypes.cs
```diff
@@ -0,0 +1,51 @@
...
+        MethodBase method;
+
+        try
+        {
+            method = frames[1].GetMethod();
+        }
+        catch
+        {
+            return "undefined";
+        }
```
VadymLevkovskyi: (Performance, readability) Please avoid exception-driven control flow. It's ususally non-obvious and bad from performance perspective. 
As for me (if I'm not missing something) you could just check if non-nullable `frame[1]` exists - this simple if will be faster than creating `catch` context and throwing/catching exception.

Ofc, the @strivitech comment is still valid and it's better to avoid using of reflection at all.

OutOfSchool/OutOfSchool.WebApi/Services/ProviderAdminService.cs
VadymLevkovskyi: @strivitech I'm not Dmytro, but have a question. 

Why don't use exceptions control flow at all? As we already have Either monad, developers get used to use it, so let's avoid exceptions wherever possible. 

Ofc exceptions are simple enough for call stack unrolling (and skipping few stack frames, till the first catch), but at same time they give lack of clarity, and provoke to grow your own "schizophrenia" for error handling. Also developers are used to use all "try catch bad practices" - to catch all exceptions, use `try`/`catch` when simple `if` is possible, therefore, maybe, it's time to forget them )

Summing up, let's do error handling in only way not both ways. What do you think?

VadymLevkovskyi:
Let me add my 5 cents. 

The whole idea of this `ApiErrorService` is an `Ambient context` pattern implementation. 

This pattern adds indirectness, possibility to have unexpected control flows (esp. when you need for this service narrower scope  than provided by `services.AddScoped<IApiErrorService..`), it's hard to debug. That's why my instinctive reaction to such pattern is to avoid it by any means, until someone convince me that there is no other way to implement the feature.

Please convince me )

OutOfSchool/OutOfSchool.WebApi/Services/ProviderAdminService.cs
strivitech: @VadymLevkovskyi I think, ideally, it's good practice to have it one way. I can't say that using Either monads always better as it adds more "checks" on all nodes of chain calls and sometimes I would be glad to throw 😄. I agree that Either approach gives more clarity as it directly describes how logic works.

So, basically, I offered some workaround how to adapt existing code to use new ApiError classes.  I guess it's simpler to start with than change everything on Either. Can be wrong though.

VadymLevkovskyi (commented):
Changes seems to be OK. Code coverage is low.

OutOfSchool/OutOfSchool.Common/Responses/ApiError.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+
+namespace OutOfSchool.Common.Responces;
```
VadymLevkovskyi: (Mistype) Namespace `Responces` should contain no "c" in it's name. Please fix it. Check also other files.

VadymLevkovskyi (approved):
Seems good )

PR #1314, opened 2023-11-30
Fix error with provider admin creation endpoint

olsavchenk:


OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ProviderAdminController.cs
```diff
@@ -3,6 +3,7 @@
 using Microsoft.AspNetCore.Mvc.Filters;
 using OutOfSchool.Common.Models;
 using OutOfSchool.Services.Enums;
+using OutOfSchool.Services.Models;
```
VadymLevkovskyi: (Code style) I'm not sure if ithis using is required. Could you check it?

VadymLevkovskyi:
(Logic) I don't see how changes in http response status code could be a fix, but it the least issue. 
The more important issue is absence of related unit-/integration test. 
Is it possible to be written?

VadymLevkovskyi (approved):

PR #1317, opened 2023-12-06
Cache control header was added

alrom:


OutOfSchool/OutOfSchool.WebApi/Startup.cs
```diff
@@ -120,7 +121,30 @@ public static void AddApplicationServices(this WebApplicationBuilder builder)
                     .AllowAnyHeader()
                     .AllowCredentials()));
 
-        services.AddControllers().AddNewtonsoftJson()
+        var cacheProfilesConfig = configuration.GetSection(CacheProfilesConfig.Name).Get<CacheProfilesConfig>();
+
+        services.Configure<CacheProfilesConfig>(configuration.GetSection(CacheProfilesConfig.Name));
```
VadymLevkovskyi: (Readability, performance) It's minor, but it bugs me. 
Please extract repeated config section reading (`configuration.GetSection(CacheProfilesConfig.Name)`) into local variable and then use it in both places.

alrom: Done

VadymLevkovskyi (approved):
There is no test coverage, but it looks to be hard to achieve

PR #1321, opened 2023-12-07
Fix ExternalExportProvider

VitaliyYuras:


OutOfSchool/OutOfSchool.WebApi/Models/ProvidersInfo/CodeficatorAddressWorkshopInfoDto.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace OutOfSchool.WebApi.Models.ProvidersInfo;
+
+public class CodeficatorAddressWorkshopInfoDto
+{
+    public int Id { get; set; }
+
+    public string Category { get; set; }
```
VadymLevkovskyi: Could you explain this `string Category` prop and why it is not present in other `CodeficatorAddressProviderInfoDto`?

OutOfSchool/OutOfSchool.WebApi/Services/ExternalExportProviderService.cs
```diff
@@ -111,4 +118,40 @@ private WorkshopInfoBaseDto MapToInfoWorkshopDto(Workshop workshop)
...
+
+    private async Task FillRatingsForProviders(List<ProviderInfoBaseDto> providersDTO)
+    {
+        var providerIds = providersDTO.Select(p => p.Id).ToList();
+        var averageRatings = await averageRatingService.GetByEntityIdsAsync(providerIds).ConfigureAwait(false);
+
+        foreach (var providerDto in providersDTO)
+        {
+            if (providerDto is ProviderInfoDto providerInfoDto)
+            {
```
VadymLevkovskyi: (Code style) Instead of inner `if` you could use `.OfType<T>`. It removes extra nesting, and make your intention more clear:
`foreach (var providerInfoDto in providersDTO.OfType<ProviderInfoDto>())`

OutOfSchool/OutOfSchool.WebApi/Services/ExternalExportProviderService.cs
```diff
@@ -111,4 +118,40 @@ private WorkshopInfoBaseDto MapToInfoWorkshopDto(Workshop workshop)
...
+    private async Task FillRatingsForWorkshops(Dictionary<Guid, List<WorkshopInfoBaseDto>> workshopsDtoMap)
+    {
+        var workshopIds = workshopsDtoMap.SelectMany(kv => kv.Value.Select(w => w.Id)).ToList();
+        var averageRatings = await averageRatingService.GetByEntityIdsAsync(workshopIds).ConfigureAwait(false);
+
+        foreach (var (providerId, workshopsDto) in workshopsDtoMap)
+        {
+            foreach (var workshopDto in workshopsDto)
+            {
+                if (workshopDto is WorkshopInfoDto workshopInfoDto)
```
VadymLevkovskyi: Same `.OfType<>` approach could be used here

OutOfSchool/OutOfSchool.WebApi/Models/ProvidersInfo/AddressProviderInfoDto.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace OutOfSchool.WebApi.Models.ProvidersInfo;
+
+public class AddressProviderInfoDto
```
VadymLevkovskyi: As I see the only difference between `AddressProviderInfoDto` and `AddressWorkshopInfoDto` is `CodeficatorAddress` type. Is it expected that mentioned classes will later obtain more differences?
If not - why don't use inheritance, and have one `AddressBaseInfoDto` class without this field and other, derived class, with it?
I'm asking, because I see lot of duplicates, and don't see reasoning for duplications

OutOfSchool/OutOfSchool.WebApi/Models/ProvidersInfo/CodeficatorAddressWorkshopInfoDto.cs
DmyMi: it was my mistake during ctrl+c/v of example JSON in the task. This field is not needed in both cases

OutOfSchool/OutOfSchool.WebApi/Models/ProvidersInfo/AddressProviderInfoDto.cs
VitaliyYuras11: I've fixed this issue. Now  _Address_ and _Codeficatoraddress_  are the same for both  provider and  workshop.

VadymLevkovskyi: Thank you

VadymLevkovskyi (approved):
Approved, though there is duplication alert, but it seems to be only originated from `...InfoDto` / `...Dto` properties

PR #1324, opened 2023-12-08
Chat error fix

min4uk:
Fixed error when after new chat created, ChatRoomWorkshopDto has null workshop

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Services/ChatRoomWorkshopService.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Services/ChatRoomWorkshopService.cs
@@ -48,8 +48,8 @@ public class ChatRoomWorkshopService : IChatRoomWorkshopService
             {
                 var newChatRoom = await this.CreateAsync(workshopId, parentId).ConfigureAwait(false);
                 logger.LogDebug($"{nameof(ChatRoomWorkshop)} id:{newChatRoom.Id} was saved to DB.");
-                var chatRoomDto = await roomRepository.GetById(newChatRoom.Id).ConfigureAwait(false);
-                return mapper.Map<ChatRoomWorkshopDto>(chatRoomDto);
+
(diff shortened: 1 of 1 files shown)


DmyMi (approved):

OutOfSchool/OutOfSchool.WebApi/Services/ChatRoomWorkshopService.cs
```diff
@@ -48,8 +48,8 @@ public async Task<ChatRoomWorkshopDto> CreateOrReturnExistingAsync(Guid workshop
             {
                 var newChatRoom = await this.CreateAsync(workshopId, parentId).ConfigureAwait(false);
                 logger.LogDebug($"{nameof(ChatRoomWorkshop)} id:{newChatRoom.Id} was saved to DB.");
-                var chatRoomDto = await roomRepository.GetById(newChatRoom.Id).ConfigureAwait(false);
-                return mapper.Map<ChatRoomWorkshopDto>(chatRoomDto);
+
+                return await this.GetUniqueChatRoomAsync(newChatRoom.WorkshopId, newChatRoom.ParentId).ConfigureAwait(false);
```
VadymLevkovskyi, line 52: Just double-checking. Can't we map `newChatRoom` to necessary `...Dto` type and gain some performance? 
Asking, b/c I've checked `GetUniqueChatRoomAsync()` implementation, and it seems to use same `roomRepository` as `CreateAsync()`, just do extra map under the hood.

min4uk: In `CreateAsync` method, I use `roomRepository.Create` to create a `ChatRoomWorkshop`. The `Create` method of the repository returns a `ChatRoomWorkshop` without the `Workshop` property (null). That is why I use an additional call to the repository to get it with the included properties

VadymLevkovskyi: So repository works asymmetrically (and that's sad). Meanwhile, thanks for info, approved

VadymLevkovskyi (approved):

PR #1325, opened 2023-12-08
Alter of StatisticReportData view

min4uk:
Changed generation of the Id column in StatisticReportData view to auto-incremental to fix the problem that only the first line repetitions are saved to the report file


VadymLevkovskyi (approved):
Can't do a good review for such a ton of SQL/db changes, thus approving it blindly. 

Do we have any integration tests addressing suggested changes? (As simple as verify that query returns something meaningful, or validating a "snapshot" results for some expected input data)

min4uk:
> Do we have any integration tests addressing suggested changes? (As simple as verify that query returns something meaningful, or validating a "snapshot" results for some expected input data)

We don't have such tests. I looked at what options exist, but did not find any option for in-memory databases. If you have a suggestion/hint on how to implement this, I would be grateful.
Also, we discussed this report at the meetup and since this report does not give us the possibility to simply filter by region/territorial community without spending a lot of resources, the approach will most likely change after advice from the data analyst.

VadymLevkovskyi:
> > Do we have any integration tests ..? ...
> 
> We don't have.... I looked at what options exist, but did not find any option for in-memory databases. ....

Well, talking abstractly, there are 2 in-memory options:
- either mocked db by EF (and thus it's not SQL, but rather LINQ version, so views functionality couldn't be tested)
- or EF to SQLite (that is not our current db provider).

Ofc it's possible to adopt SQLite for testing purposes, but I'd like to know do we need it ("don't fix what ain't broken"). 
@DmyMi are db-related bugs frequent/hard to be fixed? Should we invest into autotesting this scope?

DmyMi:
@VadymLevkovskyi i don't think we should invent complex testing procedures here, as we will probably re-write this eventually and might not use the same approach.

VadymLevkovskyi (approved):

PR #1326, opened 2023-12-11
Fix total amount providers

VitaliyYuras:


OutOfSchool/OutOfSchool.WebApi/Services/ExternalExportProviderService.cs
```diff
@@ -58,9 +57,11 @@ public async Task<SearchResult<ProviderInfoBaseDto>> GetProvidersWithWorkshops(D
                 ? "Providers table is empty."
                 : $"All {providerWorkshopDtos.Count} records were successfully received from the Provider table");
 
+            int count = providerRepository.CountWithDeleted();
```
DmyMi: This is only the first part :) total amount should reflect the provided filter. Which means you need to also use the `updatedAfter` if it is present for count.

If we provide this value, we need to know only the total amount of providers that were changed or who's workshops have been changed after the given date (like you use it in `GetAllUpdatedProviders`).

DmyMi (changes requested):

VadymLevkovskyi (approved):

DmyMi (approved):

PR #1328, opened 2023-12-13
Tsylia/Admins Template pattern

andriitsylia:
It has added new admins's dtos, services, controllers, and utility classes to WebApi. This is an attempt to refactor these classes according to the Template method.


andriitsylia:
@VadymLevkovskyi, could you review this PR? It attempts to use the Template pattern to organize the admins' services in the WebAPI project. I've left the old services to compare their functionality. I will remove them, rename new ones, and start to write tests when this code improves after your reviews.

PR #1330, opened 2023-12-14
Chat fix

min4uk:
changed path of hubs so they have unified subpath
Added TODO for LocalDistributedHubLifetimeManager if chat will stop working

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Middlewares/AuthorizationTokenMiddleware.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Middlewares/AuthorizationTokenMiddleware.cs
@@ -2,7 +2,7 @@
 
 public class AuthorizationTokenMiddleware
 {
-    private static readonly string[] Hubs = new string[] { "/hubs/chat", "/hubs/notification" };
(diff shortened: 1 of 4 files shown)


VadymLevkovskyi (approved):

OutOfSchool/OutOfSchool.WebApi/Middlewares/AuthorizationTokenMiddleware.cs
```diff
@@ -2,7 +2,7 @@
 
 public class AuthorizationTokenMiddleware
 {
-    private static readonly string[] Hubs = new string[] { "/chathub", "/notificationhub" };
+    private static readonly string[] Hubs = new string[] { "/hubs/chat", "/hubs/notification" };
```
VadymLevkovskyi: (Code style) It looks like these string constants are reused. So they could be extracted from the list, made public and reused.

VadymLevkovskyi (approved):
Approved. Sonar is mumbling about low coverage and 1 issue (TODO), but (as for me) it could be ignored for this specific case.

PR #1335, opened 2023-12-20
Fix some of the codesmells

DmyMi:
Fixed some of the "simple" codesmells that do not require rewriting logic


DmyMi:
It says 0 tests because tests were out of scope of this change :) detailed view of Sonar shows PR has 0 smells though

alrom (approved):

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminService.cs
```diff
@@ -18,9 +18,8 @@ public class ProviderAdminService : IProviderAdminService
     private readonly IMapper mapper;
     private readonly ILogger<ProviderAdminService> logger;
     private readonly IProviderAdminRepository providerAdminRepository;
-    private readonly GRPCConfig gPRCConfig;
+    private readonly GrpcConfig gPrcConfig;
```
VadymLevkovskyi: (Code style) Please rename field from `gPrcConfig` to `grpcConfig`.

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminServiceGRPC.cs
```diff
@@ -6,12 +6,12 @@
 namespace GrpcServiceServer;
 
 [Authorize]
-public class ProviderAdminServiceGRPC : GRPCProviderAdmin.GRPCProviderAdminBase
+public class ProviderAdminServiceGrpc : GRPCProviderAdmin.GRPCProviderAdminBase
```
VadymLevkovskyi, line 9: (Code style) Please rename file to match class name

OutOfSchool/OutOfSchool.AuthorizationServer/KeyManagement/ExternalCertificate.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace OutOfSchool.AuthorizationServer.KeyManagement;
 
-public class ExternalCertificate
+public static class ExternalCertificate
```
VadymLevkovskyi, line 7: (Code design) Static classes are not testable/mockable. As this one introduce dependencies to file system, it's highly required to be mocked. Please consider rewriting it to "service" and injecting via usual or "poor-man" DI.

OutOfSchool/OutOfSchool.AuthorizationServer/KeyManagement/KeyManager.cs
```diff
@@ -55,7 +55,7 @@ private static async Task<X509Certificate2> HandleConcurrentUpdate(DbUpdateExcep
         {
             // Data was deleted but we didn't do it, this case should be unreal
             // throw exception and hope for the better next time
-            throw new Exception();
+            throw new InvalidOperationException();
```
VadymLevkovskyi: (Code maintainability) Please add message to the exception. Comment above could be a good source of the message text (and thus could be eliminated or transformed into //TODO: comment).

OutOfSchool/OutOfSchool.AuthorizationServer/KeyManagement/KeyManager.cs
```diff
@@ -69,7 +69,7 @@ private static async Task<X509Certificate2> HandleConcurrentUpdate(DbUpdateExcep
 
         // Updated and expired right away, this case should be unreal
         // throw exception and hope for the better next time?
-        throw new Exception();
+        throw new InvalidOperationException();
```
VadymLevkovskyi: (Code maintainability, similar to above) Please add message to the exception. Comment above could be a good source of the message text (and thus could be eliminated or transformed into //TODO: comment).

OutOfSchool/OutOfSchool.Redis/CacheService.cs
```diff
@@ -18,6 +18,8 @@ public class CacheService : ICacheService, IDisposable
     private readonly bool isEnabled = false;
 
     private readonly object lockObject = new object();
+    
+    private bool _disposed;
```
VadymLevkovskyi: (Code style) Please be consistent with private fields naming. All other fields don't have underscore prefix, so could do `_disposed`. And, IMO, `isDisposed` is even better name for the flag.

OutOfSchool/OutOfSchool.WebApi/Hubs/LocalDistributedHubLifetimeManager.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace OutOfSchool.WebApi.Hubs;
 
-public class LocalDistributedHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposable
+public sealed class LocalDistributedHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposable
```
VadymLevkovskyi, line 9: (Curiosity) I wonder why precisely this class is made `sealed`? No objections, just asking

VadymLevkovskyi (changes requested):
Thanks for so many changes. No objections re logic, there are few comments mainly re code style. 

And there is a big question about coverage. I know, that logic wasn't changed, but, you know, the time has come to write tests )

OutOfSchool/OutOfSchool.WebApi/Hubs/LocalDistributedHubLifetimeManager.cs
DmyMi: @VadymLevkovskyi  for the previous class (`CacheService`) we may have an inheriting class sometime, so i followed a recommended approach for a Disposable pattern there. For this class there will be no inheritance, so adding `sealed` makes it compliant to "implement Disposable pattern correctly i case someone will make a child class" warning in Sonar.

VadymLevkovskyi: Ah, `Disposable` pattern. Thanks!

OutOfSchool/OutOfSchool.AuthorizationServer/KeyManagement/ExternalCertificate.cs
DmyMi: @VadymLevkovskyi  this code is used only in Startup as an util method (single method call) to provide correctly formatted PEM certificate to OpenIdDict. It uses only static methods of `Path`, `File`, etc.

So to test it we need to run integration tests, e.g., we load sample certificate from filesystem, put it in OpenIdDict and check if it works. If it is needed, i will try to write such test

OutOfSchool/OutOfSchool.AuthCommon/Services/ProviderAdminServiceGRPC.cs
DmyMi: file renamed

OutOfSchool/OutOfSchool.AuthorizationServer/KeyManagement/ExternalCertificate.cs
VadymLevkovskyi: Well, I was worriing about usage of this code somewhere in services, and then having no possibility to unit-test that services. So, seems it is an exception, and there is no need to make this specific class "not static". Sorry for bothering.

OutOfSchool/OutOfSchool.AuthServer.Tests/Services/ProviderAdminServiceGrpcTests.cs
```diff
@@ -0,0 +1,131 @@
...
+        // Act
+        var result = await providerAdminServiceGrpc.CreateProviderAdmin(request, serverCallContextMock);
+
+        // Assert
+        result.Should().NotBeNull();
+        result.IsSuccess.Should().BeTrue();
+    }
+
+    [Test]
+    public async Task CreateProviderAdmin_WhenEntityInalid_ShouldReturnSuccessFalse()
```
min4uk: `_WhenEntityIn v alid_`
Missing letter v in word Invalid

DmyMi:
@VadymLevkovskyi There are a lot of uncovered lines in `OutOfSchool.AuthCommon/Services/ProviderAdminService.cs`, but it is completely re-written in another PR so there is no point writing tests here to maximise coverage :(
Added tests for `CacheService`, `FormValueRequiredAttribute` and `ProviderAdminServiceGrpc`.

VadymLevkovskyi (approved):

PR #1336, opened 2023-12-21
Partly removed duplicates (mapping profiles as example)

VadymLevkovskyi:
This PR tries to address duplicates and wrong code placement issues. 
As a playground mapping profiles area was chosen.

Key takeaways:
- make your models "smart" - logic should reside closer to the model or in extension classes, not in usage sites, because for latter case you'll write same code again and again (see CATOTTG mapping, or usage of `User.LastLogin` and `AccountStatus`);
- when some logic is span over several models - marker interfaces could help a lot (`IHasRatings` ...);
- mapping profiles could be writen in more elgant way using mapping rules from base class, helper methods, etc.

Suggested review process:
- To get a main idea of suggested changes - please check the MappingProfile.cs changes.
- To look through changes step-by-step - it's better to review PR by separate commit.

Notes:
- Not all possible "deduplication" changes were introduced, as PR is already slightly big.
- (Tests for uncovered code will be written here soon) Irrelevant, as it's already covered

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs
@@ -1,10 +1,9 @@
 using AutoMapper;
 using Google.Protobuf.WellKnownTypes;
 using GrpcService;
-using OutOfSchool.Common.Enums;
 using OutOfSchool.Common.Models;
 using OutOfSchool.Services.Enums;
-using OutOfSchool.WebApi.Enums;
+using OutOfSchool.Services.Models.Images;
 using OutOfSchool.WebApi.Models;
 using OutOfSchool.WebApi.Models.Achievement;
 using OutOfSchool.WebApi.Models.Application;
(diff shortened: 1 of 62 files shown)


OutOfSchool/Tests/OutOfSchool.WebApi.IntegrationTests/OutOfSchool.WebApi.IntegrationTests.csproj
```diff
@@ -27,6 +27,7 @@
     </ItemGroup>
 
     <ItemGroup>
+        <ProjectReference Include="..\..\OutOfSchool.AuthorizationServer\OutOfSchool.AuthorizationServer.csproj" />
```
min4uk: I don't see any new tests/code that would require this reference, what is it for?

VadymLevkovskyi: Yes, that's wrong dependency. Removed

OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs
```diff
@@ -779,9 +625,43 @@ public MappingProfile()
...
\ No newline at end of file
+
+    private IMappingExpression<Provider, T> AddCommonProvider2ProviderBaseDto<T>(IMappingExpression<Provider, T> mappings)
+        where T : ProviderBaseDto
+        => mappings
+            .ForMember(dest => dest.ActualAddress, opt => opt.MapFrom(src => src.ActualAddress))
+            .ForMember(dest => dest.LegalAddress, opt => opt.MapFrom(src => src.LegalAddress))
+            .ForMember(dest => dest.Institution, opt => opt.MapFrom(src => src.Institution))
+            .Apply(IgnoreAllImages)
+            .ForMember(dest => dest.ImageIds, opt => opt.MapFrom(src => src.Images.Select(x => x.ExternalStorageId)))
```
min4uk: Can we use here ` MapImageIds` method?

min4uk (approved):

min4uk (approved):

OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs
VadymLevkovskyi: Thanks for pointing this, as a result one more issue fixed (related to initially getter-only destination property described by interface)

PR #1338, opened 2023-12-21
Creation email error responses

Tfront1:


OutOfSchool/OutOfSchool.WebApi/Services/AreaAdminService.cs
```diff
@@ -103,7 +104,14 @@ public async Task<AreaAdminDto> GetByUserId(string id)
...
                 "AreaAdmin creating is not possible. Username {Email} is already taken",
                 areaAdminBaseDto.Email);
-            throw new InvalidOperationException($"Username {areaAdminBaseDto.Email} is already taken.");
+            return new ErrorResponse
+            {
+                HttpStatusCode = HttpStatusCode.BadRequest,
+                ApiErrorResponse = new ApiErrorResponse(new List<ApiError>()
+                {
+                    ApiErrorsTypes.Common.EmailAlreadyTaken("AreaAdmin", areaAdminBaseDto.Email),
+                }),
```
VadymLevkovskyi: (Code duplication, suggestion) It looks like we have a repetitive pattern:
```
    ... new ApiErrorResponse(new List<ApiError>()
    {
        .../*construct single specific error*/... ,
    })
```
What if extract it to some extension method (i.e. `ToResponse()`) that could be used like
```
  return new ErrorResponse
  {
      HttpStatusCode = HttpStatusCode.BadRequest,
      ApiErrorResponse = ApiErrorsTypes.Common.EmailAlreadyTaken("AreaAdmin", areaAdminBaseDto.Email).ToResponse()
  }
```

Additionally another pattern is pretty common
```
    return new ErrorResponse
    {
        HttpStatusCode = HttpStatusCode.BadRequest,
        ApiErrorResponse = .../*some ApiErrorResponse*/...,
    };
```
Couldn't it be extracted as some "builder" method of `ErrorResponse`, i.e.
```
    return ErrorResponse.BadRequest( .../*some ApiErrorResponse*/...);
```

OutOfSchool/OutOfSchool.WebApi/Services/ProviderAdminService.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System.Linq.Expressions;
 using AutoMapper;
+using Castle.Core.Internal;
```
VadymLevkovskyi: (Code maintainability) Is this using required?

VadymLevkovskyi (changes requested):
(Test coverage) It is now 4.2%. Please improve our test suit to reach expected >80% goal.

OutOfSchool/OutOfSchool.WebApi/Services/AreaAdminService.cs
VadymLevkovskyi: _(Please don't resolve conversations I've started. I do it after my review. If you need to mark that you've done something you could either "react" on my message or write the comment below)_

Ok, thanks you for implementation of second idea from previous comment (`ErrorResponse.BadRequest()`), but what about first one? 
Hav you disliked it or skipped it or ...? 

Asking this, because I see same pattern in your tests `SetUp()`'s

Tfront1: I removed this pattern from SetUp()'s.

As for the first pattern, I changed it to look like this:

-"Creating list apiErrors"
-"Filling it with errors as needed"
-"Return errors, if any"

I made this decision because in the future the errors will not only apply to entering already existing mail when creating. There will still be such errors for phone numbers, and maybe something else.

I understood the solution with ToResponse() like this:
There is a scoped service to which we add errors during operation using the ToResponse() method. Such solutions have already been implemented by me, but you said that it was not ok, due to heavy testing.

Btw, sorry for resolving conversations.

Tfront1: @VadymLevkovskyi

VadymLevkovskyi: > Btw, sorry for resolving conversations.

Np, it was just attempt to align communication.

Thank you for your changes, and your explanation.

VadymLevkovskyi (approved):

Issue #1316, opened 2023-12-06
[Admin OTG] There is no information about the new service provider on the "Історія змін" page.


ViktoriiaHrytchyna:
**Environment:** Windows 10, 64-bit, Google Chrome 107.0.5304.107.
**Reproducible:** always.

**Preconditions**
1. Go to [https://oos.dmytrominochkin.cloud/#](https://oos.dmytrominochkin.cloud/).
2. Register a new service provider.

**Steps to reproduce**
1. Log in as Admin OTG ("Email" = '[jevahaj880@synclane.com](mailto:jevahaj880@synclane.com)', "password" = '?_7jrRnL')
2. Go to "Адміністрування" on the top right corner drop down list.
3. Click on "Історія змін" tab.
4. Pay attention to the lack of information about the new service provider in the table.

**Actual result**
Information about the registration of a new service provider is not displayed on the "Історія змін" page for the OTG administrator.
![image](https://github.com/ita-social-projects/OoS-Backend/assets/115226563/29f82caa-06fb-46d8-b9f8-0d5fd234be64)
![image](https://github.com/ita-social-projects/OoS-Backend/assets/115226563/e2b43bb5-e86f-40a2-8cc6-023dad234c45)

**Expected result**
If a new service provider has registered, the OTG administrator should see information about him on the "Історія змін" page.

**Labels to be added**
"Bug", Priority ("pri: medium"), Severity ("severity: medium"), Type ("Functional"), "API" (for back-end bugs).

PR #1341, opened 2023-12-22
Romanchuk / Add changes log to Provider creating

alrom:
#1316

### DIFF
--- a/OutOfSchool/OutOfSchool.DataAccess/Repository/ChangesLogRepository.cs
+++ b/OutOfSchool/OutOfSchool.DataAccess/Repository/ChangesLogRepository.cs
@@ -71,7 +71,7 @@ public class ChangesLogRepository : EntityRepository<long, ChangesLog>, IChanges
         return result;
     }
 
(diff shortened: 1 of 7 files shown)


OutOfSchool/OutOfSchool.DataAccess/Repository/ChangesLogRepository.cs
```diff
@@ -68,6 +71,27 @@ public ChangesLogRepository(OutOfSchoolDbContext dbContext)
         return result;
     }
 
+    public async Task<ChangesLog> AddCreatingOfEntityToChangesLog<TEntity>(
```
VadymLevkovskyi: (Code style, performance) `async`/`await` seems to be excessive for this method, as have only one `await` in method body that is immediately returned. Please refactor

alrom: Done. Thanks.

VadymLevkovskyi (approved):

PR #1342, opened 2023-12-22
Changed healths endpoint port

olsavchenk:


VadymLevkovskyi (commented):
I'm curious why approach suggested in [SO discussion](https://stackoverflow.com/questions/57721900/healthchecks-on-different-port-in-net-core-web-application) doesn't work

OutOfSchool/OutOfSchool.WebApi/Startup.cs
```diff
@@ -93,6 +91,7 @@ public static void Configure(this WebApplication app)
                 Predicate = healthCheck => healthCheck.Tags.Contains("readiness"),
                 AllowCachingResponses = false,
             })
+            .RequireHost("*:9000")
```
VadymLevkovskyi: (Code style) Magic number, that probably has to be `ApplicationPorts:HealthPort`.

olsavchenk: Somehow missed that, it's fixed now

VadymLevkovskyi (approved):
Approved, though suggested middleware seems excessive
At same time I have no more resources than mentioned SO discussion to prove that it could be simplified, so let it be.

PR #1346, opened 2024-01-04
Fixed workshop creation issue with avaliable seats

### LINKED ISSUE #1339: [Service Provider,Workshops] The number of participants is displayed as 0/0 after creating a new workshop on the My Workshops page.
**Environment:** Windows 10, 64-bit, Google Chrome 118.0.5993.70,
Windows 10, 64-bit, Microsoft Edge 114.0.1823.67.
**Reproducible:** always.

**Preconditions**

1. Go to https://oos.dmytrominochkin.cloud/;
2. Log in as a State Service provider;

**Steps to reproduce**
1. Go to the "My workshops" page.
2. Click on the "Add workshop" button.
3. Fill in all required fields on all pages and create a workshop.
4. Pay attention to the number of participants that is displayed in the new workshop.

**Actual result**
The number of participants in the new workshop is displayed as 0/0.
image

**Expected result**
In the new workshop of the public service provider, the number of participants is displayed as 0 after creation.

**Labels to be added**
"Bug", Priority ("pri: medium "), Severity ("severity: medium "), Type "Functional").

olsavchenk:
Fixed #1339

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/WorkshopServiceTests.cs
+++ b/OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/WorkshopServiceTests.cs
@@ -672,14 +672,46 @@ public class WorkshopServiceTests
             .Returns(Task.FromResult(provider));
         workshopRepository.Setup(
                 w => w.Create(It.IsAny<Workshop>()))
-            .Returns(Task.FromResult(It.IsAny<Workshop>()));
+            .ReturnsAsync((Workshop workshop) =>
+            {
+                return new Workshop
+                {
+                    Id = new Guid("8f91783d-a68f-41fa-9ded-d879f187a94e"),
+                    AvailableSeats = workshop.AvailableSeats,
+                };
+            });
         workshopRepository.Setup(
                 w => w.RunInTransaction(It.IsAny<Func<Task<Workshop>>>()))
-            .ReturnsAsync(new Workshop() { Id = new Guid("8f91783d-a68f-41fa-9ded-d879f187a94e") });
+            .ReturnsAsync((Func<Task<Workshop>> func) =>
+            {
+                return new Workshop
+                {
+                    Id = new Guid("8f91783d-a68f-41fa-9ded-d879f187a94e"),
+                    AvailableSeats = func.Invoke().Result.AvailableSeats,
+                };
+            });
         mapperMock.Setup(m => m.Map<WorkshopBaseDto>(It.IsAny<Workshop>()))
-            .Returns(new WorkshopBaseDto() { Id = new Guid("8f91783d-a68f-41fa-9ded-d879f187a94e") });
+            .Returns((Workshop workshop) =>
+            {
+                var dto = new WorkshopBaseDto
+                {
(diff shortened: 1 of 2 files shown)


OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
```diff
@@ -92,7 +92,7 @@ public async Task<WorkshopBaseDto> Create(WorkshopBaseDto dto)
 
         if (dto.AvailableSeats is 0 or null)
         {
-            dto.AvailableSeats = uint.MaxValue;
+            workshop.AvailableSeats = uint.MaxValue;
```
VadymLevkovskyi: (Tests) Test coverage is 100%, but I don't see any test changed/failed while logic was changed. Could you check and write missing test if necessary for this case?

olsavchenk: I've fixed tests, but I'm not sure is it ok to add same guid in every method mock or should I do same as with avaliable seats?

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/WorkshopServiceTests.cs
```diff
@@ -672,14 +672,46 @@ private void SetupCreate()
...
+            });
         workshopRepository.Setup(
                 w => w.RunInTransaction(It.IsAny<Func<Task<Workshop>>>()))
-            .ReturnsAsync(new Workshop() { Id = new Guid("8f91783d-a68f-41fa-9ded-d879f187a94e") });
+            .ReturnsAsync((Func<Task<Workshop>> func) =>
+            {
+                return new Workshop
+                {
+                    Id = new Guid("8f91783d-a68f-41fa-9ded-d879f187a94e"),
+                    AvailableSeats = func.Invoke().Result.AvailableSeats,
```
VadymLevkovskyi: (Code readability)  
1. I believe that `func.Invoke()` could be rewritten to just `func()`.
2. Not sure, but probably ssync part with `.Result` could be better highlighted using `.Returns(async ` vs your current `.ReturnsAsync( .... /* and lot later we have */ .Result. ....)`, see below:
```
... .Returns(async (Func<Task<Workshop>> func) =>
{
    var workshop = await func();
    return new Workshop
    {
        Id = new Guid("8f91783d-a68f-41fa-9ded-d879f187a94e"),
        AvailableSeats = workshop.AvailableSeats; 
...
```
In any case please avoid usage of `.Result`

OutOfSchool/OutOfSchool.WebApi/Services/Database/WorkshopService.cs
VadymLevkovskyi: 1. (Readability) If you really need same `GUID` - please introduce it as local variable before and use it in setup, don't recreate same each time. This will make reuse of same `id` more obvious (without keeping an eagle eye for few lines comparing 20+ symbols) 
2. (The logic) Honestly I don't get if `id` should be same - is any test failing because of wrong `id`?

VadymLevkovskyi (approved):

PR #1347, opened 2024-01-04
Fix getting TakenSeats for workshop

alrom:


OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs
```diff
@@ -285,11 +285,11 @@ public MappingProfile()
             .IncludeBase<Workshop, WorkshopBaseCard>()
             .ForMember(dest => dest.AmountOfPendingApplications, opt => opt.MapFrom(src =>
                 src.Applications.Count(x =>
-                    x.Status == ApplicationStatus.Pending && !x.IsDeleted)))
+                    x.Status == ApplicationStatus.Pending && !x.IsDeleted && (x.Child == null || !x.Child.IsDeleted) && (x.Parent == null || !x.Parent.IsDeleted))))
```
VadymLevkovskyi: (Tests) Changes in logic are not covered by tests. (I see changes in test data generators, but no related success/fail tests). 

As a suggestion - you could extract `AmountOfPendingApplications` and `TakenSeats` logic into separate methods under `WorkshopProviderViewCard` or extension for `Application`. Thus extracted methods are more testable and mapper wouldn't contain the logic.

alrom:
@VadymLevkovskyi I added tests and extracted code to extensions.

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/ApplicationExtensionTests.cs
```diff
@@ -0,0 +1,108 @@
...
+        // Arrange
+        var applications = GetApplications();
+
+        // Act
+        var result = applications.AmountOfPendingApplications();
+        var expected = applications.Count(x =>
+            x.Status == ApplicationStatus.Pending
+            && !x.IsDeleted
+            && (x.Child == null || !x.Child.IsDeleted)
+            && (x.Parent == null || !x.Parent.IsDeleted));
```
VadymLevkovskyi: (Fragile tests, duplicated code) Please avoid testing method's code by writing the same code in test. This way you are not checking the requirements, but rather your understanding of them (and if you had taken/coded them wrong once - you wouldn't have got any chance to catch this by unit-test). 

Instead check the requirements. Separately. Because it's a `unit`-test. Not a `method`-test.

Assert, that `deleted` applications doesn't affect the count (1 test), applications in other status than `Pending` doesn't affect (another test), check how `not-`/`null` `not-`/`deleted` `child` affects the count, same for `parent`. 
(Yes, it would be many, but pretty simple (dumb) tests.) 

So, in each test you could start with a list of ideal applications, make N applications "wrong" due to one of the requirement (or their combination) and ensure that count returned by `AmountOfPendingApplications()` is changed by N.

Summing up: the idea is that your test`s code should be simplier than code of your logic, thus you could do less mistakes in unit-test, it is easier to maintain, failed test has only one reason to fail ... y'know, all this FIRST principles.

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/ApplicationExtensionTests.cs
```diff
@@ -0,0 +1,108 @@
...
+            (x.Status == ApplicationStatus.Approved || x.Status == ApplicationStatus.StudyingForYears)
+            && !x.IsDeleted
+            && (x.Child == null || !x.Child.IsDeleted)
+            && (x.Parent == null || !x.Parent.IsDeleted));
+
+        // Assert
+        Assert.AreEqual(expected, result);
+    }
+
+    private List<Application> GetApplications()
```
VadymLevkovskyi: (Pitfalls of separating logic from the test)
It's common approach to reuse the common code by extracting it to helper methods, even in testing, but this method has more than one responsibility and hides the complexity.

I.e., it looks like `child == null` or `parent == null` conditions are never addressed by the generated data, so never asserted in the related tests. Please reconsider and (maybe) refactor it.

OutOfSchool/OutOfSchool.WebApi/Util/MappingProfile.cs
VadymLevkovskyi: Thanks for the code changes. Looks better, though there are some concerns, see below comments.

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/ApplicationExtensionTests.cs
alrom: Thanks. I understood. I made short tests for different situations.

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/ApplicationExtensionTests.cs
alrom: Deleted it.

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/ApplicationExtensionTests.cs
alrom: @VadymLevkovskyi I missed your nick in the comment before.  Sorry.

VadymLevkovskyi: Np, you don't need to mention me. I get notifications on any changes in PR where I'm set as a reviewer.
And thanks for redesign!

VadymLevkovskyi (approved):

PR #1349, opened 2024-01-06
Fix sorting for GetFilteredProviderAdmins

alrom:


VadymLevkovskyi (commented):
(Tests) Sorting/paging logic was changed - so there should be a test checking it? Could you show me any (or write new one?)

alrom:
@VadymLevkovskyi I added test.

VadymLevkovskyi (approved):

PR #1350, opened 2024-01-08
Fixed issue with "StudyingForYears" application status

olsavchenk:


olsavchenk:
Some of the changes are not connected to my issue I've found them during my investigation and I decided to fix them. In InstitutionHierarchyServiceTest there was unused mock and in ChildControllerTests there were extra tabs

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ChildServiceTests.cs
```diff
@@ -1,178 +1,286 @@
...
+            parentRepositoryMock.Object,
+            socialGroupRepositoryMock.Object,
+            loggerMock.Object,
+            mapperMock.Object,
+            applicationRepositoryMock.Object,
+            parentConfigMock.Object);
+    }
+
+    [Test]
+    public async Task GetApprovedByWorkshopId_ExistingChildren_ShouldReturnChildren()
```
VadymLevkovskyi, line 63: (Test logic) It looks that this test would never be "red", as it checks something, but not you code changes. 

Try to remove from your `ChildService` added piece of code `|| p.Status == ApplicationStatus.StudyingForYears` and this test will not fail.

VadymLevkovskyi (changes requested):
Please improve your tests

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ChildServiceTests.cs
olsavchenk: Hopefully, I have fixed the tests, and now they will fail when the filter in ChildService is changed

VadymLevkovskyi (approved):

PR #1351, opened 2024-01-08
Tsylia/Fixed the workshop's updating

andriitsylia:


VadymLevkovskyi (commented):
(Tests) Could you show what test is checking these changes in functionality or write the new one?

andriitsylia:
We already have a test method called `Update_WhenEntityIsValid_ShouldReturnUpdatedEntity([Random(2, 5, 1)] int teachersInWorkshop)` in the `OutOfSchool\OutOfSchool.WebApi.Tests\Services\Database\WorkshopServiceTests.cs`.

VadymLevkovskyi:
> We already have a test method called `Update_WhenEntityIsValid_ShouldReturnUpdatedEntity`

@andriitsylia I've read this test and see no assert related to `.Where(x => !x.IsDeleted)` condition you've added in this PR. 

Am I missing something, or should new test still be written checking entities having both `IsDeleted` set to `true` and `false` (the test that will fail if you remove mentioned `.Where()`)?

VadymLevkovskyi (approved):

PR #1354, opened 2024-01-10
Semkiv/chat rooms list for parent fix

min4uk:


OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ChatWorkshopControllerTests.cs
```diff
@@ -0,0 +1,177 @@
...
+            .Returns(Task.FromResult(expectedSearchResult));
+
+        // Act
+        var result = await controller
+            .GetParentsRoomsAsync(new ChatWorkshopFilter())
+            .ConfigureAwait(false) as OkObjectResult;
+
+        // Assert
+        Assert.That(result, Is.Not.Null);
+        Assert.AreEqual(Ok, result.StatusCode);
```
VadymLevkovskyi, line 175: (Naming or logic issue) Test is named `_ShouldReturnBadRequest`, but asserts that result is `OK`.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/ChatRoomWorkshopServiceWithDBTests.cs
```diff
@@ -0,0 +1,321 @@
...
+                    LastName = item.Parent.User.LastName,
+                    Gender = item.Parent.Gender,
+                    Email = item.Parent.User.Email,
+                    PhoneNumber = item.Parent.User.PhoneNumber,
+                },
+            });
+
+        var expectedResultEntities = expectedRooms
+            .Select(r => mapper.Map<ChatRoomWorkshopDtoWithLastMessage>(r))
+            .ToList();
```
VadymLevkovskyi: (Not a call to action right now, just asking for you to reconsider test writing approach) 
You are doubling the method logic in your test. It makes test very fragile to any changes in less relaetd code than test suppose to check - i.e. changes in mapper. 

As for me, for this specific test we could assert that `GetChatRoomByFilter` returns non-empty result (as we have prepared data for this) and that all returned rooms has expected parents id. 

Ofc, other tests could be rewritten as well.

Again, I'm not insisting on any changes right now, just think about it, please

VadymLevkovskyi (commented):
Logic and test seems good, there is only one minor gap and small consideration. See comment below

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ChatWorkshopControllerTests.cs
min4uk: Thanks for pointing it out
I confused myself
I wanted to test what would happen if I call parent method as a provider, but the controller attribute doesn't work when i call method that way, and then I decided to leave this test as positive for the admin provider and forgot to change the name...

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/ChatRoomWorkshopServiceWithDBTests.cs
min4uk: Thanks, that's really better. I will change it as you suggested

VadymLevkovskyi (approved):

PR #1361, opened 2024-01-16
Adding isBlocked

min4uk:
Added IsBlockedByProvider property to ChatRoomWorkshopDto
Moved IsBlocked property from ParentInfoForChatList to ChatRoomWorkshopForChatList and renamed to IsBlockedByProvider to avoid misunderstanding because when parent isBlocked it means that user is blocked for service ( as it is in ParentDtoWithContactInfo ) and when user is blocked by provider, then ChatRoomWorkshop IsBlocked(now IsBlockedByProvider) becomes true. So chat IsBlocked(now IsBlockedByProvider) is for blocking by provider, and parent IsBlocked is for blocking by admins
Also renamed same property for aplications for same reason

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/BlockedProviderParentRepositoryTests.cs
+++ b/OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/BlockedProviderParentRepositoryTests.cs
@@ -19,7 +19,7 @@ public class BlockedProviderParentRepositoryTests
     private List<Parent> parents;
     private List<Provider> providers;
     private List<Workshop> workshops;
-    private List<Application> aplications;
+    private List<Application> applications;
     private List<ChatRoomWorkshop> chatRooms;
 
     [SetUp]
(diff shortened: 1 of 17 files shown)


OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/BlockedProviderParentRepositoryTests.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        // Assert
+        Assert.AreEqual(expectedBlockedProviderParentCount, context.BlockedProviderParents.Count());
+        Assert.That(result is not null);
+        Assert.AreEqual(result.Id, blockedProviderParentToUnblock.Id);
+        Assert.That(result.DateTimeTo is not null);
+    }
+
+    [Test]
+    public async Task Block_ShouldBlockApplicaionAndChatRoomWorkshop()
```
VadymLevkovskyi: Please fix mistype in test names (this and next test) `...Applicaion...`.

min4uk: TY
fixed it

VadymLevkovskyi: Wow, you've found even more places than I had ). Thanks for thorough inspection!

VadymLevkovskyi (approved):

PR #1370, opened 2024-01-19
Add FormOfLearning to workshop

alrom:


alrom:
@VadymLevkovskyi missed test coverage is about Elastic.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/WorkshopServiceDBTests.cs
```diff
@@ -0,0 +1,191 @@
...
+
+        // Act
+        var result = await workshopService.GetByFilter(filter).ConfigureAwait(false);
+
+        // Assert
+        Assert.AreEqual(0, result.TotalAmount);
+    }
+
+    [Test]
+    public async Task GetByFilter_WhenSetFormOfLearning_ReturnAll()
```
VadymLevkovskyi: (Test naming) Probably this test should be named `...WhenNotSetFormOfLearning...`?

VadymLevkovskyi (commented):
One small issue found - unclear test name.

Also see nother "issue" (not sure what requirements you have). As for me "mixed" form of learning should match both "online" and "offline" filter, but now it matches only "mixed". 
Was it requested so in requirements? 
Do you think it should be reconsidered?

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/WorkshopServiceDBTests.cs
alrom: Thanks. Renamed.

VadymLevkovskyi: Yes it was, and no it doesn't meet expectations ) 
And sorry, I wasn't specific enough. `GetByFilter_WhenNotSetFormOfLearning_ReturnAll` was the proposed name, not the `GetByFilter_WhenSetFormOfLearning_WhenNotSetFormOfLearning`

Please, be more gentle to you names (and subsequent readers)

alrom:
> One small issue found - unclear test name.
> 
> Also see nother "issue" (not sure what requirements you have). As for me "mixed" form of learning should match both "online" and "offline" filter, but now it matches only "mixed". Was it requested so in requirements? Do you think it should be reconsidered?

"Mixed form" was in requirements for this task. It is the separate type of FormOfLearning for Workshop. In the filter they can send us a list of FormOfLearning if they need it.

VadymLevkovskyi:
> > One small issue found - unclear test name.
> > Also see nother "issue" (not sure what requirements you have). As for me "mixed" form of learning should match both "online" and "offline" filter, but now it matches only "mixed". Was it requested so in requirements? Do you think it should be reconsidered?
> 
> "Mixed form" was in requirements for this task. It is the separate type of FormOfLearning for Workshop. In the filter they can send us a list of FormOfLearning if they need it.

Ok, if it is separate form, describing activities that have some classes online, while others offline - let it be. 
I'd call it not a good naming/UX, but requirements are set, so I give up. 

Thanks for clarification!

alrom:
> > > One small issue found - unclear test name.
> > > Also see nother "issue" (not sure what requirements you have). As for me "mixed" form of learning should match both "online" and "offline" filter, but now it matches only "mixed". Was it requested so in requirements? Do you think it should be reconsidered?
> > 
> > 
> > "Mixed form" was in requirements for this task. It is the separate type of FormOfLearning for Workshop. In the filter they can send us a list of FormOfLearning if they need it.
> 
> Ok, if it is separate form, describing activities that have some classes online, while others offline - let it be. I'd call it not a good naming/UX, but requirements are set, so I give up.
> 
> Thanks for clarification!

It could be: Monday and Wednesday - online, other days - offline.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/WorkshopServiceDBTests.cs
alrom: Ok. Sorry. Renamed.

VadymLevkovskyi (approved):

PR #1372, opened 2024-01-25
Kymachynskyi/Add notification for parent when he was blocked by service provider

DKymachynskyi:


OutOfSchool/OutOfSchool.WebApi.Tests/Services/BlockedProviderParentServiceTests.cs
```diff
@@ -0,0 +1,228 @@
...
+            Id = Guid.NewGuid(),
+            ParentId = unblockDto.ParentId,
+            ProviderId = unblockDto.ProviderId,
+        };
+    }
+
+    #region Block
+
+    [Test]
+    public async Task Block_WhenDtoAndUserIdIsValid_ShouldSendNotificationAndReturnSuccess()
```
VadymLevkovskyi, line 88: (Test readability) Could some if `It.IsAny` be replaced by real data? 
I expect that some simple data like `Guid`, `string`, are known and could be checked during this test. 
This also would help to set clear expectations.

Same for other tests

VadymLevkovskyi (approved):

OutOfSchool/OutOfSchool.WebApi.Tests/Services/BlockedProviderParentServiceTests.cs
DKymachynskyi: Thank you! I replaced them.

VadymLevkovskyi (approved):

PR #1379, opened 2024-02-02
Kymachynskyi/ Add sorting for notifications and read all user's notifications endpoint

DKymachynskyi:


DKymachynskyi:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=ita-social-projects-oos-backend&pullRequest=1379) **Quality Gate failed**
> Failed conditions  [78.6% Coverage on New Code](https://sonarcloud.io/component_measures?id=ita-social-projects-oos-backend&pullRequest=1379&metric=new_coverage&view=list) (required ≥ 80%)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=ita-social-projects-oos-backend&pullRequest=1379)

@VadymLevkovskyi, the code in this PR has 78.6% coverage due to the new method `SetReadDateTimeForAllUnreaded` in `NotificationRepository`, which temporarily uses `ExecuteSqlRawAsync` to retrieve necessary data from DB in one query. I think, it can't be tested under current conditions because `ExecuteSqlRawAsync` is not supported by InMemory DB, which we are using for testing purposes.
After the solution is migrated to EF Core 8, it will be changed to bulk update and necessary tests will be added.
Could you review this PR?

OutOfSchool/OutOfSchool.WebApi.Tests/Services/NotificationServiceTests.cs
```diff
@@ -0,0 +1,270 @@
...
+        // Act
+        await notificationService.ReadAll(userId).ConfigureAwait(false);
+
+        // Assert
+        notificationRepositoryMock
+            .Verify(
+                x => x.SetReadDateTimeForAllUnreaded(
+                userId,
+                It.IsAny<DateTimeOffset>()),
+                Times.Once());
```
VadymLevkovskyi: (Test code style) You could simplify this test code (and remove duplications) using expected `userId` value (instead of `It.IsAny<string>()`) in `mock.Setup()`, and using parameterless `mock.VerifyAll()` (instead of specific `.Verify()`). 

Interesting fact is that you've used this approach in the subsequent tests, but not here.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/NotificationServiceTests.cs
```diff
@@ -0,0 +1,270 @@
...
+    private List<Notification> notifications;
+
+    [SetUp]
+    public void SetUp()
+    {
+        notificationRepositoryMock = new Mock<INotificationRepository>();
+        loggerMock = new Mock<ILogger<NotificationService>>();
+        localizerMock = new Mock<IStringLocalizer<SharedResource>>();
+        mapper = TestHelper.CreateMapperInstanceOfProfileTypes<CommonProfile, MappingProfile>();
+        notificationHubMock = new Mock<IHubContext<NotificationHub>>();
```
VadymLevkovskyi: (Redundant fields) Some dependencies used in `notificationService` - `loggerMock`, `localizerMock`, and `notificationHubMock` are not touched by tests, so could be easily dropped from fields list and even inlined in constructor. 

Having the full list of fields (as it is now) adds extra lines of code with no benefits, as set of really used mocks is more obscure. 

Ofc, one might say that this is done for the future tests writing, but it's too easy to extract a variable/field using VS refactoring, so it could really be done when needed, and there is no necessity for premature "optimisation".

OutOfSchool/OutOfSchool.WebApi.Tests/Services/NotificationServiceTests.cs
```diff
@@ -0,0 +1,270 @@
...
+
+        // Assert
+        notificationRepositoryMock.VerifyAll();
+        TestHelper.AssertTwoCollectionsEqualByValues(expectedNotifications, result.Notifications);
+        TestHelper.AssertTwoCollectionsEqualByValues(
+            expectedGroupedByType.Select(x => x.Amount),
+            result.NotificationsGroupedByType.Select(x => x.Amount));
+        Assert.AreEqual(
+            expectedGroupedByType.Select(x => x.GroupedByAdditionalData).Count(),
+            result.NotificationsGroupedByType.Select(x => x.GroupedByAdditionalData).Count());
```
VadymLevkovskyi: (Excessive code) `collection.Select(...).Count()` could be simplified to `collection.Count()`

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/NotificationControllerTests.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace OutOfSchool.WebApi.Tests.Controllers;
+
+[TestFixture]
+public class NotificationControllerTests
+{
+    private NotificationController controller;
+    private Mock<INotificationService> serviceMock;
+    private Mock<HttpContext> httpContext;
+    private string userId;
```
VadymLevkovskyi: (Redundant fields) Only `controller` is used in tests. Other fields are scoped to `SetUp` and could be moved to local variables there or even inlined

OutOfSchool/OutOfSchool.WebApi.Tests/Services/NotificationServiceTests.cs
DKymachynskyi: > (Test code style) You could simplify this test code (and remove duplications) using expected `userId` value (instead of `It.IsAny<string>()`) in `mock.Setup()`, and using parameterless `mock.VerifyAll()` (instead of specific `.Verify()`).
> 
> Interesting fact is that you've used this approach in the subsequent tests, but not here.

I used `VerifyAll` in this test but removed it before committing. Thank you!

OutOfSchool/OutOfSchool.WebApi.Tests/Services/NotificationServiceTests.cs
DKymachynskyi: > (Redundant fields) Some dependencies used in `notificationService` - `loggerMock`, `localizerMock`, and `notificationHubMock` are not touched by tests, so could be easily dropped from fields list and even inlined in constructor.
> 
> Having the full list of fields (as it is now) adds extra lines of code with no benefits, as set of really used mocks is more obscure.
> 
> Ofc, one might say that this is done for the future tests writing, but it's too easy to extract a variable/field using VS refactoring, so it could really be done when needed, and there is no necessity for premature "optimisation".

Fixed. Thank you!

OutOfSchool/OutOfSchool.WebApi.Tests/Services/NotificationServiceTests.cs
```diff
@@ -0,0 +1,270 @@
...
+        for (int i = 0; i < expectedGroupedByType.Count; i++)
+        {
+            TestHelper.AssertTwoCollectionsEqualByValues(
+                expectedGroupedByType[i].GroupedByAdditionalData,
+                result.NotificationsGroupedByType.ElementAt(i).GroupedByAdditionalData);
+        }
```
VadymLevkovskyi: (Excessive code) You are not using `i` in test message, so simpler approach to compare nested collection could be used. I.e. without explicit cycle
```
        TestHelper.AssertTwoCollectionsEqualByValues(
                expectedGroupedByType.SelectMany(_ => _.GroupedByAdditionalData),
                result.NotificationsGroupedByType.SelectMany(_ => _.GroupedByAdditionalData));
```

OutOfSchool/OutOfSchool.WebApi.Tests/Services/NotificationServiceTests.cs
```diff
@@ -0,0 +1,270 @@
...
+        // Arrange
+        var notificationType = NotificationType.Parent;
+        notificationRepositoryMock
+            .Setup(x => x.GetByFilter(
+                It.IsAny<Expression<Func<Notification, bool>>>(),
+                It.IsAny<string>()))
+            .Returns(Task.FromResult<IEnumerable<Notification>>(notifications));
+        var expected = notifications
+            .Select(n => mapper.Map<NotificationDto>(n))
+            .Where(n => n.Type == notificationType)
```
VadymLevkovskyi: (Performance tiny issue) Please filter before mapping (use `.Where()` before `.Select()`), to avoid excessive work.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/NotificationServiceTests.cs
DKymachynskyi: > (Excessive code) `collection.Select(...).Count()` could be simplified to `collection.Count()`

Fixed. Thank you!

OutOfSchool/OutOfSchool.WebApi.Tests/Services/NotificationServiceTests.cs
DKymachynskyi: > (Excessive code) You are not using `i` in test message, so simpler approach to compare nested collection could be used. I.e. without explicit cycle
> 
> ```
>         TestHelper.AssertTwoCollectionsEqualByValues(
>                 expectedGroupedByType.SelectMany(_ => _.GroupedByAdditionalData),
>                 result.NotificationsGroupedByType.SelectMany(_ => _.GroupedByAdditionalData));
> ```

Fixed. Thank you!

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/NotificationControllerTests.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.WebApi.Tests/Services/NotificationServiceTests.cs
DKymachynskyi: Fixed. Thank you!

VadymLevkovskyi (approved):

PR #1381, opened 2024-02-05
Tfront1/workshop full error response

Tfront1:


OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationServiceTests.cs
```diff
@@ -699,7 +700,11 @@ public async Task UpdateApplication_WhenIdIsValid_ShouldReturnApplication()
         applicationRepositoryMock.Setup(a => a.Count(It.IsAny<Expression<Func<Application, bool>>>())).ReturnsAsync(int.MinValue);
 
         // Act
-        var result = await service.Update(update, Guid.NewGuid()).ConfigureAwait(false);
+        var response = await service.Update(update, Guid.NewGuid()).ConfigureAwait(false);
+        Result<ApplicationDto> result = new Result<ApplicationDto>();
+        response.Match<Result<ApplicationDto>>(
+            actionResult => result = new Result<ApplicationDto>(),
+            succeed => result = succeed);
```
VadymLevkovskyi: (Excessive code) Please use `.Match` return value to write shorter and cleaner code. I.e.
` 
        var result = response.Match(errResult => new Result<ApplicationDto>(), result => result);
`

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -393,12 +393,12 @@ public async Task<IActionResult> Update(ApplicationUpdate applicationDto)
...
+            return result.Match<ActionResult>(
+            error => StatusCode((int)error.HttpStatusCode, new { error.Message, error.ApiErrorResponse }),
+            result =>
             {
-                return BadRequest(result.OperationResult.Errors.ElementAt(0).Description);
-            }
-
-            return Ok(result.Value);
+                return Ok(result.Value);
+            });
```
VadymLevkovskyi: (Excessive code) Please use expression-bodied methods when it's one short line of implementation, as it is here, thus you could omit curly braces and just write `result => Ok(result.Value);`

OutOfSchool/OutOfSchool.WebApi/Services/IApplicationService.cs
```diff
@@ -85,7 +86,7 @@ public interface IApplicationService
     /// <param name="applicationDto">Application entity to update.</param>
     /// <param name="providerId">Id of the provider for workshop.</param>
     /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
-    Task<Result<ApplicationDto>> Update(ApplicationUpdate applicationDto, Guid providerId);
+    Task<Either<ErrorResponse, Result<ApplicationDto>>> Update(ApplicationUpdate applicationDto, Guid providerId);
```
VadymLevkovskyi: (Design issue) Type `Either<ErrorResponse, Result<ApplicationDto>` violates SRP principle in a way that you could describe failed state in 2 ways - either returning `Either.Left` (that is `ErrorResponse`) or returning `Either.Right` with failed `Result`. Please remove this discrepancy and fix the `Update` method implementation.

VadymLevkovskyi (changes requested):
(Sorry for checking PR in draft, I was bit inattentive, hope my comments could help)
Please check your code design, and tests coverage

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ApplicationServiceTests.cs
olsavchenk: Done

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
olsavchenk: Done

OutOfSchool/OutOfSchool.WebApi/Services/IApplicationService.cs
olsavchenk: Done

OutOfSchool/OutOfSchool.WebApi/Services/ApplicationService.cs
```diff
@@ -943,21 +941,17 @@ private async Task<Result<ApplicationDto>> ExecuteUpdateAsync(ApplicationUpdate
 
                     if (amountOfApproved >= availableSeats)
                     {
-                        return Result<ApplicationDto>.Failed(new OperationError
-                        {
-                            Code = "400",
-                            Description = "The Workshop is full.",
-                        });
+                        badRequestApiErrorResponse.AddApiError(
```
VadymLevkovskyi: (Small code style suggestion) You could create and use small "helping" extension method here.
Reason: There are many places in code where `ApiErrorResponse` is created for the only `ApiError` and then immediately passed somewhere. Thus every time there is `var response = new ApiErrorResponse` and then `response.AddApiError(...)` and then `return ...response...`

You could create 
```
public static class ApiErrorExtensions {
    public static ApiErrorResponse ToResponse(this ApiError)
        => ...
}
```
And then use it as one-liner in your code. I.e., for the commented place it will look like:
```
return ErrorResponse.BadRequest(ApiErrorsTypes.Application.AcceptRejectedAlreadyApproved().ToResponse());
```

VadymLevkovskyi (approved):
Approved, though it could be bit improved (see other comment), but I don't insist

PR #1383, opened 2024-02-05
Fix creating of Achievement by ProviderAdmin

alrom:


OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AchievementController.cs
```diff
@@ -236,30 +243,24 @@ public async Task<IActionResult> Delete(Guid id)
 
     private async Task<bool> IsUserProvidersOwnerOrAdmin(Guid workshopId)
     {
-        if (!User.IsInRole(nameof(Role.Provider).ToLower()))
+        if (User.IsInRole(nameof(Role.Provider).ToLower()))
```
VadymLevkovskyi: (Code readability) Please invert this if condition to decrease code nesting.

VadymLevkovskyi (changes requested):
There are minor code style and major test coverage issues

alrom:
Added tests

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AchievementControllerTest.cs
```diff
@@ -31,6 +36,139 @@ public void Setup()
...
+        // Act
+        var result = await controller.Create(dto).ConfigureAwait(false) as ObjectResult;
+
+        // Assert
+        Assert.That(result, Is.Not.Null);
+        Assert.AreEqual(403, result.StatusCode);
+    }
+
+    [Test]
+    public async Task CreateAchievement_UserProviderAdminHaveRights_ShouldReturnForbidden()
```
VadymLevkovskyi: (Test naming) Please fix the expected part in test name, as `_ShouldReturnForbidden` seems to be wrong

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AchievementControllerTest.cs
```diff
@@ -31,6 +36,139 @@ public void Setup()
...
+        // Assert
+        Assert.IsInstanceOf<NotFoundObjectResult>(result);
+    }
+
+    [Test]
+    public async Task CreateAchievement_WhenProviderIsBlocked_ShouldReturnForbidden()
+    {
+        // Arrange
+        var dto = new AchievementCreateDTO();
+        workshopService.Setup(s => s.Exists(It.IsAny<Guid>())).ReturnsAsync(true);
```
VadymLevkovskyi: (Test code clarity) Please consider minimizing usage of `It.IsAny<Guid>()`. Because this way it's not clear that we are checking for specific workshop/provider id. Related to this and subsequent tests.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AchievementControllerTest.cs
```diff
@@ -31,6 +36,139 @@ public void Setup()
...
+
+        var user = new ClaimsPrincipal(new ClaimsIdentity(
+            new Claim[]
+            {
+                new Claim(ClaimTypes.Role, nameof(Role.Parent).ToLower()),
+                new Claim("subrole", nameof(Subrole.None).ToLower()),
+                new Claim("sub", Guid.NewGuid().ToString()),
+            }));
+
+        controller.ControllerContext.HttpContext = new DefaultHttpContext { User = user };
```
VadymLevkovskyi: (Test code readability) Selected part is duplicated in many tests. Please consider implementing and usage of 
`private void SetContextUser(Role role, Subrole subrole, Guid sub)`

VadymLevkovskyi (changes requested):
Logic seems to be ok. Tests code could be improved

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AchievementControllerTest.cs
alrom: Renamed

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AchievementControllerTest.cs
alrom: Added method

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/AchievementControllerTest.cs
alrom: Do we have to use the approach with concrete IDs when we test controllers? All services are mocked.

VadymLevkovskyi: In controller implementation you could pass wrong id to the service by mistake. 
(Imagine situation when there are `workshop` / `provider` / `child` / `parent` ids in same controller method - and they all are just `Guid`s. As we don't use any preventive technique like value object pattern - compiler wouldn't warn you if you pass `workshop` id as a `child` id. And your tests wouldn't warn you either!)

alrom: Thanks for your explanation. I replaced IDs in the tests.

VadymLevkovskyi (approved):

PR #1388, opened 2024-02-12
Kymachynskyi/ Add notification for parent when he was unblocked by service provider

DKymachynskyi:


OutOfSchool/OutOfSchool.WebApi/Services/BlockedProviderParentService.cs
```diff
@@ -106,7 +106,10 @@ public async Task<Result<BlockedProviderParentDto>> Unblock(BlockedProviderParen
             throw new ArgumentNullException(nameof(blockedProviderParentUnblockDto));
         }
 
-        var currentBlock = await GetBlock(blockedProviderParentUnblockDto.ParentId, blockedProviderParentUnblockDto.ProviderId).ConfigureAwait(false);
+        var currentBlock = (await GetBlockedProviderParentEntities(
```
VadymLevkovskyi: (Perf issue) You perform reading db for the same data twice - once here (using as a key `ParentId`, and `ProviderId`) and other time in `blockedProviderParentRepository.UnBlock()`. It's bit inefficient. 

Here you get everything and then specify `FirstOrDefault()` for the `IEnumerable`, while you could use `FirstOrDefaultAsync()` for the `IQueryable`.

Similar is done below in the `GetBlock()` and `IsBlocked()` - instead of using `IQueryable.FirstOrDefaultAsync()`/`IEntityAddOnlyRepository.Any()` (both async) - both time `IEnumerable`s `FirstOrDefault()`/`Any()` are used (that are synchronous and thus you have to use `async`/`await` that affects performance even more).

OutOfSchool/OutOfSchool.WebApi/Services/BlockedProviderParentService.cs
```diff
@@ -157,4 +165,13 @@ public async Task<bool> IsBlocked(Guid parentId, Guid providerId)
...
     }
+
+    private async Task<IEnumerable<BlockedProviderParent>> GetBlockedProviderParentEntities(Guid parentId, Guid providerId)
+    {
+        var currentBlock = await blockedProviderParentRepository.GetByFilter(
+            b => b.ParentId == parentId
+                 && b.ProviderId == providerId
+                 && b.DateTimeTo == null).ConfigureAwait(false);
+        return currentBlock;
+    }
```
VadymLevkovskyi: (Code style) This function (or, more precisely, the predicate passed as `whereExpression` to `GetByFilter()`) looks like something (specification pattern?) innate to the repository and thus could be moved there.

VadymLevkovskyi (changes requested):
There are minor performance issue and code style (logic related to repository doesn't reside in it). As suggested changes could affect tests - they weren't reviewed

OutOfSchool/OutOfSchool.WebApi/Services/BlockedProviderParentService.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.WebApi/Services/BlockedProviderParentService.cs
DKymachynskyi: I've moved it to repository. Thank you!

OutOfSchool/OutOfSchool.WebApi/Services/BlockedProviderParentService.cs
```diff
@@ -148,14 +150,20 @@ public async Task<Result<BlockedProviderParentDto>> Unblock(BlockedProviderParen
...
 
     public async Task<bool> IsBlocked(Guid parentId, Guid providerId)
     {
-        var currentBlock = await GetBlockedProviderParentEntities(parentId, providerId).ConfigureAwait(false);
-        return currentBlock.Any();
+        return await blockedProviderParentRepository
+            .Any(b => b.ParentId == parentId
+                 && b.ProviderId == providerId
+                 && b.DateTimeTo == null)
+            .ConfigureAwait(false);
```
VadymLevkovskyi: (Code duplication, performance) 
Couldn't `GetBlockedProviderParentEntities()` be used here as well?
Isn't `async`/`await` excessive for this method?

OutOfSchool/OutOfSchool.WebApi.Tests/Services/BlockedProviderParentServiceTests.cs
```diff
@@ -171,25 +173,43 @@ public async Task Block_WhenDtoIsAlreadyBlocked_ShouldReturnFailed()
...
+
         var blockedParents = new List<BlockedProviderParent>
         {
             unblockEntity,
         };
 
         blockedProviderParentRepositoryMock
-            .Setup(x => x.GetByFilter(It.IsAny<Expression<Func<BlockedProviderParent, bool>>>(), It.IsAny<string>()))
-            .ReturnsAsync(blockedParents);
+            .Setup(x => x.GetBlockedProviderParentEntities(It.IsAny<Guid>(), It.IsAny<Guid>()))
```
VadymLevkovskyi: (Test readability) Please whenever possible replace `It.IsAny` with expected parameters - here they are (I guess) `unblockEntity.ParentId` and `unblockEntity.ProviderId`. Similar in other tests

OutOfSchool/OutOfSchool.WebApi.Tests/Services/BlockedProviderParentServiceTests.cs
```diff
@@ -171,25 +173,43 @@ public async Task Block_WhenDtoIsAlreadyBlocked_ShouldReturnFailed()
...
         blockedProviderParentRepositoryMock
             .Setup(x => x.UnBlock(It.IsAny<BlockedProviderParent>()))
             .ReturnsAsync(unblockEntity);
-        parentRepositoryMock.Setup(x => x.GetById(parent.Id)).ReturnsAsync(parent);
 
         // Act
         var result = await service.Unblock(unblockDto, userId).ConfigureAwait(false);
 
         // Assert
+        notificationServiceMock
```
VadymLevkovskyi, line 206: (Test readability) Please use at least `blockedProviderParentRepositoryMock.VerifyAll()` in assert section, to ensure, that no excessive `mock.Setup()` is done (force you to clean up your test code) and all expected `Setup`s were called.
Similar in tests below

VadymLevkovskyi (changes requested):
Some code duplication, minor performance and test code readability issues

OutOfSchool/OutOfSchool.WebApi/Services/BlockedProviderParentService.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.WebApi.Tests/Services/BlockedProviderParentServiceTests.cs
DKymachynskyi: I've replaced them. Thank you!

OutOfSchool/OutOfSchool.WebApi.Tests/Services/BlockedProviderParentServiceTests.cs
DKymachynskyi: I've added `blockedProviderParentRepositoryMock.VerifyAll()` and cleaned tests code. Thank you!

OutOfSchool/OutOfSchool.WebApi/Services/BlockedProviderParentService.cs
```diff
@@ -124,30 +129,39 @@ public async Task<Result<BlockedProviderParentDto>> Unblock(BlockedProviderParen
...
-    public async Task<bool> IsBlocked(Guid parentId, Guid providerId)
+    public Task<bool> IsBlocked(Guid parentId, Guid providerId)
     {
-        var currentBlock = await blockedProviderParentRepository.GetByFilter(
-            b => b.ParentId == parentId
-                 && b.ProviderId == providerId
-                 && b.DateTimeTo == null).ConfigureAwait(false);
-
-        return currentBlock.Any();
+        return Task.FromResult(blockedProviderParentRepository
```
VadymLevkovskyi: (Concurrency and performance) Please use `Task.FromResult` **only** for synchronously available data (like in `GetNotificationsRecipientIds()` below). Try to find better solution for asynchronous data, usually it is possible. I.e.:
```
    public Task<bool> IsBlocked(Guid parentId, Guid providerId)
        => blockedProviderParentRepository
            .GetBlockedProviderParentEntities(parentId, providerId)
            .AnyAsync();
```

Reasons:
- you don't create another Task object (lower GC/memory pressure), 
- you move decision when to await for a task upper by call-stack (to the client code that knows if some tasks could be parallelized), 
- you don't fake client expectations - that result type `Task` usually means that real (cpu/io-heavy) job could be executed later, but for `Task.FromResult` case we wait the result here and now

OutOfSchool/OutOfSchool.WebApi/Services/BlockedProviderParentService.cs
VadymLevkovskyi: Almost. Almost fixed. Please check comment below. This one is marked resolved

OutOfSchool/OutOfSchool.WebApi/Services/BlockedProviderParentService.cs
DKymachynskyi: I've changed it. Thank you for your detailed clarification.

VadymLevkovskyi (approved):

PR #1393, opened 2024-02-15
Romanchuk/Go to 8 .net

alrom:


OutOfSchool/OutOfSchool.WebApi.Tests/Services/StatusServiceTests.cs
```diff
@@ -30,21 +30,30 @@ public class StatusServiceTests
     private IMapper mapper;
 
     [SetUp]
-    public void Setup()
+    public async Task Setup()
```
VadymLevkovskyi: (I'm curious) This `Setup()` method returns `Task` and is `async`, but I see no `await` inside the method body, and no returns. Could you provide info, why we need such change?

OutOfSchool/OutOfSchool.WebApi.Tests/Services/StatusServiceTests.cs
```diff
@@ -163,15 +175,10 @@ public async Task Delete_WhenIdIsInvalid_ThrowsArgumentOutOfRangeException()
...
-
-            var institutionStatuses = InstitutionStatusGenerator.Generate(5);
-            context.InstitutionStatuses.AddRange(institutionStatuses);
+        context.Database.EnsureDeleted();
+        context.Database.EnsureCreated();
 
-            context.SaveChanges();
-        }
+        // New version of InMemoryTestDB uses ModelBuilder to initialize data from common OutOfSchool contex.
+        // Don't need to use seeding of elements.
```
VadymLevkovskyi: (Commenting standard functionality) Thanks for the sharing this info, at same time could you reconsider the place? 
For me, it looks that this is a standard functionality, not a "hack" or tricky issue, so we don't need to put standard documentation into our code. 
Moreover, there are `SeedDatabase()` in other test classes, and it will be hard (and inefficient) to add the same comment everywhere.

Usually I suggest to put such knowledge sharing notes about important code changes in the PR comment, and also mention it somewhere in team chat/meetings.

OutOfSchool/OutOfSchool.WebApi/Models/Workshops/WorkshopBaseDto.cs
```diff
@@ -1,7 +1,7 @@
 ﻿using System.ComponentModel.DataAnnotations;
 using System.ComponentModel.DataAnnotations.Schema;
-using Castle.Core.Internal;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.IdentityModel.Tokens;
```
VadymLevkovskyi: Is this using required? Because no changes were introduced in the code, just new using added, and it seems to be irrelevant (as are irrelevant `identity tokens` and `workshop dto`)

OutOfSchool/OutOfSchool.WebApi/Services/RatingService.cs
```diff
@@ -1,7 +1,7 @@
 ﻿using System.Linq.Expressions;
 using AutoMapper;
-using Castle.Core.Internal;
 using Microsoft.Extensions.Localization;
+using Microsoft.IdentityModel.Tokens;
```
VadymLevkovskyi: Is this using required? Because no changes were introduced in the code, just new using added, and it seems to be irrelevant

VadymLevkovskyi (commented):
No logical issues found, have few comments about unclear usings and testing code

OutOfSchool/OutOfSchool.WebApi/Models/Workshops/WorkshopBaseDto.cs
alrom: In Microsoft.IdentityModel.Tokens we have extension IsNullOrEmpty() for IEnumerable<T>. We can write this method in our extensions.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/StatusServiceTests.cs
alrom: Ok. It's the new behavior for InMemoryDatabase. I didn't seed this DB. I tried to describe why.

OutOfSchool/OutOfSchool.WebApi/Models/Workshops/WorkshopBaseDto.cs
VadymLevkovskyi: Thank you. So recently it was in `Castle.Core.Internal`, now in `Microsoft.IdentityModel.Tokens`. 
I still suggest either write own implementation or find a related good library with collections extension methods.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/StatusServiceTests.cs
alrom: Sorry. I had async method SeedDatabase before. And I forgot to change it.

OutOfSchool/OutOfSchool.WebApi/Models/Workshops/WorkshopBaseDto.cs
alrom: Our implementation was added.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/StatusServiceTests.cs
alrom: Cleared.

VadymLevkovskyi (approved):

PR #1402, opened 2024-02-28
add config object for emails content url

DmyMi:
As different projects have differently named configuration fields - a way to unify them and show email data correctly

### DIFF
--- a/OutOfSchool/OutOfSchool.AuthCommon/Extensions/AuthCommonServiceExtensions.cs
+++ b/OutOfSchool/OutOfSchool.AuthCommon/Extensions/AuthCommonServiceExtensions.cs
@@ -8,6 +8,7 @@ using OutOfSchool.AuthCommon.Util;
 using OutOfSchool.AuthCommon.Validators;
 using OutOfSchool.AuthCommon.ViewModels;
 using OutOfSchool.Common.Models;
+using OutOfSchool.RazorTemplatesData.Config;
 
 namespace OutOfSchool.AuthCommon.Extensions;
 
@@ -32,7 +33,8 @@ public static class AuthCommonServiceExtensions
         services.AddEmailSenderService(
             isDevelopment,
             mailConfig.SendGridKey,
-            builder => builder.Bind(config.GetSection(EmailOptions.SectionName)));
+            builder => builder.Bind(config.GetSection(EmailOptions.SectionName)))
+            .AddEmailRendererConfiguration(new EmailContentConfig(config.GetSection("Identity")["Authority"]));
 
         services.Configure<ChangesLogConfig>(config.GetSection(ChangesLogConfig.Name));
         services.Configure<HostsConfig>(config.GetSection(HostsConfig.Name));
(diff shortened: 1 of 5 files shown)


OutOfSchool/OutOfSchool.RazorTemplatesData/Config/EmailContentConfig.cs
```diff
@@ -0,0 +1,20 @@
...
+    private readonly string EmailContentHost;
+    private static readonly string EmailContentPath = "/_content/email";
+
+    public string EmailContentUrl
+    {
+        get
+        {
+            return string.Concat(EmailContentHost, EmailContentPath);
+        }
+    }
```
VadymLevkovskyi: (Code style) Please use less verbose syntax for simple getter-only properties. I.e.
```
public string EmailContentUrl => string.Concat(EmailContentHost, EmailContentPath);
```

OutOfSchool/OutOfSchool.WebApi/Startup.cs
```diff
@@ -485,7 +486,8 @@ public static void AddApplicationServices(this WebApplicationBuilder builder)
         services.AddEmailSender(
             builder.Environment.IsDevelopment(),
             mailConfig.SendGridKey,
-            builder => builder.Bind(configuration.GetSection(EmailOptions.SectionName)));
+            builder => builder.Bind(configuration.GetSection(EmailOptions.SectionName)))
+            .AddEmailRendererConfiguration(new EmailContentConfig(configuration.GetSection("Hosts")["BackendUrl"]));
```
VadymLevkovskyi, line 500: (Logic) `configuration.GetSection("Hosts")["BackendUrl"]` - here string keys are different from mentioned in `AuthCommonServiceExtensions.cs`. Is it intended to have same thing named differently?

OutOfSchool/OutOfSchool.AuthCommon/Extensions/AuthCommonServiceExtensions.cs
```diff
@@ -31,7 +32,8 @@ public static void AddAuthCommon(this IServiceCollection services, Configuration
         services.AddEmailSender(
             isDevelopment,
             mailConfig.SendGridKey,
-            builder => builder.Bind(config.GetSection(EmailOptions.SectionName)));
+            builder => builder.Bind(config.GetSection(EmailOptions.SectionName)))
+            .AddEmailRendererConfiguration(new EmailContentConfig(config.GetSection("Identity")["Authority"]));
```
VadymLevkovskyi, line 37: (Code style) Please consider storing string literals in related classes (see f.e. `EmailOptions.SectionName` one line above). Especially, when they are reused more than twice

OutOfSchool/OutOfSchool.WebApi/Startup.cs
DmyMi: Sorry to have come back to this PR after a few months :) but had to do other email refactorings first. Yes the reason to create the config object is to unify the string keys because in their own projects they have own logic to be used that way and we wanted to reuse the keys without dublicating.

VadymLevkovskyi: > Yes the reason to create the config object is to unify the string keys ...

If I get you correctly, some other commit to this PR should be done soon, where all string literals be done as `EmailContentConfig` class constants and reused in both `WebApi` and `Auth` library, right? 

If so, could PR be in Draft mode until that commit is done?

PR #1406, opened 2024-03-01
Kymachynskyi/ Fix Null Reference Exception in CATTOTGAddressExtension for special status cities

DKymachynskyi:


OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/CATOTTGAddressExtensionsTests.cs
```diff
@@ -220,6 +233,19 @@ public void GetDistrictName_WhenCATOTTGIsCityDistrictWithParent_ShouldReturnDist
...
+    public void GetDistrictName_WhenCATOTTGIsCityDistrictWithSpecialStatusParent_ShouldReturnNull()
+    {
+        // Arrange
+        string expected = null;
+
+        // Act
+        var result = CatottgAddressExtensions.GetDistrictName(cityDistrictWithSpecialStatusParent);
+
+        // Assert
+        Assert.AreEqual(expected, result);
```
VadymLevkovskyi: (Code style) `Assert.IsNull` is more clear, and you could remove 3 lines of code of you `Arrange` part. Less code, more clarity. 
There is more than one test with such pattern. 
And I know, that it's technically not code of this PR, but as you've changed names of tests, you are kinda responsible ;)

DKymachynskyi: I've changed them. Thank you!

VadymLevkovskyi (approved):

PR #1407, opened 2024-03-03
Tsylia/Licenses

andriitsylia:


OutOfSchool/Directory.Packages.props
```diff
@@ -11,7 +11,10 @@
   <ItemGroup>
     <!--Basic Microsoft Packages-->
     <PackageVersion Include="Microsoft.AspNet.Identity.Core" Version="2.2.4" />
+    <PackageVersion Include="Microsoft.Extensions.Configuration" Version="6.0.0" />
```
alrom, line 14: You can increase version to 8

OutOfSchool/OutOfSchool.Lisences/Services/ProjectPackagesInfoProcessor.cs
```diff
@@ -0,0 +1,64 @@
...
+
+                    if (nameVersion.Length == 2
+                        && element.Value.TryGetProperty(Constants.TypeProperty, out var type)
+                        && type.ValueEquals(Constants.PackageProperty))
+                    {
+                        packages.Add(new PackageBase
+                        {
+                            Name = nameVersion[0],
+                            Version = nameVersion[1]
+                        });
```
VadymLevkovskyi, line 54: (Logic) Could `packages` contain duplicates (same name and version)? Isn't it better to avoid this explicitely?

VadymLevkovskyi:
As I understand, this project is a "helper tool" one, so might have lower quality goals. Thus you could ignore rest of this comment. And that's why I don't comment specific lines of code/files, but instead provide the summary comment here. 

If you still read it and didn't understand - please ping me in Teams.

(Tests) There are no tests.
(Design) I can't understand your OOP design decisions - as input of one class is output of another - they seem too coupled, and the whole idea of having separate classes just bloats the code (some classes could be transformed to single method). (Interfaces) They are not used.
(Input data checks) You don't check input data in right place - i.e. options/arguments, license files existence...  
(Code style) Please consider using LINQ more heavily - many of your `foreach`es could be rewritten with more LINQ and less code, nested cycles as well.

OutOfSchool/Directory.Packages.props
andriitsylia: Yes, I'll do it.

OutOfSchool/OutOfSchool.Lisences/Services/ProjectPackagesInfoProcessor.cs
andriitsylia: There are no duplicate packages within a single project, but across multiple projects, there are many. The output file includes information about packages in each project, and in this particular case, several packages appear in multiple projects.

andriitsylia:
@VadymLevkovskyi, I've intended to do a lot, but I've forgotten about KISS and YAGNI :)

PR #1418, opened 2024-03-14
Workshop title was added to notification

alrom:


VadymLevkovskyi:
Please provide new or adjust existing unit tests

alrom:
I added a test

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/WorkshopServicesCombinerDBTests.cs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    [TearDown]
+    public void Dispose()
+    {
+        dbContext.Dispose();
+    }
+
+    [Test]
+    public async Task Delete_WhenCalled_CreateNotificationWithTitleInAdditionalData()
```
VadymLevkovskyi: (Test design) I hadn't tried proposed approach in Visual Studio, so asking you to check it.

Test seems to be overcomplicated for the 1 line of code change in `WorkshopServicesCombiner`. Could you instead of creating real `notificationService` just mock it and call `.Verify(x => x.Create(...), Times.Once)` on that mock to assert that `additionalData` dictionary contains `title` data?

Thus you could probably don't use `dbContext` at all (both performance and less code, I hope). Also `SetUp()` would be simpler.

alrom: Done

VadymLevkovskyi: Thank you. One last minor thing left to improve - see other comment

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/WorkshopServicesCombinerDBTests.cs
```diff
@@ -0,0 +1,96 @@
...
+
+namespace OutOfSchool.WebApi.Tests.Services.Database;
+
+[TestFixture]
+public class WorkshopServicesCombinerDBTests
+{
+    private Mock<IWorkshopService> workshopService;
+    private Mock<IApplicationRepository> applicationRepository;
+    private Mock<IEntityRepositorySoftDeleted<long, Favorite>> favoriteRepository;
+    private Mock<IElasticsearchSynchronizationService> elasticsearchSynchronizationService;
```
VadymLevkovskyi: (Excessive code) You could get rid of these 3 fields, as they are used only in `SetUp()`, thus could be made a local variables there.

alrom: Of course. Sorry for my inattention. I fixed it.

VadymLevkovskyi (approved):

PR #1423, opened 2024-03-20
Implemented email sender as quartz job with transactional outbox pattern

### LINKED ISSUE #1420: Implement Transactional Outbox Pattern with SendGrid Email Sender and Quartz Job Scheduling
# Description
As part of ongoing efforts to ensure the reliability of our email sending functionality, we need to implement the transactional outbox pattern.

# Requirements
1. **Transactional Outbox Implementation:** Integrate the transactional outbox pattern as described in [Microservices.io - Transactional Outbox](https://microservices.io/patterns/data/transactional-outbox.html). The implementation should capture outgoing email messages in a database outbox as part of the same transaction as the business operation.
   
2. **Quartz Job for Outbox Polling:** Implement a Quartz job that regularly polls the transactional outbox for unsent email messages. Once an email message is found, the job should send it via SendGrid and mark it as sent in the database.

3. **Rate Limit Handling:** Investigate SendGrid's API to determine if it provides information on how many emails can be sent before hitting daily limits. If such an API exists or if there's any other mechanism to retrieve this information, implement a feature in the Quartz job to postpone further email sending until the next day upon reaching the limit. This is crucial for preventing rate limit errors and ensuring all emails are sent in a timely manner.

# Acceptance Criteria
- [ ] Transactional outbox pattern is implemented and integrated with the application's database.
- [ ] Quartz job is set up to poll the transactional outbox and send emails through SendGrid.
- [ ] Rate limit handling is implemented, with the ability to postpone sending emails upon reaching daily limits.

olsavchenk:
Please do not pay attention to the test coverage for now. I have questions about unit tests for this task, I will ask them during the next daily meeting and cover the code with tests.


OutOfSchool/OutOfSchool.EmailSender/EmailSenderService.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        var outboxMessage = new EmailOutbox()
+        {
+            Email = email,
+            Subject = subject,
+            HtmlContent = content.html,
+            PlainContent = content.plain,
+            CreationTime = DateTime.Now,
+        };
+        await emailOutboxRepository.Create(outboxMessage);
```
DmyMi: i think it's better to add a `Base64` encoding step for both html & plain content before saving to DB. As those emails can have special symbols or contain keywords that may break database when exporting/importing or backing up.

OutOfSchool/OutOfSchool.DataAccess/Models/EmailOutbox.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+namespace OutOfSchool.Services.Models;
+public class EmailOutbox : IKeyedEntity<Guid>
```
DmyMi: not sure if it matters much, but this is internal data so we can use `long` ids .

OutOfSchool/OutOfSchool.DataAccess/Models/EmailOutbox.cs
```diff
@@ -0,0 +1,17 @@
...
+
+    public string Email { get; set; }
+
+    public string Subject { get; set; }
+
+    public string HtmlContent { get; set; }
+
+    public string PlainContent { get; set; }
+
+    public DateTime CreationTime { get; set; }
```
DmyMi: it depends on implementation of the quartz job (will it delete an email after sending or not), but for consideration: can add some metadata fields for debugging/analysis, like the sent (delivered) time, expiration time (for example if we send an email with some token we probably should not send it if it was postponed till the token has expired), optionally store error why it was not sent (we can get it in logs so don't know if should store in db)

OutOfSchool/Directory.Packages.props
```diff
@@ -67,7 +67,7 @@
     <PackageVersion Include="AutoMapper" Version="$(AutoMapVersion)" />
     <PackageVersion Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="$(AutoMapVersion)" />
     <!--Quartz-->
-    <PackageVersion Include="Quartz.AspNetCore" Version="$(QuartzVersion)" />
+    <PackageVersion Include="Quartz.AspNetCore" Version="3.5.0" />
```
VadymLevkovskyi: (Code style) `$(QuartzVersion)` is same as your `3.5.0`. Please revert your change.

OutOfSchool/OutOfSchool.EmailSender/EmailSenderJob.cs
```diff
@@ -0,0 +1,94 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using OutOfSchool.Services.Enums;
+using OutOfSchool.Services.Repository;
+using Quartz;
+using SendGrid;
+using SendGrid.Helpers.Mail;
+
+namespace OutOfSchool.EmailSender;
+public class EmailSenderJob : IJob
```
VadymLevkovskyi: (Code style) Please add empty line between namespace and class defiition.

OutOfSchool/OutOfSchool.EmailSender/IEmailSenderService.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace OutOfSchool.EmailSender;
+
+public interface IEmailSenderService
+{
+    Task SendAsync(string email, string subject, (string html, string plain) content, DateTime expirationTime);
```
VadymLevkovskyi: (Code style) In most calls the `expirationTime` is set to `DateTime.MaxDate`. Couldn't it be specified as optional param value here in interface and in implementations?

VadymLevkovskyi (commented):
There are a few code style minor things. 

And one major design question - why not use Quarz.NET capabilities? Like [Job Store](https://www.quartz-scheduler.net/documentation/quartz-3.x/tutorial/job-stores.html), specific [job instances](https://www.quartz-scheduler.net/documentation/quartz-3.x/tutorial/more-about-jobs.html#job-instances) per separate email and job throttling using [trigger listener](https://www.quartz-scheduler.net/documentation/quartz-3.x/tutorial/trigger-and-job-listeners.html) (see [example](https://stackoverflow.com/a/45813648) on StackOverflow).

olsavchenk:
@VadymLevkovskyi Could you please provide me with more information on how to create a separate job for each email? From what I have researched, the only way to pass the email data to the job is to use `.UsingJobData()`, which seems wrong as we need to check the database table for new records every time the job is executed. Thank you in advance for your reply.

VadymLevkovskyi:
> @VadymLevkovskyi Could you please provide me with more information on how to create a separate job for each email? From what I have researched, the only way to pass the email data to the job is to use `.UsingJobData()`, which seems wrong as we need to check the database table for new records every time the job is executed. Thank you in advance for your reply.

You check db now, because you are implementing that outbox pattern by yourself. 
At same time Quarz.Net provides job persistence to db, that is already an implemented outbox pattern.

So my idea is to:
- Ask Quarz it to use persistence (so all job instances would be implicitely stored in db) - beforementioned `Job Store`. 
- Have a single job class (`identity`) for sending mails - it just takes JobData, email options and send email, and knows nothing about db.
- Use for this job class not parallel execution (to avoid concurrent bothering of already blocked/down resource) - `[DisallowConcurrentExecution]`.
- Schedule in Quarz job "instance" for mentioned class with all necessary params (email "to" address, subject, content) set via `.UsingJobData()`.
- Use single job trigger that periodically tries to perform all job instances (as it is already used in `EmailSenderExtensions`, and configured `.WithCronSchedule()`) - so at some time it will start (one-by-one) all job instances (form the Job Store).
- Use trigger listener to veto job execution when sendGrid is inaccessible.
- Use job listener to check JobExecutionException and catch when sendGrid is inaccessible.
- To share information between listeners about sendGrid accessibility - introduce simple service storing this information.
- 3 last steps (listeners and "accessibility" service) could be implemented inside one class - to have all logic encapsulated.

OutOfSchool/OutOfSchool.EmailSender/EmailSenderService.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        expirationTime ??= DateTime.MaxValue;
+
+        var jobData = new JobDataMap
+        {
+            { "Email", email },
+            { "Subject", subject },
+            { "HtmlContent", EncodeToBase64(content.html) },
+            { "PlainContent", EncodeToBase64(content.plain) },
+            { "ExpirationTime", expirationTime },
```
VadymLevkovskyi: (Code style) These 5 string literals are used here and in EmailSenderJob. Probably they could be introduced as string constants?

OutOfSchool/OutOfSchool.EmailSender/EmailSenderJob.cs
```diff
@@ -0,0 +1,92 @@
...
+        IOptions<EmailOptions> emailOptions,
+        ILogger<EmailSenderJob> logger,
+        ISendGridClient sendGridClient)
+    {
+        this.emailOptions = emailOptions;
+        this.logger = logger;
+        this.sendGridClient = sendGridClient;
+    }
+
+    public async Task Execute(IJobExecutionContext context)
```
VadymLevkovskyi: (Exception handling) From [Quartz.Net Job documantation](https://www.quartz-scheduler.net/documentation/quartz-3.x/tutorial/more-about-jobs.html#jobexecutionexception) it's seen that 
- whole job has to be wrapped in try/catch
- you have to manually throw `JobExecutionException` if something goes wrong (i.e. when `response.StatusCode == HttpStatusCode.TooManyRequests`)

OutOfSchool/OutOfSchool.EmailSender/SendGridAccessibilityService.cs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public Task JobToBeExecuted(IJobExecutionContext context, CancellationToken cancellationToken = default)
+    {
+        return Task.CompletedTask;
+    }
+
+    public Task JobWasExecuted(IJobExecutionContext context, JobExecutionException jobException, CancellationToken cancellationToken = default)
+    {
+        if(jobException != null)
```
VadymLevkovskyi: (Exception handling) As mentioned in other comment, this condition will never occur. You have to throw `JobExecutionException` manually

OutOfSchool/OutOfSchool.EmailSender/SendGridAccessibilityService.cs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public Task TriggerMisfired(ITrigger trigger, CancellationToken cancellationToken = default)
+    {
+        return Task.CompletedTask;
+    }
+
+    public Task<bool> VetoJobExecution(ITrigger trigger, IJobExecutionContext context, CancellationToken cancellationToken = default)
+    {
+        if (!sendGridAccessibilityService.IsSendGridAccessible)
```
VadymLevkovskyi: (Logic) Please convince me that `sendGridAccessibilityService.IsSendGridAccessible` would ever be `true`. It is not set to `true` in constructor, it is never set to `true` in other code. Looks like mail would never be sent.

OutOfSchool/OutOfSchool.EmailSender/SendGridAccessibilityService.cs
```diff
@@ -0,0 +1,87 @@
...
+
+namespace OutOfSchool.EmailSender;
+
+public interface ISendGridAccessibilityService 
+{
+    bool IsSendGridAccessible { get; set; }
+}
+public class SendGridAccessibilityService : ISendGridAccessibilityService
+{
+    public bool IsSendGridAccessible { get; set; }
```
VadymLevkovskyi: (Logic) I doubt if only 1 `bool` property would be enough. I'm expecting for some date of deprecation and related logic.

OutOfSchool/OutOfSchool.EmailSender/SendGridAccessibilityService.cs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+public class EmailSenderJobListener : IJobListener
```
VadymLevkovskyi: (Code style) Too many different classes in one file. Please either split the file or join them all into 1 class.

VadymLevkovskyi (changes requested):
I'm not sure that mail would ever be sent. Please check comments for details

olsavchenk:
I hope I have implemented everything needed, but I am not sure about the registration of the DI SendGridAccessibilityService, should I change it to Singleton so that it stores the state (IsSendGridAccessible property) or Scoped will be enough?

OutOfSchool/OutOfSchool.EmailSender/Services/SendGridAccessibilityService.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// </summary>
+    /// <returns>Midnight of the next day is SendGrid accessible and null if it is not.</returns>
+    public DateTimeOffset? GetNextStartDate()
+    {
+        if (IsSendGridAccessible)
+        {
+            return null;
+        }
+        return DateTimeOffset.Now.Date.AddDays(1);
+    }
```
VadymLevkovskyi: (Logic) Please help me understand the logic. 

What I see right now is that after first inaccessibility the send grid would never be accessible, because `IsSendGridAccessible` is still never set to `true` after being set to `false`. Am I right?

--- 

My initial suggestion about DateTime was bit different than you have implemented.
I've expected you to have in this class 2 methods and a private field:

```
private DateTimeOffset accessibleAfter = DateTimeOffset.Now;

public void SetSendGridInaccessible(DateTimeOffset now)
  => accessibleAfter = now + ...; // some logic to postpone the next access from now

public bool IsSendGridAccessible(DateTimeOffset now)
  => accessibleAfter <= now;
```

Thus, `job listener` could call `SetSendGridInaccessible(DateTimeOffset.Now)` in case of any issues and `trigger listener` could check `veto` condition using `IsSendGridAccessible(DateTimeOffset.Now)`.

Design remark: `now` is passed in both methods to make `SendGridAccessibilityService` more testable (it's usually hard to test logic that implicitly uses `DateTime.Now`).

Logic remark: using of `DateTimeOffset.Now.Date.AddDays(1);` is not the best from my perspective. I know nothing about server quotas so please correct me, but generally speaking there are such considerations:
- we could get different timeouts - you might postpone next mail send in a 1 second or in a 24hrs - pretty uneven; 
- secondly, midnight of our timezone looks like a good candidate for everyone in our timezone, so long queue could be formed;
- and lastly, our timezone midnight is nothing special for other timezones, and thus some inner counter/metric of the sendGrid could be not reset at this time.
Thus I would use simpler logic `now + Timespan.FromMinutes(timeoutValueFromConfig ?? defaultTimeoutValue)` where defaultTimeoutValue could be set to 360 (6hrs) or even less.

And of course the `SendGridAccessibilityService` has to be registered as singleton.

OutOfSchool/OutOfSchool.EmailSender/SendGridAccessibilityService.cs
VadymLevkovskyi: Closing this one, but issue seems to be not resolved. See my separate (bigger) comment below

OutOfSchool/OutOfSchool.EmailSender/SendGridAccessibilityService.cs
VadymLevkovskyi: Closing this one, but issue seems to be not resolved. See my separate (bigger) comment below

OutOfSchool/OutOfSchool.EmailSender/Services/EmailSenderService.cs
```diff
@@ -0,0 +1,48 @@
...
+namespace OutOfSchool.EmailSender.Services;
+
+public class EmailSenderService : IEmailSenderService
+{
+    private readonly IScheduler scheduler;
+
+    public EmailSenderService(IScheduler scheduler)
+    {
+        this.scheduler = scheduler;
+        this.scheduler.Start().Wait();
```
VadymLevkovskyi: (Logic) I'm not sure if `scheduler` has to be started here, as there is already one start in the `QuartzExtension`. Moreover, it is usually not recommended to use asynchronous (or long lasting) calls in constructor.

OutOfSchool/OutOfSchool.AuthServer.Tests/EmailSender/EmailSenderJobTests.cs
```diff
@@ -0,0 +1,157 @@
...
+            });
+        _mockSendGridClient.Setup(client => client.SendEmailAsync(It.IsAny<SendGridMessage>(), It.IsAny<CancellationToken>()))
+            .ReturnsAsync(new Response(HttpStatusCode.TooManyRequests, null, null));
+
+        // Act & Assert
+        Assert.ThrowsAsync<JobExecutionException>(() => _emailSenderJob.Execute(mockContext.Object));
+    }
+
+    [Test]
+    public async Task Execute_WithSendGridError_ShouldThrowException()
```
VadymLevkovskyi: (Test name) The test name mismatch assert. Please fix it

VadymLevkovskyi (approved):
Code seems OK, but Sonar detects 4 new issues. I'm approving the PR but could you double check those issues? (And probably fix them)

UPD: sorry, has to unresolve one of my last comments, please check it

OutOfSchool/OutOfSchool.AuthServer.Tests/EmailSender/EmailSenderJobTests.cs
VadymLevkovskyi: (I was bit inattentive while resolving this comment) Test name is still misleading. 

It states "ShouldNotSendEmail" but in assert you verify `_mockSendGridClient.Verify(... client.SendEmailAsync(...), Times.Once);`.

Maybe name should be something like `ShouldNotThrow` ?

OutOfSchool/OutOfSchool.EmailSender/Services/EmailSenderService.cs
```diff
@@ -0,0 +1,47 @@
...
+            .UsingJobData(jobData)
+            .StoreDurably()
+            .Build();
+
+        await scheduler.AddJob(job, false);
+    }
+
+    private string EncodeToBase64(string toEncode)
+    {
+        byte[] toEncodeAsBytes = Encoding.ASCII.GetBytes(toEncode);
```
vpustovoitov: Using ASCII encoding in emails can lead to issues with Ukrainian characters in the content.
Same in `EmailSenderJob`.

olsavchenk: This was indeed a problem. Thank you for your comment and attention. I fixed it, could you please re-review the code

PR #1425, opened 2024-03-20
Vladp/add default roles initialization

MrVenik:
## Summary
In the past, during the migration of our project from IdentityServer4 to OpenIddict, there was an accidental removal of the default role seeding in the database. This PR restores the necessary role seeding.


OutOfSchool/OutOfSchool.AuthCommon/Services/IdentityRolesInitializerHostedService.cs
```diff
@@ -0,0 +1,102 @@
...
+    [
+        "parent",
+        "provider",
+        "techadmin",
+        "ministryadmin",
+        "regionadmin",
+        "areaadmin",
+    ];
+
+    public async Task StartAsync(CancellationToken cancellationToken)
```
VadymLevkovskyi, line 35: (Logic) Please consider usage of cancellation token. (I expect it to be checked at least inside `foreach` cycle and while writing logs - cancel action could be logged as well).

OutOfSchool/OutOfSchool.AuthCommon/Services/IdentityRolesInitializerHostedService.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        using var scope = serviceProvider.CreateScope();
+
+        string[] defaultRoleNames = GetDefaultRoleNames();
```
VadymLevkovskyi: (Code style) Please use for local variables everywhere either `var` or explicit types, not both (I have reasons to use `var`, but don't insist). Current approach is bit inconsistent and noisy.

OutOfSchool/OutOfSchool.AuthCommon/Services/IdentityRolesInitializerHostedService.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        var existingRoleNames = await roleManager.Roles
+            .Where(role => defaultRoleNames.Contains(role.Name))
+            .Select(role => role.Name)
+            .ToArrayAsync(cancellationToken)
+            .ConfigureAwait(false);
+
+        var newRoles = defaultRoleNames
+            .Except(existingRoleNames)
+            .Select(roleName => new IdentityRole { Name = roleName });
```
VadymLevkovskyi: (Logic, performance) In selected code you are filtering roles twice: remove from existing all that are not default, and then remove from default all that are not existing. It takes excessive code and CPU. 
Instead you could make filtering one time - take all existing role names from roleManager (not filtered) and exclude them from default.

VadymLevkovskyi (approved):
See no big issues, there are few minor (performance, code style) mentioned below

VadymLevkovskyi (approved):

PR #1426, opened 2024-03-21
Kymachynskyi/ 1414 Fix password validation error message

DKymachynskyi:
Replace `Identity.PasswordOptions` validation with `AuthCommon.Validators.CustomPasswordValidator`.
Fix password validation error message.
Replace `ExecuteSqlRawAsync` with `ExecuteUpdateAsync` in `NotificationRepository.SetReadDateTimeForAllUnreaded` method.


OutOfSchool/OutOfSchool.Common/Constants.cs
```diff
@@ -30,7 +30,11 @@ public static class Constants
 
     public const string EmailRegexViewModel = @"^([\w\.\-]+)@([\w\-]+)((\.(\w){2,3})+)$";
 
-    public const string PasswordRegexViewModel = @"^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[@$!%*?&])[A-Za-z\d@$!%*?&]{6,}$";
+    public const string PasswordRegexViewModel = @"^(?=.*[\p{Lu}])(?=.*[\p{Ll}])(?=.*\d)(?=.*[@$!%*?&])[\p{Lu}\p{Ll}\d@$!%*?&]{8,}$";
```
VadymLevkovskyi: (Usage of regex) Please explain why we have to use such complex regular expression (having few `Zero-width positive lookahead assertion`s [see Regex reference](https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-language-quick-reference#grouping-constructs) ) instead of implemented check logic directly in the code (as we already have separate validator)? 
I presume that direct logic in code would be both faster and more self-explanatory and let you provide better error message for separate cases ("password is too short", "special symbols not used", etc) .

OutOfSchool/OutOfSchool.AuthCommon/Validators/CustomPasswordValidator.cs
```diff
@@ -0,0 +1,28 @@
+﻿using static OutOfSchool.AuthCommon.Validators.GeneratedRegexes;
+
+namespace OutOfSchool.AuthCommon.Validators;
+
+public class CustomPasswordValidator(string passwordValidationErrorMessage, string passwordRequiredErrorMessage) : IPasswordValidator<User>
+{
+    public Task<IdentityResult> ValidateAsync(UserManager<User> manager, User user, string? password)
+    {
+        List<IdentityError> errors = [];
```
VadymLevkovskyi: (Logic, performance) Please let me understand why we need to have list of errors in the code that could have max 1 error? You create an empty list, then add something there (thus "reallocating" the list), then convert it to array using spread operator - excessive CPU, code, memory pressure. Why? 
Please consider usage of simple immediate `return` in each `if` for errors, and at method end for no errors.

OutOfSchool/OutOfSchool.DataAccess/Repository/NotificationRepository.cs
```diff
@@ -47,10 +47,8 @@ public async Task<IEnumerable<Notification>> SetReadDateTimeByType(string userId
...
     public async Task SetReadDateTimeForAllUnreaded(string userId, DateTimeOffset dateTime)
     {
-        // TODO: replace it in EFCore8 with bulk update
-        await db.Database.ExecuteSqlRawAsync(
-            @"UPDATE Notifications SET ReadDateTime = {0} WHERE UserId = {1} AND ReadDateTime IS NULL;",
-            dateTime,
-            userId);
+        await db.Notifications
+            .Where(n => n.UserId == userId && n.ReadDateTime == null)
+            .ExecuteUpdateAsync(s => s.SetProperty(n => n.ReadDateTime, dateTime));
```
VadymLevkovskyi: (Irrelevant code changes) Thanks for the improvement, but It is usually better to introduce such changes in separate branch and PR.

VadymLevkovskyi (commented):
See no breaking issues, but suggest to avoid using regexp.
Also unit-tests are missing!

OutOfSchool/OutOfSchool.Common/Constants.cs
DKymachynskyi: I have only one explanation: I didn’t consider the possibility of replacing the regular expression with checking logic, since the previous code already used a similar regular expression, in addition, after studying Microsoft's recommendations, I added the `GeneratedRegex` attribute for its acceleration. After your recommendation, I see that the direct logic in this case will be more appropriate.
As for the validation error messages, according to the requirements, we should output only one - a general validation error message for all cases, except when password is empty ("Password is required").
Thank you!

OutOfSchool/OutOfSchool.Common/Extensions/StringExtensions.cs
```diff
@@ -35,4 +36,15 @@ public static string Right(this string value, int length)
...
+    public static bool ContainsCharacterType(this string source, Func<char, bool> characterType)
+    {
+        return source.Any(characterType);
+    }
+
+    public static bool ContainsAnySymbol(this string source, string symbols)
+    {
+        return !string.IsNullOrEmpty(source) && !string.IsNullOrEmpty(symbols)
+            && symbols.Any(source.Contains);
+    }
```
VadymLevkovskyi: (Small side note) This is not a call to action, just for your information.
I've tried to measure performance (on .Net 8.0) using different search approaches - regExp, your code, and dumb single-iteration algorithm. Code and results - on screenshot below
image

DKymachynskyi: @VadymLevkovskyi Thank you! It is very useful and interesting for me. Before I saw your benchmarks, I rewrote `IsValidPassword` because I thought that it was necessary to limit the characters in the password to only permitted requirements:
image

VadymLevkovskyi: And your variant for checking char "type" is even faster than `CharUnicodeInfo.GetUnicodeCategory(c)` I've used (for the code see my previous screen, but I've removed there early returns, as it was wrong logic). 

Your code with performance comparison is below
image

OutOfSchool/OutOfSchool.AuthCommon/Validators/CustomPasswordRules.cs
```diff
@@ -0,0 +1,35 @@
+﻿namespace OutOfSchool.AuthCommon.Validators;
+public static class CustomPasswordRules
+{
+    public static bool IsValidPassword(string? password)
```
VadymLevkovskyi: (Code design) Could you make this class and method non-static and implementing some interface? This makes related `CustomPasswordValidationAttribute` code (uncovered now) testing easier.

OutOfSchool/OutOfSchool.DataAccess/Repository/NotificationRepository.cs
VadymLevkovskyi: Additionally - this change reduces test coverage and introduces "single await in return" smell (I don't know if any related term exists) when you have method like `async Method() => await otherMethod()`, that works, but adds extra context switches, and degrades performance

OutOfSchool/OutOfSchool.AuthCommon/Validators/CustomPasswordRules.cs
DKymachynskyi: @VadymLevkovskyi I've made the requested changes. Thank you!

OutOfSchool/OutOfSchool.DataAccess/Repository/NotificationRepository.cs
DKymachynskyi: @VadymLevkovskyi should I make this method synchronous?

OutOfSchool/OutOfSchool.AuthCommon/Validators/ICustomPasswordRules.cs
```diff
@@ -0,0 +1,5 @@
+﻿namespace OutOfSchool.AuthCommon.Validators;
+public interface ICustomPasswordRules
```
VadymLevkovskyi, line 3: (Code style) Please add empty line after namespace line.

OutOfSchool/OutOfSchool.DataAccess/Repository/NotificationRepository.cs
VadymLevkovskyi: It depends on your meaning of "synchronous" ;) 

As for me, method that returns `Task` is still asynchronous, even when it doesn't have `async`/`await` keywords.
So next version is still asynchronous, but doesn't introduce state machine and context switching as `async`/`await `version does
```
    public Task SetReadDateTimeForAllUnreaded(string userId, DateTimeOffset dateTime)
    {
        return db.Notifications...

```

OutOfSchool/OutOfSchool.AuthCommon/Validators/CustomPasswordValidationAttribute.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    public CustomPasswordValidationAttribute()
+    {
+        passwordRules = new CustomPasswordRules();
+    }
+
+    public CustomPasswordValidationAttribute(ICustomPasswordRules passwordRules)
+    {
+        this.passwordRules = passwordRules;
+    }
```
VadymLevkovskyi: (Code style) Please avoid manual creation of class instance (i.e. you have `new CustomPasswordRules();`) as this usually introduce high coupling and breaks testability. Please use DI/Service Locator patterns instead. 

For this specific case, you need `ICustomPasswordRules` only inside the `IsValid`, and you have `ValidationContext.GetService()` that could serve your needs. Thus, there is no necessity to have private field, parameterless constructor and constructor with injection.

OutOfSchool/OutOfSchool.AuthCommon/Validators/CustomPasswordValidator.cs
```diff
@@ -0,0 +1,35 @@
...
+public class CustomPasswordValidator : IPasswordValidator<User>
+{
+    private readonly IStringLocalizer<SharedResource> localizer;
+    private readonly ICustomPasswordRules passwordRules;
+
+    public CustomPasswordValidator(IStringLocalizer<SharedResource> localizer)
+    {
+        this.localizer = localizer;
+        passwordRules = new CustomPasswordRules();
+    }
```
VadymLevkovskyi: (Code style) Could this constructor be removed? 

Asking, because manually created class instances (like `new CustomPasswordRules()` we have here) is usually a code smell, that create high coupling and break testability. It's still possible for edge cases, though not for current one (at least I believe so).

VadymLevkovskyi (changes requested):
Please polish your design bit more

OutOfSchool/OutOfSchool.DataAccess/Repository/NotificationRepository.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.AuthCommon/Validators/ICustomPasswordRules.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.AuthCommon/Validators/CustomPasswordValidationAttribute.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.AuthCommon/Validators/CustomPasswordValidator.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.AuthServer.Tests/Validators/CustomPasswordValidationAttributeTests.cs
```diff
@@ -0,0 +1,117 @@
...
+        // Arrange
+        var password = "123";
+        ValidationContext validationContext = null;
+
+        // Act & Assert
+        Assert.Throws<ArgumentNullException>(() => attribute.GetValidationResult(password, validationContext));
+    }
+
+    [Test]
+    public void IsValid_WithCustomPasswordRulesIsNotExist_ThrowNullReferenceException()
```
VadymLevkovskyi: (Naming issue) Test name expects other exception than asserted in test body

OutOfSchool/OutOfSchool.AuthCommon/Validators/CustomPasswordValidator.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Microsoft.Extensions.Localization;
+
+namespace OutOfSchool.AuthCommon.Validators;
+
+public class CustomPasswordValidator(IStringLocalizer<SharedResource> localizer, ICustomPasswordRules passwordRules) : IPasswordValidator<User>
```
VadymLevkovskyi, line 5: Primary constructor! Thanks for using it )

OutOfSchool/OutOfSchool.AuthServer.Tests/Validators/CustomPasswordValidationAttributeTests.cs
DKymachynskyi: Fixed. Thank you!

VadymLevkovskyi (approved):

PR #1427, opened 2024-03-22
Fix notifications for deleting workshop

alrom:


VadymLevkovskyi (changes requested):
There are 2 issues in suggested PR:
- (simpler one) lack of tests 
- (and bigger one) brittle design.

---

_(Tests)_ You are not checking your changes, as I get them you are triing to preserve notification recipients list during deletion operation. Neither test checks this.
 
--- 

_(Design issue)_ It was "complicated" before your changes and get even more complicated after. 

How notification sending worked before? 

Some service asks notification service to send the notification (`notificationService.Create()`) and passes there **itself** (same logic in all services) as notification recipient (`INotificationReceiver`) that already looks strange by naming/logic (I'm asking someone to send message to... myself?). The mentioned `INotificationReceiver` contains only one method (with 3 parameters) that returns `recipient ids`, but not all implementations use all 3 parameters, thus making this API bit ill-formed.

Below, there is short schema for beforementioned workflow:
- `service` calls 
- `notificationService.Create(..., service, ...)` that calls 
- `(service as INotificationReceiver).GetNotificationsRecipientIds(partly_ignored_params)` that returns 
- `ids` that are used by 
- `notificationService`.

Few issue could be found here: 
- 2-directional dependency between `service` and `notificationService` (or, in other words, excessive coupling) due to that `INotificationReceiver.GetNotificationsRecipientIds()` call (pay attention, that solely adding `INotificationReceiver` doesn't decouple this dependency)
- already mentioned bad naming - sending service is always an `INotificationReceiver` -  receiver?!
- ignored arguments in some implementations of `GetNotificationsRecipientIds()`, bad due to broken encapsulation (as you need to read the code to understand the abstraction), and lower maintainability (when next service be added and if it require more arguments - all other implementations should be fixed as well, refactoring of one service is blocked by others - brittle design).

After your changes another "code smells" were added:
- ["temporal coupling"](https://blog.ploeh.dk/2011/05/24/DesignSmellTemporalCoupling/) - you created private field `notificationsRecipientIds` that *should* be filled before the usage (i.e., the `Delete()` method), 
- this field scope is bigger than the problem it fixes (problem of `Delete()` but scope is - whole class) , thus it could be reused for other class needs
- as `notificationsRecipientIds` is not a local variable it could be accessed concurrently and without any locking there is a risk of wrong sequence of calls and unpredictable results.

---

I believe that all these things could be easily improved. Please check as a proof of concept some code changes I've done in 20 mins - they are not prod-ready, just to illustrate the idea and scope of changes [Remove_INotificationReceiver](
https://github.com/ita-social-projects/OoS-Backend/tree/experiment/Remove_INotificationReceiver). Pay attention, that I've not tested it (lack of unit test ;) ) so don't 100% trust that code.

VadymLevkovskyi:
> There are 2 issues in suggested PR: ....

Just one issue has left - tests! 

But pay attention, that you have to check both your changes in `WorkshopCombiner...Delete()`, and additionally all that code inside now private `GetNotificationsRecipientIds()` (I presume by using `notifyService.Setup(...)`/`.Verify(...)` and specifiing real collection of ids instead of `It.IsAny<IEnumerable<string>>()`)

alrom:
I added changes

VadymLevkovskyi:
> I added changes

Thank you. 
Please add changes, that are out of PR scope in separate PR (that extracting of `Status` string constant).
Please check Sonar coverage report to elaborate what extra tests has to be written.

VadymLevkovskyi:
Few branches in `ApplicationService` are still uncovered - see screenshot 

image

alrom:
I see. I am doing it right now ))))

alrom:
Finished.
About Status constant - I wrote it because Sonar quality control told me about it in this PR.

VadymLevkovskyi (approved):

PR #1429, opened 2024-03-25
Show for production field added

ZdorovenkoKyrylo:


VadymLevkovskyi:
@ZdorovenkoKyrylo could you specify where is logic using this new field?

ZdorovenkoKyrylo:
> @ZdorovenkoKyrylo could you specify where is logic using this new field?

This field was added earlier. But colleague from FrontEnd Team asked to check whether it does not come on Front. I added it in FeatureManagement and now it is ok

ZdorovenkoKyrylo:
> @ZdorovenkoKyrylo could you specify where is logic using this new field?

This field used in appsettings in project to show what is necessary

VadymLevkovskyi (approved):
@ZdorovenkoKyrylo thank you for the explanation. As I get it, whole related logic is hidden in Asp.Net config reading capabilities, `FeatureGate` attribute and on UI. 

Thus nothing to write / test on the backend. Ofc we could introduce related integration tests, but it seems out of scope

PR #1408, opened 2024-03-04
Kymachynskyi/ Add a short title to the workshop with the ability to search by this field

DKymachynskyi:

PR #1430, opened 2024-03-27
Vladp/change phone number format

MrVenik:
## Summary
Update the phone number format from `XX-XXX-XX-XX`, where each `X` represents a digit, to the international format `+CCC-XX-XXX-XX-XX`, where `C` denotes the country code. The new format supports phone numbers with lengths ranging from 7 (the shortest possible) to 15 digits.

## Remarks
I decided to create my own validation attribute because the existing one relied on regular expressions, which could occasionally be slow. Instead of using the [default implementation](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.phoneattribute?view=net-8.0), I benchmarked my approach against other implementations, including regular expressions and the default method. My custom variant proved to be faster, so I decided to go with it.

**Benchmark  source code:** https://github.com/MrVenik/PhoneNumberValidation
**Benchmark results:** 
image

### DIFF
--- a/OutOfSchool/OutOfSchool.AdminInitializer/AdminInitializer.cs
+++ b/OutOfSchool/OutOfSchool.AdminInitializer/AdminInitializer.cs
@@ -2,7 +2,6 @@ using Microsoft.AspNetCore.Identity;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Options;
 using OutOfSchool.AdminInitializer.Config;
-using OutOfSchool.Common;
 using OutOfSchool.Services;
 using OutOfSchool.Services.Models;
 
@@ -36,7 +35,7 @@ internal class AdminInitializer
             LastName = adminConfiguration.LastName,
             MiddleName = adminConfiguration.MiddleName,
             Email = adminConfiguration.Email,
-            PhoneNumber = Constants.PhonePrefix + adminConfiguration.PhoneNumber,
+            PhoneNumber = adminConfiguration.PhoneNumber,
             CreatingTime = DateTimeOffset.UtcNow,
             Role = adminConfiguration.Role,
             IsRegistered = true,
--- a/OutOfSchool/OutOfSchool.AdminInitializer/appsettings.json
+++ b/OutOfSchool/OutOfSchool.AdminInitializer/appsettings.json
@@ -24,7 +24,7 @@
     "MiddleName": "Прикладович",
     "Password": "replace_me",
     "Email": "admin@gmail.com",
-    "PhoneNumber": "671234567",
+    "PhoneNumber": "+380671234567",
     "Role": "techadmin",
     "Reset": false
   },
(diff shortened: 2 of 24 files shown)


MrVenik:
Migrations include many instances of `migrationBuilder.AlterColumn<DateTimeOffset>`.
I’m not sure why `dotnet-ef migrations add` inserts this code into the migration.
If it’s a critical issue, we can consider reverting the last commit.

VadymLevkovskyi:
> Migrations include many instances of `migrationBuilder.AlterColumn<DateTimeOffset>`. I’m not sure why `dotnet-ef migrations add` inserts this code into the migration. If it’s a critical issue, we can consider reverting the last commit.

This change (from datetime to datetime(6)) seems to be a correct one (just introduced in wrong PR), the only thing I'm worriing is dataloss, have you checked whether datetime data remains correct after migration?

The same issue is also related to your changes done for phone. But thy are more "broken". Before changes "380" was prefixed during phone number mapping (in code), now that part of code is removed and thus user will see old numbers "cropped". 

Please consider if old phone numbers data should be changed as a part of migration up/down.

MrVenik:
> > Migrations include many instances of `migrationBuilder.AlterColumn<DateTimeOffset>`. I’m not sure why `dotnet-ef migrations add` inserts this code into the migration. If it’s a critical issue, we can consider reverting the last commit.
> 
> This change (from datetime to datetime(6)) seems to be a correct one (just introduced in wrong PR), the only thing I'm worriing is dataloss, have you checked whether datetime data remains correct after migration?
> 
> The same issue is also related to your changes done for phone. But thy are more "broken". Before changes "380" was prefixed during phone number mapping (in code), now that part of code is removed and thus user will see old numbers "cropped".
> 
> Please consider if old phone numbers data should be changed as a part of migration up/down.

I am not sure about data loss, but based on my understanding and the information in the [MySQL documentation](https://dev.mysql.com/doc/refman/8.3/en/datetime.html), the `datetime(6)` data type offers greater precision than the standard `datetime`. Therefore, there should be no risk of data loss. The same migration issue can be found in #1408.

As for phone numbers in the database that are not prefixed with `+CCC`, we discussed this during our daily meeting. The decision was to manually add the appropriate prefix to the existing phone numbers in the database.

MrVenik:
> > Migrations include many instances of `migrationBuilder.AlterColumn<DateTimeOffset>`. I’m not sure why `dotnet-ef migrations add` inserts this code into the migration. If it’s a critical issue, we can consider reverting the last commit.
> 
> This change (from datetime to datetime(6)) seems to be a correct one (just introduced in wrong PR), the only thing I'm worriing is dataloss, have you checked whether datetime data remains correct after migration?
> 
> The same issue is also related to your changes done for phone. But thy are more "broken". Before changes "380" was prefixed during phone number mapping (in code), now that part of code is removed and thus user will see old numbers "cropped".
> 
> Please consider if old phone numbers data should be changed as a part of migration up/down.

Now migration does not include any `datetime` changes.

MrVenik:
@VadymLevkovskyi Can you review and provide feedback? Thank you!

VadymLevkovskyi (approved):

PR #1436, opened 2024-04-01
Add validation exists Edrpou and Emails for providers

alrom:


OutOfSchool/OutOfSchool.DataAccess/Repository/ProviderRepository.cs
```diff
@@ -103,4 +103,20 @@ public Task<int> CountWithDeleted(DateTime updatedAfter)
 
         return query.CountAsync();
     }
+
+    public async Task<List<string>> CheckExistsByEdrpous(List<string> edrpous)
```
VadymLevkovskyi: (Performance) Please avoid using excessive `async`/`await`. Please fix both `Check...` methods.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ProviderServiceTests.cs
```diff
@@ -1064,6 +1064,49 @@ public async Task Block_ReturnsNull_IfNotAdmin()
...
+        data.Edrpous.Add("test");
+        data.Emails.Add("test");
+
+        // Act
+        var result = await providerService.ValidateImportData(data).ConfigureAwait(false);
+
+        // Assert
+        providersRepositoryMock.Verify(x => x.CheckExistsByEdrpous(data.Edrpous), Times.Once);
+        providersRepositoryMock.Verify(x => x.CheckExistsByEmails(data.Emails), Times.Once);
+    }
```
VadymLevkovskyi: (Test atomicity) Please consider adding to this test 2 more. 

In your current implementation of `ValidateImportData()` the `CheckExistsByEdrpous()` and `CheckExistsByEmails()` are called independently. So there should be independent tests as well. 

One existing - when both fields are set.

One with only `Edrpous` set, with assertion like
```
providersRepositoryMock.Verify(x => x.CheckExistsByEdrpous(data.Edrpous), Times.Once);
providersRepositoryMock.Verify(x => x.CheckExistsByEmails(It.IsAny<List<string>>))), Times.Never);
```

And one more with only `emails` set, with assertion like
```
providersRepositoryMock.Verify(x => x.CheckExistsByEdrpous(It.IsAny<List<string>>()), Times.Never);
providersRepositoryMock.Verify(x => x.CheckExistsByEmails(data.Emails), Times.Once);
```

VadymLevkovskyi (changes requested):
Logic seems ok. There are minor issues with async/await and tests. See other ocomments

OutOfSchool/OutOfSchool.DataAccess/Repository/ProviderRepository.cs
alrom: Done

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ProviderServiceTests.cs
alrom: Done

VadymLevkovskyi (approved):

PR #1446, opened 2024-04-10
Kymachynskyi/ Fix city name with apostrophe is not found bug

DKymachynskyi:


OutOfSchool/OutOfSchool.DataAccess/Repository/CodeficatorRepository.cs
```diff
@@ -36,6 +38,8 @@ public async Task<List<CodeficatorAddressDto>> GetFullAddressesByPartOfName(stri
     {
         int cityAmountIfNamePartIsEmpty = 100;
 
+        namePart = namePart.RemoveCharsByRegexPattern(Constants.ExceptUkrainianLettersDashSpaceRegex);
```
VadymLevkovskyi: (Performance) Please consider usage of compiled regexp (either static instance created with `RegexOptions.Compiled` or using code generation via `GeneratedRegex` attribute) OR get rid of regular expression (simple `StringBuilder` filled by `for` cycle).

DKymachynskyi: I’ve made benchmarks:
image
The fastest method is `StringBuilder`, but it consumes much more memory than other.
We discussed with @DmyMi what approach we should use – fastest execution or low memory usage.
He suggested using the lowest memory usage approach because when we have a lot of users searching for their settlement, high memory usage could be a bottleneck.
Following this suggestion we should use `GeneratedRegex`. 
Should I create a separate static partial class for regular expressions that use `GeneratedRegex` attribute?
Thank you!

VadymLevkovskyi: Regarding **memory allocation and performance**. 
You could use `stackalloc`, that is always faster than string builder and for small strings doesn't allocate memory - see screenshot with code and benchmark. At same time I don't insist, compiled regexp could be used as well

image

Regarding **place of that generated regexp**. "Exclude everything except space, Ukrainian chars and dash" looks to me as a pretty generic task, that could be reused, so could be part of `StringExtensions`. 
If you believe that it has to be part of `CodeficatorRepository` you could move it there. 
And I see no necessity to create separate class just for this method.

OutOfSchool/OutOfSchool.DataAccess/Repository/CodeficatorRepository.cs
```diff
@@ -46,9 +50,9 @@ public async Task<List<CodeficatorAddressDto>> GetFullAddressesByPartOfName(stri
                         && (string.IsNullOrEmpty(namePart) && (parentId == 0) &&
                             !(categories.Contains(CodeficatorCategory.SpecialStatusCity.Name) || categories.Contains(CodeficatorCategory.Region.Name))
                        ? EF.Property<bool>(e, "IsTop")
-                       : ((e.Name.StartsWith(namePart) &&
+                       : ((e.Name.Replace('’'.ToString(), string.Empty).StartsWith(namePart) &&
```
VadymLevkovskyi: (Code style) Just reminding you, that there is shorter variant for `'’'.ToString()` - this one `"’"`. Could be changed here and below.

VadymLevkovskyi:
(Tests) Coverage

OutOfSchool/OutOfSchool.DataAccess/Repository/CodeficatorRepository.cs
DKymachynskyi: Fixed. Thank you!

VadymLevkovskyi (approved):

PR #1450, opened 2024-04-11
Fix quartz registration for Authorization server

olsavchenk:


OutOfSchool/OutOfSchool.AuthorizationServer/Extensions/QuartzExtension.cs
```diff
@@ -55,6 +56,14 @@ public static class QuartzExtension
 
         services.AddQuartzServer(options => { options.WaitForJobsToComplete = true; });
 
+        services.AddSingleton<IScheduler>(provider =>
```
h4wk13: is it possible to make this lambda `async` to avoid calling to `.Result` ?

h4wk13 (approved):

OutOfSchool/OutOfSchool.AuthorizationServer/Extensions/QuartzExtension.cs
olsavchenk: As I discovered there are 2 varients of how we can change it:
1. Make extension method async, create instance of scheduler and pass is to Singleton registation
2. Change it to 
`services.AddSingleton<IScheduler>(provider =>
    {
        var schedulerFactory = new StdSchedulerFactory();
        var scheduler = schedulerFactory.GetScheduler().GetAwaiter().GetResult();
        scheduler.Start().GetAwaiter().GetResult();
        return scheduler;
    });`
    
I'm not sure which one is better, as they both seem pretty wrong to me

VadymLevkovskyi: Quartz [documentation example](https://www.quartz-scheduler.net/documentation/quartz-3.x/tutorial/using-quartz.html#configure-program-cs) doesn't ask to register the scheduler as a dependency (and I see no mentions of resolving `IScheduler` in the code). 
Instead you could just create it and run it in `Program.cs` before `builder.Build()`.

PR #1452, opened 2024-04-12
add Microsoft unobtrusive validation

DmyMi:
- fix regex for names so it can be used both in .Net & JS.
- Add aspnet unobtrusive validation to dynamically use .Net DataAnnotations in JS client (with custom adapter for custom password validation).


DmyMi:
@VadymLevkovskyi untested code is only JS code, we don't have JS tests setup yet so can't do 80% coverage :(

VadymLevkovskyi (approved):

PR #1459, opened 2024-04-18
Kymachynskyi/ Fix password generation

DKymachynskyi:


OutOfSchool/OutOfSchool.AuthCommon/Services/Password/PasswordGenerator.cs
```diff
@@ -1,65 +1,50 @@
...
-        if (passwordOptions.RequireLowercase)
+        while (password.Count < Constants.PasswordMinLength)
         {
-            chars.Insert(
-                rand.Next(0, chars.Count),
-                randomChars[1][rand.Next(0, randomChars[1].Length)]);
+            var randomCharSetIndex = GetRandomInt32(allowedCharSets.Length);
+            var randomChar =
+                allowedCharSets[randomCharSetIndex][
+                GetRandomInt32(allowedCharSets[randomCharSetIndex].Length)];
```
VadymLevkovskyi: (Computational hardness) Here you loose the true "randomness". Now you are taking a random character set and then taking random character in the set. 

Imagine, that you have 2 sets: the very small one (1 character in extreme) and the very big one (100 characters). In your scenario - that 1 character from the first set would be taken with 50% probablity, while characters from bigger set would be taken with 0.5% probability. This is pretty different probability.

Please consider concatenating all sets into 1 string and taking random char from it.

OutOfSchool/OutOfSchool.AuthCommon/Services/Password/PasswordGenerator.cs
```diff
@@ -1,65 +1,50 @@
...
+
+            if (!password.Contains(randomChar))
+            {
+                password.Add(randomChar);
+            }
```
VadymLevkovskyi: (Computational hardness) Please explain why we can't have duplicated symbols. Now you are loosing "randomness" here. 

Details. For case of `n` allowed chars for password having length `m`, where `m<n`, you schema provides `n!/(n-m)!` different passwords, while schema where duplications are allowed provides `n^m` paswords - the latter is the higher number.

OutOfSchool/OutOfSchool.AuthCommon/Services/Password/PasswordGenerator.cs
```diff
@@ -1,65 +1,50 @@
...
+            Constants.ValidationSymbols,
+        ];
 
-        CryptoRandom rand = new CryptoRandom();
+        List<char> password = [];
 
-        List<char> chars = new List<char>();
-
-        if (passwordOptions.RequireUppercase)
+        foreach (var charSet in allowedCharSets)
```
VadymLevkovskyi: (Computational hardness) Here you loose the true "randomness". The algorithm makes first 4 characters to strictly be from different sets. 

Please consider adding these 4 characters in random places after the initiating password in the `while (password.Count < Constants.PasswordMinLength)` loop.

OutOfSchool/OutOfSchool.AuthServer.Tests/Services/Password/PasswordGeneratorTests.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    [Test]
+    public void PasswordGenerator_ShouldReturnValidPassword()
+    {
+        // Act
+        var result = PasswordGenerator.GenerateRandomPassword();
+
+        // Assert
+        Assert.IsNotNull(result);
+        Assert.That(result, Does.Match(@"^(?=.*[A-Z])(?=.*[a-z])(?=.*\d)(?=.*[@$!%*?&])[A-Za-z\d@$!%*?&]{8}$"));
```
VadymLevkovskyi: 2 issues:
(Readability) This regexp is as cool as complicated. Unit test code should be simple. 

Please consider writing instead of 1 complicated 6 simple asserts: 4 checking that characters from each charset are present in the password, and 1 checking that no other characters are present, and one checking password length. 
It will improve readablility and make subsequent changes easier (just imagine adding 5th checked character set, or changing the existing one).

(Logic) Is it strict requirement that generated password length should be `8`? Or should it be `>= 8`? 
If latter, please change the code.

VadymLevkovskyi (changes requested):
See other comments. Main reason for requesting changes - ineffective "random" algorithm.

OutOfSchool/OutOfSchool.AuthCommon/Services/Password/PasswordGenerator.cs
DKymachynskyi: I agree with your argument, but in this case the password must include at least one character from each set.

OutOfSchool/OutOfSchool.AuthCommon/Services/Password/PasswordGenerator.cs
DKymachynskyi: I was wrong in thinking that it would be better if the password didn't have duplicate symbols. Thank you!

OutOfSchool/OutOfSchool.AuthCommon/Services/Password/PasswordGenerator.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.AuthServer.Tests/Services/Password/PasswordGeneratorTests.cs
DKymachynskyi: The length of the generated password is equal to the PasswordMinLength constant.
Fixed. Thank you!

MrVenik:
Sorry for the interruption, but I have a few suggestions about password generation. Firstly, instead of using a `List`, for creating small strings, consider using `stackalloc` or a string creation method that accepts a delegate (as shown below), or if `List` approach is preferred, consider construct `List` with maximum possible capacity, to reduce number of reallocations. Secondly, for improved randomness, you can use the `GetItems` and `Shuffle` methods from the `RandomNumberGenerator` class (as demonstrated below).

```csharp
Span<char> password = stackalloc char[passwordLength];

var index = 0;

// fill required characters
foreach (var charSet in CharSets)
{
    var rndIndex = RandomNumberGenerator.GetInt32(charSet.Length);
    var rndChar = charSet[rndIndex];

    password[index++] = rndChar;
}

// fill remaining characters
RandomNumberGenerator.GetItems(AllPasswordCharacters, password[index..]);

// shuffle to achieve better randomness
RandomNumberGenerator.Shuffle(password);

return new string(password);
```

Or you can even fill remaining characters and shuffle in one place
```csharp
return string.Create(passwordLength, 0, static (span, count) =>
{
    // fill required characters
    foreach (var charSet in CharSets)
    {
        span[count++] = charSet[RandomNumberGenerator.GetInt32(charSet.Length)];
    }

    // fill remaining characters and swap replace them in random place
    while (count < span.Length)
    {
        span[count] = AllPasswordCharacters[RandomNumberGenerator.GetInt32(AllPasswordCharacters.Length)];

        var rndIndex = RandomNumberGenerator.GetInt32(count);

        (span[rndIndex], span[count]) = (span[count], span[rndIndex]);

        count++;
    }
});

```

Feel free to use this code as you want and of course these are only suggestions and I do not force you to implement this method as above.

DKymachynskyi:
@MrVenik in one of the code variants I used `span<char>` - `stackalloc`, but yours is more concise. Thank you!

MrVenik:
> @MrVenik in one of the code variants I used `span<char>` - `stackalloc`, but yours is more concise. Thank you!

Personally, I would prefer the second variant with `string.Create(...)`, because it would allocate memory only once and avoid performing memory copying.

OutOfSchool/OutOfSchool.AuthCommon/Services/Password/PasswordGenerator.cs
VadymLevkovskyi: > the password must include at least one character from each set.

... and you get this in `foreach` loop. That's why there is no necessity to do this here in `while` loop. Moreover, due to randomness theer is (small) probability to get same charset for one password generation in this `while` loop, so your argument is bit irrelevant. 

Please reconsider your comment.

OutOfSchool/OutOfSchool.AuthServer.Tests/Services/Password/PasswordGeneratorTests.cs
VadymLevkovskyi: Thanks for fixing. One last touch of perfection could be made if you add message in your `Assert`s. To get better explanation for people reading the test code and failed test message.

Like 
```
Assert.IsFalse(result.Any(c => !char.IsLetterOrDigit(c) && !Constants.ValidationSymbols.Contains(c)), 
    "Password shouldn't contain any other symbols than digits, upper/lowercase letters, and allowed special symbols");
```

OutOfSchool/OutOfSchool.AuthCommon/Services/Password/PasswordGenerator.cs
DKymachynskyi: @VadymLevkovskyi You are right. I fixed this in my latest commit. Thank you!

OutOfSchool/OutOfSchool.AuthServer.Tests/Services/Password/PasswordGeneratorTests.cs
DKymachynskyi: I've added message. Thank you!

VadymLevkovskyi:
Guys, @DKymachynskyi and @MrVenik, let me first thank you for thinking about performance. 

One small comment - don't trust you feelings how code should work, therefore please benchmark your changes (and share it), and don't overcomplicate the code to slightly increase the performance. 

I've performed some measurements locally (see screen) and it looks that **allocated memory** for both `stackalloc` and `string.Create` cases is the same. 

From **CPU** standpoint, `stackalloc` is worst (I believe that `RandomNumberGenerator.Shuffle()` is guilty), but it's more readable (and here `RandomNumberGenerator.Shuffle()` really shines, while `(span[rndIndex], span[count]) = (span[count], span[rndIndex])` is a risky move). 

Initially, in this comment I had an intention to show that `string.Create` is obviously the winner (with benchmark) and ask to not overcomplicate the code in other cases, when it's not so obvious. Especially for such infrequent cases as password generation.

But then, I've realized, that "shuffle" could be "fixed" in `stackalloc` version - the `GenStackallocManualShuffle()` - that is the new winner... or not (4% of CPU time is not a big deal for this case). That's why I don't insist on using this version, just asking you to be precise in your experiments, benchmarking and measuring what is really the fastest.

image

DKymachynskyi:
@VadymLevkovskyi I've run tests for all my code variations in this task (but haven't shared them), but they're slower than your `GenStackallocManualShuffle()`, so I'm choosing that one.
Should I write additional tests to achieve 80% coverage?
Thank you!

VadymLevkovskyi:
> @VadymLevkovskyi I've run tests for all my code variations in this task (but haven't shared them), but they're slower than your `GenStackallocManualShuffle()`, so I'm choosing that one. 

I trust your benchmarks )

> Should I write additional tests to achieve 80% coverage? 

Yes, please. (Unfortunately, you'll find how awful static methods are for testing)

MrVenik:
@VadymLevkovskyi and @DKymachynskyi,

Thank you, Vadym, for sharing your measurements, and my apologies for not including benchmarks in my previous comment.

Regarding the `GenStackalloc` method, the performance drawback is indeed caused by the `RandomNumberGenerator.Shuffle()` operation. Under the hood, this method performs additional iterations through the entire string for the extra shuffle. The implementation is relatively straightforward: it swaps items by replacing an item with another at a random index.

Now, let’s compare `stackalloc` and `string.Create`. While you won’t observe these “extra” allocations directly (since they occur on the stack), consider scenarios where you’re developing a public API for a library. If callers need to create a random password with a large length (> 1KB), using `stackalloc` isn’t usable. Instead, you might create a new array or rent an array from `ArrayPool`. In such cases, `string.Create` shines because it avoids heap allocations. As for memory copying, the `new string(ReadOnlySpan)` constructor doesn’t merely use the span; it actually copies memory from the span into newly allocated memory on the heap.

Lastly, for experimental purposes, I’ve created an alternative approach to password generation. This method is slightly more complex. In summary, it generates a password with a length ranging from `MinPassLength` to `MinPassLength + CharSets.Length`, inclusively. The algorithm is straightforward: generate the password, check for each required character, and add a random required character if necessary. For the checking existence of characters this approach uses `SearchValues` class, which is really powerful and can be used everywhere where you have some static set of characters and you need perform fast checking for existence (eg `PasswordValidation`, `PhoneValidation` etc).

Below, you’ll find the code and associated benchmarks:

Spans for big string:
```csharp
char[]? rented = null;
Span<char> password = maxPasswordLength < 512 ? stackalloc char[maxPasswordLength] : (rented = ArrayPool<char>.Shared.Rent(maxPasswordLength));
```

Fixed `GenStringCreate` that uses your fixes from `GenStackallocManualShuffle`:
```csharp
[Benchmark]
public string GenStringCreateFixed()
{
    return string.Create(PasswordMinLength, PasswordMinLength - CharSets.Length, static (span, index) =>
    {
        RandomNumberGenerator.GetItems(AllPasswordCharacters, span[..index]);

        foreach (var charset in CharSets)
        {
            var rndIndex = RandomNumberGenerator.GetInt32(index);

            (span[rndIndex], span[index]) = (charset[RandomNumberGenerator.GetInt32(charset.Length)], span[rndIndex]);

            index++;
        }
    });
}
```

Complex approach:
```csharp
private static SearchValues<char> UppercaseAlphabetSearchValues { get; } = SearchValues.Create(UppercaseAlphabet);
private static SearchValues<char> LowercaseAlphabetSearchValues { get; } = SearchValues.Create(UppercaseAlphabet);
private static SearchValues<char> DigitsSearchValues { get; } = SearchValues.Create(Digits);
private static SearchValues<char> ValidationSymbolsSearchValues { get; } = SearchValues.Create(ValidationSymbols);

private static (string CharSet, SearchValues<char> search)[] CharSetsWithSearchers { get; } =
[
    (UppercaseAlphabet, UppercaseAlphabetSearchValues),
    (LowercaseAlphabet, LowercaseAlphabetSearchValues),
    (Digits, DigitsSearchValues),
    (ValidationSymbols, ValidationSymbolsSearchValues),
];

[Benchmark]
public string GenComplex()
{
    var maxPasswordLength = PasswordMinLength + CharSetsWithSearchers.Length;

    Span<char> password = stackalloc char[maxPasswordLength];

    var currentLength = PasswordMinLength;

    RandomNumberGenerator.GetItems(AllPasswordCharacters, password[..currentLength]);

    foreach (var (charSet, searchValues) in CharSetsWithSearchers)
    {
        if (password[..currentLength].IndexOfAny(searchValues) is -1)
        {
            var rndIndex = RandomNumberGenerator.GetInt32(currentLength);

            (password[rndIndex], password[currentLength]) = (charSet[RandomNumberGenerator.GetInt32(charSet.Length)], password[rndIndex]);

            currentLength++;
        }
    }

    return new string(password[..currentLength]);
}
```

Benchmarks on my PC:
image

And, of course, a huge thanks for the interesting investigations and experiments!

OutOfSchool/OutOfSchool.AuthServer.Tests/Services/CommonMinistryAdminServiceTests.cs
```diff
@@ -122,4 +180,34 @@ private async Task Seed()
...
+            InstitutionId = Guid.NewGuid(),
+            CATOTTGId = 28675,
+        };
+
+        var areaAdmin = new AreaAdmin
+        {
+            InstitutionId = areaAdminBaseDto.InstitutionId,
+            User = user,
+            UserId = string.Empty,
+        };
```
VadymLevkovskyi: (Test design) I believe we have related test data generators. Please consider using them (and create one if any is missing).

OutOfSchool/OutOfSchool.AuthServer.Tests/Services/CommonMinistryAdminServiceTests.cs
```diff
@@ -21,35 +24,87 @@
...
+        fakeUserManager = new Mock<UserManager<User>>(
+             new Mock<IUserStore<User>>().Object,
+             new Mock<IOptions<IdentityOptions>>().Object,
+             new Mock<IPasswordHasher<User>>().Object,
+             new IUserValidator<User>[0],
+             new IPasswordValidator<User>[0],
+             new Mock<ILookupNormalizer>().Object,
+             new Mock<IdentityErrorDescriber>().Object,
+             new Mock<IServiceProvider>().Object,
+             new Mock<ILogger<UserManager<User>>>().Object);
```
VadymLevkovskyi: (Test design) By using mocks created in constructor you are introducing state shared between tests. Slightly better performance, worse side-effects. It's common approach to not do this. So, please consider moving all initialization in `SetUp()`. 
Additionally consider removing private fields that are not used anywhere except `SetUp()` (like `fakeEmailSender`, etc.).

OutOfSchool/OutOfSchool.AuthServer.Tests/Services/CommonMinistryAdminServiceTests.cs
```diff
@@ -1,18 +1,21 @@
 ﻿using AutoMapper;
 using FluentAssertions;
 using Microsoft.AspNetCore.Identity;
-using Microsoft.AspNetCore.Identity.EntityFrameworkCore;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Routing;
```
VadymLevkovskyi: (Excessive usings) Please double-check if all usings here are necessary

OutOfSchool/OutOfSchool.AuthServer.Tests/Services/CommonMinistryAdminServiceTests.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.AuthServer.Tests/Services/CommonMinistryAdminServiceTests.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.AuthServer.Tests/Services/CommonMinistryAdminServiceTests.cs
DKymachynskyi: Fixed. Thank you!

VadymLevkovskyi (approved):

PR #1461, opened 2024-04-19
Fix permissions get

ZdorovenkoKyrylo:


VadymLevkovskyi (approved):
Approved, though I'm not sure why db snapshot should be part of this PR

PR #1463, opened 2024-04-22
Remove redundant fields from register and age validation

MrVenik:
Removed redundant fields (Gender, BirthDate, PhoneNumber) from registration, as these fields will be set later.
Implemented age validation.


OutOfSchool/OutOfSchool.AuthCommon/Controllers/AuthController.cs
```diff
@@ -397,13 +395,13 @@ public async Task<IActionResult> Register(RegisterViewModel model)
...
-                    if (user.Role == Role.Parent.ToString().ToLower())
+                    if (user.Role.Equals(nameof(Role.Parent), StringComparison.OrdinalIgnoreCase))
                     {
                         var parent = new Parent()
                         {
                             UserId = user.Id,
-                            Gender = model.Gender,
-                            DateOfBirth = model.DateOfBirth,
+                            Gender = Gender.Male,
+                            DateOfBirth = DateTime.Now.AddYears(-18),
```
VadymLevkovskyi: (Code clarity) This is bit unclear why all parents are 18 yrs old males, until you read the PR description. 
Later, in the plain code, there will be no PR description, thus it would be even more obscure. Please consider a way to shed more light here.

In example:
- either put a comment that you are creating some default parent with specific id and other valid fileds 
- or create helper method in `Parent` class with appropriate name like `CreateDefaultWithUserId(Guid userId)`

As for me latter is better, because of encapsulation (better reuse, refactoring, testing, search), and because I don't like comments in the code (there are a few reasons)

VadymLevkovskyi: (Code clarity) "Magic numbers" is a code smell. Here you have an `18`! Please consider using `AdultAge` constant (you've created)

OutOfSchool/OutOfSchool.Common/Validators/CustomAgeAttribute.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        if (possibleBirthDate is not DateTime birthDate)
+        {
+            return false;
+        }
+
+        var totalDays = (currentDate - birthDate).TotalDays;
+        var approximateAge = (int)(totalDays / ApproximateDaysInYear);
+
+        return approximateAge >= MinAge && approximateAge <= MaxAge;
```
VadymLevkovskyi: (DateTime logic) In general manipulating with datetime as numbers is not a correct way. 

The same task you have could be solved using `DateTime.AddYear`, like
```
currentDate = currentDate.Date;
return currentDate < birthDate.AddYears(MaxAge+1) && currentDate >= birthDate.AddYears(MinAge);
```

This way we get rid of magical `ApproximateDaysInYear = 365.2425` (that is correct for long periods, but wrong for short - imagine period 2097..2103 - 7 years with no leap year, having 7x365 days that is 1,... day less than 7x365.2425 days)
Also there is no time of the day manipulation - as noone uses it in age calculation.

OutOfSchool/OutOfSchool.Common/Validators/CustomAgeAttribute.cs
```diff
@@ -0,0 +1,99 @@
...
+
+    /// <summary>
+    /// Gets a value indicating whether to use UTC now as beginning of time or not. Default value is <c>true</c>.
+    /// </summary>
+    public bool UseUTC { get; init; } = true;
+
+    /// <summary>
+    /// Gets the date time kind based on <see cref="UseUTC"/> value.
+    /// </summary>
+    private DateTimeKind DateTimeKind => UseUTC ? DateTimeKind.Utc : DateTimeKind.Local;
```
VadymLevkovskyi: (Design) I'm not insisting on redesign, you could leave it as is.

Personally me would avoid such things
- private property - could be field or method;
- getter-only property (that could never have a setter) - could be method;
- simple private class member used in only one place - could be inlined.

OutOfSchool/OutOfSchool.WebApi.Tests/Validators/CustomAgeAttributeTests.cs
```diff
@@ -0,0 +1,177 @@
...
+{
+    [Test]
+    public void IsValid_WhenDateIsNull_ShouldReturnTrue()
+    {
+        // Arrange
+        var dateOfBirth = null as DateTime?;
+        var customDayOfBirthAttribute = new CustomAgeAttribute();
+
+        // Act
+        var isValid = customDayOfBirthAttribute.IsValid(dateOfBirth);
```
VadymLevkovskyi: (Test design) You could leave the code as it is now. I'm just suggesting improvements, but don't insist.

More code doesn't mean better code. "Arrange-Act-Assert" test pattern doesn't mean you should always have each separate part in your test. Sometimes simpler code is also easy to read. 

For this specific test you could omit "Arrange" part like 
```
        // Act
        var isValid = new CustomAgeAttribute().IsValid(null as DateTime?);
```
3 lines of code less, same meaning.

I agree, that sometimes system-under-test creation is more complex and could be done in "Arrange" part (i.e. in later tests you have `new CustomAgeAttribute() { MinAge = -1 }`), but not this time. Same for input data.

OutOfSchool/OutOfSchool.WebApi.Tests/Validators/CustomAgeAttributeTests.cs
```diff
@@ -0,0 +1,177 @@
...
+    {
+        // Arrange
+        var dateOfBirth = DateTime.UtcNow;
+        var customDayOfBirthAttribute = new CustomAgeAttribute() { MinAge = -1 };
+
+        // Act
+        TestDelegate action = () => customDayOfBirthAttribute.IsValid(dateOfBirth);
+
+        // Assert
+        Assert.Throws<InvalidOperationException>(action);
```
VadymLevkovskyi: (Test design) Creation of delegate is not an "Act", it's closer to "Arrange" or "Assert". As for me, I'll rewrite this as 
```
        // Assert
        Assert.Throws<InvalidOperationException>(() => customDayOfBirthAttribute.IsValid(dateOfBirth));
```

OutOfSchool/OutOfSchool.WebApi.Tests/Validators/CustomAgeAttributeTests.cs
```diff
@@ -0,0 +1,177 @@
...
+
+        // Assert
+        Assert.IsTrue(isValid);
+    }
+
+    [Test]
+    public void IsValid_WhenMinAgeIsNotPositive_ShouldThrowInvalidOperationException()
+    {
+        // Arrange
+        var dateOfBirth = DateTime.UtcNow;
```
VadymLevkovskyi: (Test design) It doesn't matter what specific date is passed to the test. Please state it in your code. 
I.e. `var dateOfBirth = It.IsAny<DateTime>();`.

Also, as it is simple code and not reused, you could inline it in `.IsValid()` call

VadymLevkovskyi: This comment is also related to other test where specific date is not important.

OutOfSchool/OutOfSchool.WebApi.Tests/Validators/CustomAgeAttributeTests.cs
VadymLevkovskyi: Please check other tests as well

VadymLevkovskyi (changes requested):

OutOfSchool/OutOfSchool.AuthCommon/Controllers/AuthController.cs
vpustovoitov: > (Code clarity) This is bit unclear why all parents are 18 yrs old males, until you read the PR description. Later, in the plain code, there will be no PR description, thus it would be even more obscure. Please consider a way to shed more light here.
> 
> In example:
> 
> * either put a comment that you are creating some default parent with specific id and other valid fileds
> * or create helper method in `Parent` class with appropriate name like `CreateDefaultWithUserId(Guid userId)`
> 
> As for me latter is better, because of encapsulation (better reuse, refactoring, testing, search), and because I don't like comments in the code (there are a few reasons)

I completely agree with you that creating all parents as 18-year-old males is unclear. I’ve considered a few solutions:

1. **Enum Modification:** Initially, I thought about adding a `None` case to the `Gender` enum. This would make things clearer, but it would require changing the entire existing parent structure in the database. Currently, we have `Male` as the default enum value, so introducing `None` would require adjustments. However, this approach wouldn’t fix the issue of parent age.

2. **Remove Required Attribute:** Another option is to remove the `Required` attribute from the properties within the `Parent` class. While this approach is less harmful to the database than the first one, we need to be careful when working with these properties in the future.

3. **Plug Values (current int this PR):** The simplest approach would be to set these properties to some default values (such as the current ones). However, as you rightly pointed out, this approach lacks clarity.

Given these options, should we create a factory method for the default parent inside the `Parent` class, or try implement other variants?

OutOfSchool/OutOfSchool.WebApi.Tests/Validators/CustomAgeAttributeTests.cs
vpustovoitov: > (Test design) You could leave the code as it is now. I'm just suggesting improvements, but don't insist.
> 
> More code doesn't mean better code. "Arrange-Act-Assert" test pattern doesn't mean you should always have each separate part in your test. Sometimes simpler code is also easy to read.
> 
> For this specific test you could omit "Arrange" part like
> 
> ```
>         // Act
>         var isValid = new CustomAgeAttribute().IsValid(null as DateTime?);
> ```
> 
> 3 lines of code less, same meaning.
> 
> I agree, that sometimes system-under-test creation is more complex and could be done in "Arrange" part (i.e. in later tests you have `new CustomAgeAttribute() { MinAge = -1 }`), but not this time. Same for input data.

What do you think about collapsing similar tests (like all that returns true, all that returns false, all that throws) into something like (this not valid code):

```csharp
public static object[][] ValidInputs =>
[
    [1, 2, default(object?)], // null case is valid
    [1, 3, DateTime.Now.AddYears(-2)],
    // .. etc
];

[TestCaseSource(nameof(ValidInputs))]
public void IsValid_WhenInputDataValid_ShouldReturnTrue(int minAge, int maxAge, object testValue) { }
```

TestCaseSource docs: https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcasesource.html

OutOfSchool/OutOfSchool.Common/Validators/CustomAgeAttribute.cs
vpustovoitov: > (Design) I'm not insisting on redesign, you could leave it as is.
> 
> Personally me would avoid such things
> 
> * private property - could be field or method;
> * getter-only property (that could never have a setter) - could be method;
> * simple private class member used in only one place - could be inlined.

Got it!
But one question: what is wrong with getter-only property? Same with compute-only property `int Prop => a + b` ?

OutOfSchool/OutOfSchool.WebApi.Tests/Validators/CustomAgeAttributeTests.cs
vpustovoitov: > (Test design) It doesn't matter what specific date is passed to the test. Please state it in your code. I.e. `var dateOfBirth = It.IsAny<DateTime>();`.
> 
> Also, as it is simple code and not reused, you could inline it in `.IsValid()` call

Personally, I prefer not use `It.IsAny<T>()` from `Moq` as variable assignment, I think this is considered to use only in setup the method invocation (eg `.Setup(..)`, `.Verify(..)` etc). But I can be wrong so can you explain why using `It.IsAny<DateTime>();` is preffered over using some value or even default keyword ?

It class docs: https://documentation.help/Moq/FBE0FFA5.htm

OutOfSchool/OutOfSchool.Common/Validators/CustomAgeAttribute.cs
vpustovoitov: > (DateTime logic) In general manipulating with datetime as numbers is not a correct way.
> 
> The same task you have could be solved using `DateTime.AddYear`, like
> 
> ```
> currentDate = currentDate.Date;
> return currentDate < birthDate.AddYears(MaxAge+1) && currentDate >= birthDate.AddYears(MinAge);
> ```
> 
> This way we get rid of magical `ApproximateDaysInYear = 365.2425` (that is correct for long periods, but wrong for short - imagine period 2097..2103 - 7 years with no leap year, having 7x365 days that is 1,... day less than 7x365.2425 days) Also there is no time of the day manipulation - as noone uses it in age calculation.

Refactored! I decided to work only with dates through the `DateOnly` struct and slightly rework your suggestion.

OutOfSchool/OutOfSchool.Common/Validators/CustomAgeAttribute.cs
```diff
@@ -96,4 +90,22 @@ private void EnsureThatMinAndMaxAgesAreValid()
...
+    /// Calculates min and max possible date.
+    /// </summary>
+    /// <param name="currentDate">Current date.</param>
+    /// <returns>Min and max possible dates.</returns>
+    private (DateOnly Min, DateOnly Max) CalculateMinAndMaxDate(DateOnly currentDate)
+    {
+        var maxYearOffset = currentDate.Year - 1;
+
+        var minAge = int.Clamp(MinAge, 0, maxYearOffset);
+        var maxAge = int.Clamp(MaxAge, 0, maxYearOffset);
```
VadymLevkovskyi: (Wow!) I had never expected to see person age limiting by 2+ thousand years. It's so unjustly to seniors )))

OutOfSchool/OutOfSchool.Common/Validators/CustomAgeAttribute.cs
```diff
@@ -96,4 +90,22 @@ private void EnsureThatMinAndMaxAgesAreValid()
...
+
+        var minDate = currentDate.AddYears(-maxAge);
```
VadymLevkovskyi, line 101: (Logic) I believe that `MaxAge` age should be allowed and we should disallow age starting from `MaxAge + 1`. Thus I expect here to have `.AddYears(-maxAge - 1)`. 

Ofc, this should be tested. And I've rechecked related tests and didn't find any check for age of person having tomorrow `maxAge+1` (it has to be allowed today, but denied tomorrow). Now tests are checking birthday dates precisely. That's why your current `IsValid_WhenDateIsExactlyMaxAge_ShouldReturnTrue` test is green but not covering the edge cases.

OutOfSchool/OutOfSchool.AuthCommon/Controllers/AuthController.cs
VadymLevkovskyi: Mentioned options:
1. `None` case in `enum` - has it's own downsides, like `null` for reference type, there is no compile-time check for valid/existing data, thus you either checks in code or ignore them (or forget) and have run-time issues.
2. No `Required` attribute and nullable properties - breaks encapsulation, same absence of compile-time check if `Parent` is valid, need additional `if`s to enforce this, same issues as for previous
3. Dummy values - encapsulation is better, but data are still "wrong", you can't trust age/gender data taken from any `Parent`. Compiler is happy, but you should not forget to put real data for specific `Parent` later in the code.

So either option is "bad". 

Solution could be introducing of separate `DraftParent` class (without gender and age) in addition to existing `Parent` class (with mentioned properties). Benefit - compile-time control of correctness, clear point in the code where data are converted from one type to another. Downside (generic for many "design-changing" solution) - wider changes of code, more entities to support.

That's why here I suggest to take "bad" solution (current), but at least document it correctly, using "factory" method from `Parent`. 

Ofc, you could check that additional type introducing approach, maybe it won't introduce many changes.

OutOfSchool/OutOfSchool.Common/Validators/CustomAgeAttribute.cs
VadymLevkovskyi: Thanks, looks cool, resolving this comment, but see my other comment below regarding `maxAge`.

OutOfSchool/OutOfSchool.Common/Validators/CustomAgeAttribute.cs
VadymLevkovskyi: > But one question: what is wrong with getter-only property?

TLDR; It depends ) 

Usually they work in a same way as method, but promise more, provide wider API / potential for changes and thus need to be maintained (both in implementation and usage sites) more carefully. Also semantic is different. And usage.

---

Talking about "wider API" I mean that (in pseudocode):
`public int SomeNumber => SomeCodeToGetNumber();` 
could be changed to
`public int SomeNumber { get => SomeCodeToGetNumber(); private set; }`, 
that could be changed to
`public int SomeNumber { get => SomeCodeToGetNumber(); private set => DoNothingSpecial(value); }`, 
and later in next version of code or descendants it could be changed to (_setter_ is changed)
`public int SomeNumber { get => SomeCodeToGetNumber(); private set => DoSomething(value); }`, 
or even (_public_ added)
`public int SomeNumber { get => SomeCodeToGetNumber(); public set => DoSomething(value); }`.

Consumer could left such changes unnoticed (you even don't need to recompile consumer code), but logic can by drastically and (more important) unexpectedly changed. That's why, though you usually don't need/use these extension capabilities, you still should take them into account/support them for properties. 

Methods - have more limited "API" and potential for extensions and thus easier for support/change.

---

Also there is semantical difference, as property is something that could be set (independently), it could be hidden but still "settable", while method can't be "set".

I.e. let's check imaginary `Shape` class, calculated property `public int Area => Width * Height`. 

Could you set it? In generic case - no, you should do this via related properties. Ofc, there could be cases with some logic for "set" (like for circles/squares/rectangles with fixed aspect ratio) but to make that "set" isolated and not to break class invariant it is better to either have immutable class and construct a new `Shape` with new `Width` and `Height` or (for mutable class case) have specific `SetDimensions(int width, int height)` method.

---

Usage. As a "rule of thumb" class properties are main target for de/serialization, mapping, input validation, etc. Sometimes you can't control such activities and need something as property even if it breaks beforementioned concerns.

---

So, in general it depends. But not for this specific `DateTimeKind` case.

OutOfSchool/OutOfSchool.WebApi.Tests/Validators/CustomAgeAttributeTests.cs
VadymLevkovskyi: > ... collapsing similar tests ...

It's ok, but only if they are similar - describing one test case. (In other words, if test fails for every case it should be due to problems in same piece of implementation code). Re your specific example - test named `...WhenInputDataValid...` has too many acceptance criteria and thus I would disagree with naming and implementation if it relates to naming.

I.e. - check `minAge =0, maxAge=17` for `17 years`, `17 years and 2 days`, `18 years minus 1 day` -  seems similar, but `0 years`, `5 years`, `null years` - seems less similar.

Btw, fake test data generators usually provides better way to get more test inputs for specific test case. It is not achieved in one test run, it requires better test design (more code to configure faker), but in long run it is better.

OutOfSchool/OutOfSchool.WebApi.Tests/Validators/CustomAgeAttributeTests.cs
VadymLevkovskyi: > ... why using It.IsAny<DateTime>(); is preffered over using some value or even default keyword ?

"Some" value or "default" value are specific values. When I read test code I could think that test will work only for this specific value, but not for others (especially when you use one of the most specific values for `DateTime` - `UtcNow`).

For this test `DateTime` value doesn't matter. It could be any. `It.IsAny` is one of the way to literally specify this fact in test code. 

The other way - use random date, i.e. from fake data generator.  Using of test data generators - is better approach, but I've seen more projects without them, than projects having them, that's why `It.IsAny` - is in general a cheaper way. 

For this specific project we have `Bogus` and you could use it as well. It's up to you.

Btw, there is other comment regarding `maxAge` check implementation and related test, and gap in test logic introduced by using the very specific input data.

OutOfSchool/OutOfSchool.WebApi.Tests/Validators/CustomAgeAttributeTests.cs
vpustovoitov: > > ... collapsing similar tests ...
> 
> It's ok, but only if they are similar - describing one test case. (In other words, if test fails for every case it should be due to problems in same piece of implementation code). Re your specific example - test named `...WhenInputDataValid...` has too many acceptance criteria and thus I would disagree with naming and implementation if it relates to naming.
> 
> I.e. - check `minAge =0, maxAge=17` for `17 years`, `17 years and 2 days`, `18 years minus 1 day` - seems similar, but `0 years`, `5 years`, `null years` - seems less similar.
> 
> Btw, fake test data generators usually provides better way to get more test inputs for specific test case. It is not achieved in one test run, it requires better test design (more code to configure faker), but in long run it is better.

Thanks! I have already fixed some test design, but this changes with collapsing similar tests I will leave to another PR, because as you said it requires better test design and it can lead to change whole bunch of tests that already implemented.

OutOfSchool/OutOfSchool.Common/Validators/CustomAgeAttribute.cs
vpustovoitov: > (Logic) I believe that `MaxAge` age should be allowed and we should disallow age starting from `MaxAge + 1`. Thus I expect here to have `.AddYears(-maxAge - 1)`.
> 
> Ofc, this should be tested. And I've rechecked related tests and didn't find any check for age of person having tomorrow `maxAge+1` (it has to be allowed today, but denied tomorrow). Now tests are checking birthday dates precisely. That's why your current `IsValid_WhenDateIsExactlyMaxAge_ShouldReturnTrue` test is green but not covering the edge cases.

Oh I got it, but I think that you mean something like `.AddYears(-maxAge).AddDays(-1)`, because if add one year to `MaxAge` then we allow every date that ends up to `MaxAge + 1` which is not correct, but I can be wrong.

As for day before `MinAge` and day after `MaxAge` I will fix this and create tests for these cases.

OutOfSchool/OutOfSchool.Common/Validators/CustomAgeAttribute.cs
vpustovoitov: > > But one question: what is wrong with getter-only property?
> 
> TLDR; It depends )
> 
> Usually they work in a same way as method, but promise more, provide wider API / potential for changes and thus need to be maintained (both in implementation and usage sites) more carefully. Also semantic is different. And usage.
> 
> Talking about "wider API" I mean that (in pseudocode): `public int SomeNumber => SomeCodeToGetNumber();` could be changed to `public int SomeNumber { get => SomeCodeToGetNumber(); private set; }`, that could be changed to `public int SomeNumber { get => SomeCodeToGetNumber(); private set => DoNothingSpecial(value); }`, and later in next version of code or descendants it could be changed to (_setter_ is changed) `public int SomeNumber { get => SomeCodeToGetNumber(); private set => DoSomething(value); }`, or even (_public_ added) `public int SomeNumber { get => SomeCodeToGetNumber(); public set => DoSomething(value); }`.
> 
> Consumer could left such changes unnoticed (you even don't need to recompile consumer code), but logic can by drastically and (more important) unexpectedly changed. That's why, though you usually don't need/use these extension capabilities, you still should take them into account/support them for properties.
> 
> Methods - have more limited "API" and potential for extensions and thus easier for support/change.
> 
> Also there is semantical difference, as property is something that could be set (independently), it could be hidden but still "settable", while method can't be "set".
> 
> I.e. let's check imaginary `Shape` class, calculated property `public int Area => Width * Height`.
> 
> Could you set it? In generic case - no, you should do this via related properties. Ofc, there could be cases with some logic for "set" (like for circles/squares/rectangles with fixed aspect ratio) but to make that "set" isolated and not to break class invariant it is better to either have immutable class and construct a new `Shape` with new `Width` and `Height` or (for mutable class case) have specific `SetDimensions(int width, int height)` method.
> 
> Usage. As a "rule of thumb" class properties are main target for de/serialization, mapping, input validation, etc. Sometimes you can't control such activities and need something as property even if it breaks beforementioned concerns.
> 
> So, in general it depends. But not for this specific `DateTimeKind` case.

Wow thanks! I never thought about such problems with properties, as for my PR I already removed `DateTimeKind`.

OutOfSchool/OutOfSchool.WebApi.Tests/Validators/CustomAgeAttributeTests.cs
vpustovoitov: > > ... why using It.IsAny(); is preffered over using some value or even default keyword ?
> 
> "Some" value or "default" value are specific values. When I read test code I could think that test will work only for this specific value, but not for others (especially when you use one of the most specific values for `DateTime` - `UtcNow`).
> 
> For this test `DateTime` value doesn't matter. It could be any. `It.IsAny` is one of the way to literally specify this fact in test code.
> 
> The other way - use random date, i.e. from fake data generator. Using of test data generators - is better approach, but I've seen more projects without them, than projects having them, that's why `It.IsAny` - is in general a cheaper way.
> 
> For this specific project we have `Bogus` and you could use it as well. It's up to you.
> 
> Btw, there is other comment regarding `maxAge` check implementation and related test, and gap in test logic introduced by using the very specific input data.

Thanks, in my case I would consider using some fake generated date from `Bogus` rather than using `It.IsAny`.

vpustovoitov: I use Bogus for generating random date, but is it okay that I created a few `TestFixture`s where they have different seeds for bogus randomizer ?

OutOfSchool/OutOfSchool.Common/Validators/CustomAgeAttribute.cs
```diff
@@ -0,0 +1,111 @@
...
+/// Validation attribute for dates. It allows only dates that are at least <see cref="MinAge"/> and at most <see cref="MaxAge"/>.
+/// This attribute can be used on <see cref="DateTime"/> or <see cref="DateOnly"/> types.
+/// </summary>
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Field | AttributeTargets.Parameter, AllowMultiple = false)]
+public class CustomAgeAttribute : DataTypeAttribute
+{
+    /// <summary>
+    /// The approximate number of days in a year including leap years.
+    /// </summary>
+    public const double ApproximateDaysInYear = 365.2425;
```
VadymLevkovskyi: (Redundant code) Is `ApproximateDaysInYear` still used?

vpustovoitov: > (Redundant code) Is `ApproximateDaysInYear` still used?

Forgot to remove, fixed

OutOfSchool/OutOfSchool.Common/Validators/CustomAgeAttribute.cs
```diff
@@ -0,0 +1,106 @@
...
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Field | AttributeTargets.Parameter, AllowMultiple = false)]
+public class CustomAgeAttribute : DataTypeAttribute
+{
+    public CustomAgeAttribute()
+        : base(DataType.Date)
+    {
+    }
+
+    /// <summary>
+    /// Gets the minimum age of the persom in years.
```
VadymLevkovskyi: (Mistype) `persom` is expected to have "n" at the end - perso**n**. 

Same in `MaxAge` property description.

OutOfSchool/OutOfSchool.WebApi.Tests/Validators/CustomAgeAttributeTests.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        // Act
+        var isValid = new CustomAgeAttribute() { MinAge = 3 }.IsValid(dateOfBirth);
+
+        // Assert
+        Assert.IsFalse(isValid);
+    }
+
+    [Test]
+    public void IsValid_WhenDateIsOneDayBeforeMaxAge_ShouldReturnFalse()
```
VadymLevkovskyi: (Test logic) I believe you are misinterpreting age calculation. Please explain why this test says that 3yrs and 1 day old person shouldn't fit `MaxAge=3` condition. 

---

Here is my line of thoughts:

The `CustomAgeAttribute.MaxAge` description says
```
    /// Gets the maximum age of the persom in years.
    /// </summary>
    public int MaxAge { get; init; } = int.MaxValue; 
```

So, `MaxAge` is integer and shows maximum allowed age in years. For me it means, that `MaxAge=3` condition disallows any 4 years old person and older. But every person below 4 years - i.e. 3 yrs and 1 day, as in this test - is allowed.

OutOfSchool/OutOfSchool.WebApi.Tests/Validators/CustomAgeAttributeTests.cs
VadymLevkovskyi: > ... is it okay ... have different seeds for bogus randomizer ?

The only profit - you test more variants in one test run. The downside - test suit run takes longer time (4 runs should be done for all tests, even for tests having no usage of Bogus); more test code has to be maintained (the seeding code); you would never get random data each time it will deterministic 4 specified, so the whole idea of randomness is lost.

If I were you I wouldn't use seeding at all. 
Bogus by default uses different seeds (and thus provides different random values) from one test run to another, so sooner or later problematic data would be found if it exists.

OutOfSchool/OutOfSchool.Common/Validators/CustomAgeAttribute.cs
VadymLevkovskyi: Closing this one, though it's not resolved. See other comment below

OutOfSchool/OutOfSchool.WebApi.Tests/Validators/CustomAgeAttributeTests.cs
vpustovoitov: > (Test logic) I believe you are misinterpreting age calculation. Please explain why this test says that 3yrs and 1 day old person shouldn't fit `MaxAge=3` condition.
> 
> Here is my line of thoughts:
> 
> The `CustomAgeAttribute.MaxAge` description says
> 
> ```
>     /// Gets the maximum age of the persom in years.
>     /// </summary>
>     public int MaxAge { get; init; } = int.MaxValue; 
> ```
> 
> So, `MaxAge` is integer and shows maximum allowed age in years. For me it means, that `MaxAge=3` condition disallows any 4 years old person and older. But every person below 4 years - i.e. 3 yrs and 1 day, as in this test - is allowed.

Now I get It, sorry for being stupid )

But for more clarity of how exactly I understood this:

dates

OutOfSchool/OutOfSchool.WebApi.Tests/Validators/CustomAgeAttributeTests.cs
vpustovoitov: > > ... is it okay ... have different seeds for bogus randomizer ?
> 
> The only profit - you test more variants in one test run. The downside - test suit run takes longer time (4 runs should be done for all tests, even for tests having no usage of Bogus); more test code has to be maintained (the seeding code); you would never get random data each time it will deterministic 4 specified, so the whole idea of randomness is lost.
> 
> If I were you I wouldn't use seeding at all. Bogus by default uses different seeds (and thus provides different random values) from one test run to another, so sooner or later problematic data would be found if it exists.

I remove seeding from tests, and run these tests until failure, and after finding some bugs and fixing them, everything is works fine (or maybe something would be founded later)

MrVenik:
@VadymLevkovskyi, I have one more question: In some models (e.g. `RegisterViewModel,` `TeacherDTO`, `ChildBaseDto`, `ShortUserDto`), we have inconsistencies in name validation (e.g. `FirstName`, `LastName`, `MiddleName`). During our daily meeting, we discussed that this should be fixed. Should these fixes be a part of this pull request, or should we open a new one with these changes?

VadymLevkovskyi:
> @VadymLevkovskyi, I have one more question: In some models ...  inconsistencies in name validation.... Should these fixes be a part of this pull request, or should we open a new one with these changes?

I'd prefer separate PR as name validation it is not the age validation or default field values. 

If it could be delivered separately (it compiles when separated, it could be auto-/manually tested in isolation, it could be deployed to environment in separate chunks) - it should be done in separate PRs/user stories. 

Usually it's better to have separate PRs for code changes that are not cohesive enough. Smaller parts of functionality could be better estimated, could fit into sprint backlog, could be reviewed in isolation and there will be no blockers from other code parts, in general more specific reviewers could be invited (sometimes, due to code ownership, different reviewers has to be present), could be deployed faster, could be tested faster. At same time there is freedom to join together two (or more) separate user stories/PRs to be processed as one piece during each mentioned process. The opposite (to split one big PR into smaller artifacts) is way harder.

Kind of "SRP principle" for requirements/PRs )

VadymLevkovskyi:
(Tests) Please improve code coverage. (The `CreateDraft()` implementation and call in `AuthController` should be tested)

MrVenik:
> > @VadymLevkovskyi, I have one more question: In some models ...  inconsistencies in name validation.... Should these fixes be a part of this pull request, or should we open a new one with these changes?
> 
> I'd prefer separate PR as name validation it is not the age validation or default field values.
> 
> If it could be delivered separately (it compiles when separated, it could be auto-/manually tested in isolation, it could be deployed to environment in separate chunks) - it should be done in separate PRs/user stories.
> 
> Usually it's better to have separate PRs for code changes that are not cohesive enough. Smaller parts of functionality could be better estimated, could fit into sprint backlog, could be reviewed in isolation and there will be no blockers from other code parts, in general more specific reviewers could be invited (sometimes, due to code ownership, different reviewers has to be present), could be deployed faster, could be tested faster. At same time there is freedom to join together two (or more) separate user stories/PRs to be processed as one piece during each mentioned process. The opposite (to split one big PR into smaller artifacts) is way harder.
> 
> Kind of "SRP principle" for requirements/PRs )

Got it! Thanks!

MrVenik:
> (Tests) Please improve code coverage. (The `CreateDraft()` implementation and call in `AuthController` should be tested)

I understand how to write tests for `CreateDraft` method, but how to test that static method called inside `AuthController` I don`t know (

OutOfSchool/OutOfSchool.WebApi.Tests/Models/ParentDraftCreationTests.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    [Test]
+    public void CreateDraft_ReturnsMaleValidAge()
+    {
+        // Arrange
+        var expectedDate = DateTime.Now.AddYears(-Constants.AdultAge).Date;
+
+        // Assert
+        Assert.AreEqual(expectedDate, Parent.CreateDraft("test").DateOfBirth.GetValueOrDefault().Date);
+    }
```
VadymLevkovskyi: (Flaky test) In this test you check the `Now` date. It will be broken from time to time.

There are 2 issues: 
- if test is running at 23:59:59.xxx the date of `Now` set into `expectedDate` and via `CreateDraft()` could be different day;
- test uses `Now` while implementation is using `UtcNow`, it would work correctly only for UTC timezone, for other timezones you could have problems during several hours of each day.

Solutions:
 - either use `TimeProvider` and `FakeTimeProvider` (added in .Net 8, see [example](https://grantwinney.com/how-to-use-timeprovider-and-faketimeprovider/)) - lot of code changes, good to have everywhere in code, not sure if this PR is a right place
 - or just pass the now datetime to the `CreateDraft()`.

vpustovoitov: > (Flaky test) In this test you check the `Now` date. It will be broken from time to time.
> 
> There are 2 issues:
> 
> * if test is running at 23:59:59.xxx the date of `Now` set into `expectedDate` and via `CreateDraft()` could be different day;
> * test uses `Now` while implementation is using `UtcNow`, it would work correctly only for UTC timezone, for other timezones you could have problems during several hours of each day.
> 
> Solutions:
> 
> * either use `TimeProvider` and `FakeTimeProvider` (added in .Net 8, see [example](https://grantwinney.com/how-to-use-timeprovider-and-faketimeprovider/)) - lot of code changes, good to have everywhere in code, not sure if this PR is a right place
> * or just pass the now datetime to the `CreateDraft()`.

Fixed, now `CreateDraft` accepts `currentTime` parameter as starting point for subtracting years.

VadymLevkovskyi: (Naming) Thanks for the fix, but please work on wording. 

Should the argument `currentTime` be UTC or local time? If you name it like `utcNow` and provide appropriate xml doc comment (right now it is `<param name="currentTime">Current time.</param>` that doesn't help at all) you would avoid this ambiguity.

VadymLevkovskyi:
> > ...
>
> I understand how to write tests for `CreateDraft` method, but how to test that static method called inside `AuthController` I don`t know (

Usage of static methods is untestable, for sure.

Thus you have 2 approaches 
- either check the behavior (that `parentRepo.Create()` in `Register` is called with `Parent` having expected field values - the quintessence of this whole PR) - but it will partly duplicate your test of `CreateDraft()`
- or redesign static `CreateDraft()` method into instance method of injected dependency (that could be mocked) - that is even weirder. (Remember, that "better" design was to use `DraftParent` class, and this way you would have it as parameter for `parentRepo.Create()` and all "testing" will be performed by compiler)

Your current coverage is 87%, so you could do nothing, but if you are curious here is implementation of test from 1st approach. 

```
    [Test]
    public async Task Register_UserInParentRole_CreatesParentWithDefaultValuesAndReturnsView()
    {
        // Arrange 
        var user = default(User);
        var viewModel = new Faker<RegisterViewModel>().Generate();

        fakeUserManager.Setup(manager => manager.CreateAsync(
                It.Is<User>(user => user.Email == viewModel.Email && user.UserName == viewModel.Email && !user.IsRegistered && !user.IsBlocked),
                viewModel.Password
            ))
            .Callback<User, string>((theUser, pwd) => user = theUser) // used to capture the generated user, especially .Id
            .Returns(Task.FromResult(IdentityResult.Success));
        fakeUserManager.Setup(manager => manager.AddToRoleAsync(It.Is<User>(theUser => user == theUser), nameof(Role.Parent).ToLower()))
            .Returns(Task.FromResult(IdentityResult.Success));
        fakeUserManager.Setup(manager => manager.GenerateEmailConfirmationTokenAsync(It.Is<User>(theUser => user == theUser)))
            .Returns(Task.FromResult<string>("some token"));
        fakeEmailSender.Setup(sender => sender.SendAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<(string, string)>(), null))
            .Returns(Task.CompletedTask);
        fakeSignInManager.Setup(manager => manager.SignInAsync(It.Is<User>(theUser => user == theUser), false, null))
            .Returns(Task.CompletedTask);
        fakeparentRepository.Setup(repo => repo.RunInTransaction(It.IsAny<Func<Task<Parent>>>()))
            .Callback<Func<Task<Parent>>>(func => func().GetAwaiter().GetResult()) // used to force repo.Create() (below) to be invoked
            .Returns(Task.FromResult(It.IsAny<Parent>())); 
        fakeparentRepository.Setup(repo => repo.Create(
                It.Is<Parent>(parent => parent.UserId == user.Id 
                    && parent.Gender == Gender.Male
                    && parent.DateOfBirth.HasValue
                    && (DateTime.UtcNow.AddYears(-Common.Constants.AdultAge) - parent.DateOfBirth.Value).TotalMinutes < 1 )
            ))
            .Returns(Task.FromResult(It.IsAny<Parent>()));

        var httpContext = new DefaultHttpContext();
        httpContext.Request.Scheme = "http";
        httpContext.Request.Host = new HostString("localhost");
        httpContext.Request.ContentType = "application/x-www-form-urlencoded";
        httpContext.Request.Form = new FormCollection(new Dictionary<string, StringValues> { { nameof(Role.Parent), "1" } });

        authController.ControllerContext = new ControllerContext
        {
            HttpContext = httpContext
        };
        authController.Url = new Mock<IUrlHelper>().Object;

        // Act
        var result = await authController.Register(viewModel);

        // Assert
        fakeUserManager.VerifyAll();
        fakeSignInManager.VerifyAll();
        fakeEmailSender.VerifyAll();
        fakeparentRepository.VerifyAll();

        Assert.IsEmpty(authController.ModelState.Values);
        Assert.IsInstanceOf<ViewResult>(result);
        Assert.IsInstanceOf<RegisterViewModel>((result as ViewResult).Model);
    }
```

MrVenik:
> > > ...
> > 
> > 
> > I understand how to write tests for `CreateDraft` method, but how to test that static method called inside `AuthController` I don`t know (
> 
> Usage of static methods is untestable, for sure.
> 
> Thus you have 2 approaches
> 
> * either check the behavior (that `parentRepo.Create()` in `Register` is called with `Parent` having expected field values - the quintessence of this whole PR) - but it will partly duplicate your test of `CreateDraft()`
> * or redesign static `CreateDraft()` method into instance method of injected dependency (that could be mocked) - that is even weirder. (Remember, that "better" design was to use `DraftParent` class, and this way you would have it as parameter for `parentRepo.Create()` and all "testing" will be performed by compiler)
> 
> Your current coverage is 87%, so you could do nothing, but if you are curious here is implementation of test from 1st approach.
> 
> ```
>     [Test]
>     public async Task Register_UserInParentRole_CreatesParentWithDefaultValuesAndReturnsView()
>     {
>         // Arrange 
>         var user = default(User);
>         var viewModel = new Faker<RegisterViewModel>().Generate();
> 
>         fakeUserManager.Setup(manager => manager.CreateAsync(
>                 It.Is<User>(user => user.Email == viewModel.Email && user.UserName == viewModel.Email && !user.IsRegistered && !user.IsBlocked),
>                 viewModel.Password
>             ))
>             .Callback<User, string>((theUser, pwd) => user = theUser) // used to capture the generated user, especially .Id
>             .Returns(Task.FromResult(IdentityResult.Success));
>         fakeUserManager.Setup(manager => manager.AddToRoleAsync(It.Is<User>(theUser => user == theUser), nameof(Role.Parent).ToLower()))
>             .Returns(Task.FromResult(IdentityResult.Success));
>         fakeUserManager.Setup(manager => manager.GenerateEmailConfirmationTokenAsync(It.Is<User>(theUser => user == theUser)))
>             .Returns(Task.FromResult<string>("some token"));
>         fakeEmailSender.Setup(sender => sender.SendAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<(string, string)>(), null))
>             .Returns(Task.CompletedTask);
>         fakeSignInManager.Setup(manager => manager.SignInAsync(It.Is<User>(theUser => user == theUser), false, null))
>             .Returns(Task.CompletedTask);
>         fakeparentRepository.Setup(repo => repo.RunInTransaction(It.IsAny<Func<Task<Parent>>>()))
>             .Callback<Func<Task<Parent>>>(func => func().GetAwaiter().GetResult()) // used to force repo.Create() (below) to be invoked
>             .Returns(Task.FromResult(It.IsAny<Parent>())); 
>         fakeparentRepository.Setup(repo => repo.Create(
>                 It.Is<Parent>(parent => parent.UserId == user.Id 
>                     && parent.Gender == Gender.Male
>                     && parent.DateOfBirth.HasValue
>                     && (DateTime.UtcNow.AddYears(-Common.Constants.AdultAge) - parent.DateOfBirth.Value).TotalMinutes < 1 )
>             ))
>             .Returns(Task.FromResult(It.IsAny<Parent>()));
> 
>         var httpContext = new DefaultHttpContext();
>         httpContext.Request.Scheme = "http";
>         httpContext.Request.Host = new HostString("localhost");
>         httpContext.Request.ContentType = "application/x-www-form-urlencoded";
>         httpContext.Request.Form = new FormCollection(new Dictionary<string, StringValues> { { nameof(Role.Parent), "1" } });
> 
>         authController.ControllerContext = new ControllerContext
>         {
>             HttpContext = httpContext
>         };
>         authController.Url = new Mock<IUrlHelper>().Object;
> 
>         // Act
>         var result = await authController.Register(viewModel);
> 
>         // Assert
>         fakeUserManager.VerifyAll();
>         fakeSignInManager.VerifyAll();
>         fakeEmailSender.VerifyAll();
>         fakeparentRepository.VerifyAll();
> 
>         Assert.IsEmpty(authController.ModelState.Values);
>         Assert.IsInstanceOf<ViewResult>(result);
>         Assert.IsInstanceOf<RegisterViewModel>((result as ViewResult).Model);
>     }
> ```

I used your code, and it’s quite insane that we need such extensive testing just to verify a simple static method.

VadymLevkovskyi:
> > ...
> 
> I used your code, and it’s quite insane that we need such extensive testing just to verify a simple static method.

---

Regarding you `Fix ?` commit.

The `.VerifyAll()` was important, please don't comment it out. Either drop the whole test or fix the real origin of failure.

The real problem is that test cases in this class are not isolated, because instead of doing it in `Setup()` dependencies are configured in constructor and then reused. Ofc, this way you'll get lot of method invocations from all tests, and therefore `.VerifyAll()` will go crazy.

The solution is either to perform all setup in `Setup()` (pretty unexpectedly, ha?) or to clear in `Arrange` part all invocations of mocks (like `fakeUserManager.Invocations.Clear()`) that are later verified in `Assert` part, but latter wouldn't work for parallel execution of tests (which we have).

---

Regarding "such extensive testing just to verify a simple static method"

No, it's not extensive. It's coincidence of 2 preexisting issues:
- lot of "mixed" logic in controller method (that is bad design, definitely)
- absence of any test verifying "happy-path" for parents registration.

So it's not the "static method" problem, but result of the whole situation with tests.

MrVenik:
> > > ...
> > 
> > 
> > I used your code, and it’s quite insane that we need such extensive testing just to verify a simple static method.
> 
> Regarding you `Fix ?` commit.
> 
> The `.VerifyAll()` was important, please don't comment it out. Either drop the whole test or fix the real origin of failure.
> 
> The real problem is that test cases in this class are not isolated, because instead of doing it in `Setup()` dependencies are configured in constructor and then reused. Ofc, this way you'll get lot of method invocations from all tests, and therefore `.VerifyAll()` will go crazy.
> 
> The solution is either to perform all setup in `Setup()` (pretty unexpectedly, ha?) or to clear in `Arrange` part all invocations of mocks (like `fakeUserManager.Invocations.Clear()`) that are later verified in `Assert` part, but latter wouldn't work for parallel execution of tests (which we have).
> 
> Regarding "such extensive testing just to verify a simple static method"
> 
> No, it's not extensive. It's coincidence of 2 preexisting issues:
> 
> * lot of "mixed" logic in controller method (that is bad design, definitely)
> * absence of any test verifying "happy-path" for parents registration.
> 
> So it's not the "static method" problem, but result of the whole situation with tests.

I understand that the underlying issue lies in the bad design of controllers and services. Unfortunately, I cannot easily solve it directly.

VadymLevkovskyi (approved):

PR #1464, opened 2024-04-22
Deleting of parent and provider was fixed

alrom:


OutOfSchool/OutOfSchool.WebApi.Tests/Services/ParentServiceTests.cs
```diff
@@ -159,6 +164,41 @@ public void BlockUnblockParent_WhenBlockUnblockParentDtoIsNull_ShouldThrowArgume
...
+        Assert.ThrowsAsync<ArgumentException>(async () => await parentService.Delete(parentId).ConfigureAwait(false));
+    }
+
+    [Test]
+    public async Task Delete_WhenIdIsValid_ShouldCallDeleteMethods()
+    {
+        // Arrange
+        Guid parentId = Guid.NewGuid();
+        Parent parent = ParentGenerator.Generate();
+        parent.Id = parentId;
```
VadymLevkovskyi: (Excessive code) As `ParentGenerator` already creates `Id` as `Guid.NewGuid()`, you don't need to do it additionally.
So it could be as simple as 
```
        Parent parent = ParentGenerator.Generate();
        var parentId = parent.Id;
```

OutOfSchool/OutOfSchool.WebApi/Services/ParentService.cs
```diff
@@ -52,19 +55,25 @@ public async Task Delete(Guid id)
...
-        catch (DbUpdateConcurrencyException)
-        {
-            logger.LogError("Deleting failed. Parent with Id = {Id} doesn't exist in the system", id);
-            throw;
-        }
+
+            await userService.Delete(entity.UserId);
+
+            logger.LogInformation("User with Id = {UserId} successfully deleted", entity.UserId);
+        });
```
VadymLevkovskyi: (Logic, logs) There is potential of having inconsistent logs. Transaction could be rollbacked after execution of this lambda. As a result you could have logs stating that "X successfully deleted", but nothing changed in db. 
Please consider logging after the `await ...RunInTransaction(...)`. 

(Concurrency) `.ConfigureAwait()` is inconsistently used, either use it for both awaitables or remove it for both.

OutOfSchool/OutOfSchool.WebApi/Services/ProviderServices/ProviderService.cs
```diff
@@ -753,13 +753,16 @@ await UpdateWorkshopsProviderStatus(providerUpdateDto.Id, providerUpdateDto.Stat
...
 
-        logger.LogInformation("Provider with Id = {Id} successfully deleted", id);
+            logger.LogInformation("Provider with Id = {Id} successfully deleted", id);
 
-        await userService.Delete(entity.UserId);
+            await userService.Delete(entity.UserId);
 
-        logger.LogInformation("User with Id = {entity.UserId} successfully deleted", entity.UserId);
+            logger.LogInformation("User with Id = {entity.UserId} successfully deleted", entity.UserId);
+        });
```
VadymLevkovskyi: (Logic, logs) There is potential of having inconsistent logs. Transaction could be rollbacked after execution of this lambda. As a result you could have logs stating that "X successfully deleted", but nothing changed in db. 
Please consider logging after the `await ...RunInTransaction(...)`. 

(Concurrency) `.ConfigureAwait()` is inconsistently used, either use it for both awaitables or remove it for both.

VadymLevkovskyi (changes requested):

OutOfSchool/OutOfSchool.WebApi/Services/ParentService.cs
vpustovoitov: > (Logic, logs) There is potential of having inconsistent logs. Transaction could be rollbacked after execution of this lambda. As a result you could have logs stating that "X successfully deleted", but nothing changed in db. Please consider logging after the `await ...RunInTransaction(...)`.
> 
> (Concurrency) `.ConfigureAwait()` is inconsistently used, either use it for both awaitables or remove it for both.

Sorry for interrupting, but one more point regarding logs: While I believe that `userService` may already include some logging logic, even if it doesn’t, in my opinion, `ParentService` should not be responsible for logging information related to user removal status.

OutOfSchool/OutOfSchool.WebApi/Services/ProviderServices/ProviderService.cs
alrom: Done

OutOfSchool/OutOfSchool.WebApi/Services/ParentService.cs
alrom: Done

VadymLevkovskyi (approved):

PR #1473, opened 2024-05-03
Move extensions to extensions folder

MrVenik:
I think that extensions should live in extensions folder


OutOfSchool/OutOfSchool.BusinessLogic/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        using var streamReader = new StreamReader(stream);
+
+        return JsonConvert.DeserializeObject<T>(streamReader.ReadToEnd());
+    }
+
+    public static void SerializeToJsonAndWrite<T>(this Stream stream, T objectToWrite)
+    {
+        ArgumentNullException.ThrowIfNull(stream);
+
+        if (!stream.CanRead)
```
VadymLevkovskyi: (Logic) I believe here should be `CanWrite`

VadymLevkovskyi (changes requested):
Taking into account wrong check for `.CanWrite` I'm not sure that existing test suit is correct and complete. Could you check it as well?

VadymLevkovskyi (approved):

Issue #1458, opened 2024-04-18
[Admins cabinets] Any Admin cann't unblock a blocked Institution 


OlenaYakubova:
**Environment:** Windows 11, 64 bit
Google Chrome 124.0.6367.61
**Reproducible:** always.

**Preconditions**
1. Go to https://oos.dmytrominochkin.cloud/
3. Log in as Admin (e. g. Region Admin)
4. Admin from Preconditions has at least one blocked Institution

**Steps to reproduce**
1. Click on Name or arrow at top right corner of page
2. Select "Адміністрування" line from drop-down list
3. Click on 'Заклади' tab
4. Click on the three vertical dots opposite the blocked Institution
5. Select "Розблокувати заклад" line

**Actual result**
Admin cann't unblock blocked Institution. 
![image](https://github.com/ita-social-projects/OoS-Backend/assets/73597578/9aa8d076-17f1-4936-9c80-d129cfdd8dde)

**Expected result**
Admin can block and unblock Institution

**User story and test case links**
[Test case OS-739](https://jira.softserve.academy/browse/OS-739)
[User story #478](https://github.com/ita-social-projects/OoS-Backend/issues/478)
[Roles and Permissions matrix](https://docs.google.com/spreadsheets/d/1yD1PH5b-4RjtNjNkTZYhOERCRvD3aFK71KVL2noImAw/edit?pli=1#gid=404315089)

**Labels to be added**
"Bug", Priority ("High"), Severity ("Medium"), Type ("Functional").

PR #1476, opened 2024-05-09
Kymachynskyi/ Fix the unblocking of the institution by the admin #1458

DKymachynskyi:
The `ProviderBlockDto` is responsible for transferring data used to block or unblock an institution. The `BlockPhoneNumber` property is required to block an institution, but it is not used to unblock one.

During the unblocking process, the `BlockPhoneNumber` property, which is initialized with the value `string.Empty`, fails validation as the `CustomPhoneNumber` validation attribute does not check if `BlockPhoneNumber` equals `string.Empty`. This failure in validation renders it impossible to unblock the institution.


VadymLevkovskyi (approved):
Approved, 
if possible I'd like to have some description in PR why change of phone number validation affects the unblocking functionality

DKymachynskyi:
@VadymLevkovskyi I've added the description. Thank you!

PR #1472, opened 2024-05-03
Move background jobs from bl

MrVenik:
# Summary
- Move quartz constants into `OutOfSchool.Common` project
- Move quartz configs into `OutOfSchool.BackgroundJobs` project
- Move quartz jobs into `OutOfSchool.BackgroundJobs` project
- Move quartz DI registration extension methods into `OutOfSchool.BackgroundJobs` project
- Add tests for moved quartz jobs
- Add tests for moved quartz DI registration extension methods

PR #1477, opened 2024-05-12
Email sender fix

olsavchenk:
This is a fix for the email sender. The problem was that emails were not sent because the created task was not executed.

Now, EmailSenderService will create a task and an associated trigger instance and add it to the database, the task will be run depending on SendGridAccessibility. After the task is executed, the created task and trigger will be automatically deleted from the database. I also deleted the TriggerListener because it does not work with the current implementation (it does not run for some reason), so I moved its logic to the service and Listener. Now everything works fine: the email is sent, and if an error occurs, the task is rescheduled.

NOTE: Also I encountered an issue with email format. It brokes for some reason(or probably it is local issue). I will discuss this on daily meet-up tomorrow(13.05) and create ticket with fix for it. I can try to fix it in this PR but as I believe this PR is blocker for QA's


OutOfSchool/OutOfSchool.AuthorizationServer/Extensions/EmailSenderExtension.cs
```diff
@@ -0,0 +1,29 @@
...
+using Quartz;
+using Quartz.Impl.Matchers;
+
+namespace OutOfSchool.AuthorizationServer.Extensions;
+
+public static class EmailSenderExtension
+{
+    public static void AddEmailSender(
+        this IServiceCollectionQuartzConfigurator quartz,
+        IServiceCollection services,
```
VadymLevkovskyi: (Excessive parameter) `services` seems to be never used. Please consider removing it

OutOfSchool/OutOfSchool.AuthServer.Tests/EmailSender/EmailSenderServiceTests.cs
```diff
@@ -16,18 +16,22 @@ public class EmailSenderServiceTests
...
+    private Mock<ISendGridAccessibilityService> mockEndGridAccessibilityService;
     private EmailSenderService emailSenderService;
 
     [SetUp]
     public void Setup()
     {
         mockSchedulerFactory = new Mock<ISchedulerFactory>();
         mockScheduler = new Mock<IScheduler>();
-        emailSenderService = new EmailSenderService(mockSchedulerFactory.Object);
+        mockEndGridAccessibilityService = new Mock<ISendGridAccessibilityService>();
```
VadymLevkovskyi: (Variable scope) The `mockEndGridAccessibilityService` seems to be never used outside the `Setup()`. Please consider moving it from `class field` to `method local variable`.

VadymLevkovskyi (commented):
There are minor code style issues. 
And test coverage. 
Please at least cover new method in `SendGridAccessibilityService` and `EmailSenderExtension` (for latter - you could check [this](https://github.com/ita-social-projects/OoS-Backend/pull/1472#issuecomment-2107537701) comment from other PR)

OutOfSchool/OutOfSchool.AuthServer.Tests/EmailSender/SendGridAccessibilityServiceTests.cs
```diff
@@ -83,4 +83,32 @@ public void IsSendGridAccessible_WithOptionsTimeoutTimeNull_ReturnsTrueAfterTime
...
+    }
+
+    [Test]
+    public void GetAccessibilityTime_WithInaccessibleSendGrid_ReturnsAccessibilityTime()
+    {
+        // Arrange
+        var now = DateTimeOffset.Now;
+
+        // Act
+        _emailService.SetSendGridInaccessible(now);
```
VadymLevkovskyi: (Small naming nuance) Technically saying `.SetSendGridInaccessible()` relates to `Arrange` part, not `Act`.

OutOfSchool/OutOfSchool.AuthServer.Tests/EmailSender/SendGridAccessibilityServiceTests.cs
```diff
@@ -83,4 +83,32 @@ public void IsSendGridAccessible_WithOptionsTimeoutTimeNull_ReturnsTrueAfterTime
...
+        var accessTime = _emailService.GetAccessibilityTime(now);
+
+        // Assert
+        Assert.AreNotEqual(now, accessTime);
+        Assert.AreEqual(now.AddMinutes(300), accessTime);
```
VadymLevkovskyi: (Fragile test) This test is checking that next access time should be higher than now. We don't know in test the precise difference (currently used 5hrs could be changed in implementation), and shouldn't know/rely on it. 

That's why please assert only one condition `now < accessTime`.

OutOfSchool/OutOfSchool.AuthServer.Tests/Extensions/EmailSenderExtensionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+using OutOfSchool.EmailSender.Services;
+using Quartz;
+
+namespace OutOfSchool.AuthServer.Tests.Extensions;
+
+[TestFixture]
+public class EmailSenderExtensionsTests
+{
+    [Test]
+    public async Task AddEmailSender_WithValidConfig_AddsJobAndTrigger()
```
VadymLevkovskyi: (Test naming) Name part `_AddsJobAndTrigger` promise to check adding job and trigger, but in the code only job existence is asserted. Please either change the name or add trigger existence check.

VadymLevkovskyi (commented):
Logic ok, few minor test code issues

VadymLevkovskyi (approved):

Issue #1478, opened 2024-05-13
[Notification] The notification is not sent to the parent after the service provider change the application status to “Прийнято до конкурсного відбору”


OMGaranina:
**Environment:** OS Windows 11Pro Version 23H2, Browser Chrome Version 124.0.6367.119.
**Reproducible:** always
**Build found:** commit [58fe071](https://github.com/ita-social-projects/OoS-Backend/commit/58fe071ee4fbbbd6e6722090c9616c1788f17d3a)

**Preconditions**
1. The "Позашкілля" site https://oos.dmytrominochkin.cloud/ is opened
2. The user is logged in as a service provider
3. The workshop with competitive selection is created
4. The application from the Parent_1 to the workshop is sent

**Steps to reproduce**
1. Click on the dropdown button near the username in the header
2. Click on the 'Заявки' link
3. Click on the 'Прийняти до відбору' button
4. Log in as a Parent_1
5. Observe the notification about application status changing 

**Actual result**
The notification about application status changing to “Прийнято до конкурсного відбору” is not sent to Parent_1
![image](https://github.com/ita-social-projects/OoS-Backend/assets/95030828/da8fa3fe-6cb7-4c20-b62a-cb4ee1f5cc93)
![image](https://github.com/ita-social-projects/OoS-Backend/assets/95030828/f39a5c58-c91d-4d7f-82ce-63a0f0d713f3)

**Expected result**
Parent_1 gets the notification about the application status changing to “Прийнято до конкурсного відбору”.

**User story and test case links**
User story #182

PR #1480, opened 2024-05-14
Add notification for updating status of application to AcceptedForSelection

alrom:
#1478

### DIFF
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Services/ApplicationService.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Services/ApplicationService.cs
@@ -979,7 +979,8 @@ public class ApplicationService : IApplicationService, ISensitiveApplicationServ
                 && Enum.TryParse(additionalData[StatusTitle], out ApplicationStatus applicationStatus))
             {
                 if (applicationStatus == ApplicationStatus.Approved
-                    || applicationStatus == ApplicationStatus.Rejected)
+                    || applicationStatus == ApplicationStatus.Rejected
+                    || applicationStatus == ApplicationStatus.AcceptedForSelection)
                 {
                     recipientIds.Add(application.Parent.UserId);
                 }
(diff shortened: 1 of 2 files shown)


OutOfSchool/OutOfSchool.BusinessLogic/Services/ApplicationService.cs
```diff
@@ -979,7 +979,8 @@ await AllowedNewApplicationByChildStatus(applicationDto.WorkshopId, applicationD
                 && Enum.TryParse(additionalData[StatusTitle], out ApplicationStatus applicationStatus))
             {
                 if (applicationStatus == ApplicationStatus.Approved
-                    || applicationStatus == ApplicationStatus.Rejected)
+                    || applicationStatus == ApplicationStatus.Rejected
+                    || applicationStatus == ApplicationStatus.AcceptedForSelection)
```
VadymLevkovskyi: (Design) Please consider if this check could be extracted and reused as a helper method. Currently there are 2 other places where kind of same set of condition is checked - in `ExecuteUpdateAsync()` the `if` wrapping call to `SendApplicationUpdateStatusEmail()`, and in mentioned `SendApplicationUpdateStatusEmail()` the `switch` throwing "unsupported" `ArgumentException` when condition is not met. Here you've introduced the 3rd place. 
If it's not the coincidence and it's expected that logic for all mentioned places should be the same - please make it technically the same and implemented in 1 place.

VadymLevkovskyi (changes requested):
There is small improvement in design (see other comment)
And there are **no tests** for changed logic!

OutOfSchool/OutOfSchool.BusinessLogic/Services/ApplicationService.cs
alrom: Done

VadymLevkovskyi (approved):

PR #1481, opened 2024-05-16
Rework of registration logic

MrVenik:
This PR is based on #1468


OutOfSchool/OutOfSchool.Common/Validators/CustomAgeAttribute.cs
```diff
@@ -0,0 +1,106 @@
...
+/// Validation attribute for dates. It allows only dates that are at least <see cref="MinAge"/> and at most <see cref="MaxAge"/>.
+/// This attribute can be used on <see cref="DateTime"/> or <see cref="DateOnly"/> types.
+/// </summary>
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Field | AttributeTargets.Parameter, AllowMultiple = false)]
+public class CustomAgeAttribute : DataTypeAttribute
+{
+    public CustomAgeAttribute()
+        : base(DataType.Date)
+    {
+    }
```
VadymLevkovskyi: (Code style) Please consider using of the primary constructor "feature" here. I know, that it's not a big improvement ;) , but it removes extra code, so it is worthy

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ParentServiceTests.cs
```diff
@@ -48,9 +53,170 @@ public void SetUp()
...
+    {
+        // Arrange
+        currentUserServiceMock.SetupGet(s => s.UserId).Returns(userId);
+
+        // Act & Assert
+        Assert.ThrowsAsync<InvalidOperationException>(async () => await parentService.Create(new ParentCreateDto()).ConfigureAwait(false));
+
+        loggerMock.Verify(
+            logger => logger.Log(
+                It.Is<LogLevel>(logLevel => logLevel == LogLevel.Error),
```
VadymLevkovskyi: (Code style) It is possible to write less code here. Instead of
`It.Is<LogLevel>(logLevel => logLevel == LogLevel.Error)`
just simple
`LogLevel.Error`

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ParentServiceTests.cs
```diff
@@ -48,9 +53,170 @@ public void SetUp()
...
+        userRepositoryMock
+            .Setup(r => r.GetById("userId"))
+            .ReturnsAsync(null as User);
+
+        // Act & Assert
+        Assert.ThrowsAsync<InvalidOperationException>(async () => await parentService.Create(new ParentCreateDto()).ConfigureAwait(false));
+
+        loggerMock.Verify(
+            logger => logger.Log(
+                It.Is<LogLevel>(logLevel => logLevel == LogLevel.Error),
```
VadymLevkovskyi: (Code style, duplicate) It is possible to write less code here. Instead of
`It.Is<LogLevel>(logLevel => logLevel == LogLevel.Error)`
just simple
`LogLevel.Error`

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ParentServiceTests.cs
```diff
@@ -48,9 +53,170 @@ public void SetUp()
...
+        parentRepositoryMock
+            .Setup(r => r.Any(It.IsAny<Expression<Func<Parent, bool>>>()))
+            .ReturnsAsync(true);
+
+        // Act & Assert
+        Assert.ThrowsAsync<InvalidOperationException>(async () => await parentService.Create(new ParentCreateDto()).ConfigureAwait(false));
+
+        loggerMock.Verify(
+            logger => logger.Log(
+                It.Is<LogLevel>(logLevel => logLevel == LogLevel.Error),
```
VadymLevkovskyi: (Code style, duplicate # 3) It is possible to write less code here. Instead of
`It.Is<LogLevel>(logLevel => logLevel == LogLevel.Error)`
just simple
`LogLevel.Error`

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ParentServiceTests.cs
```diff
@@ -48,9 +53,170 @@ public void SetUp()
...
+        // Assert
+        Assert.True(user.IsRegistered);
+    }
+
+    [Test]
+    public async Task Create_WhenUserExists_ShouldSetUserPhoneNumber()
+    {
+        // Arrange
+        var user = UserGenerator.Generate();
+        var expectedPhoneNumber = "+380999999999";
```
VadymLevkovskyi: (Fake test data generation, not insisting) Please consider usage of Bogus for fake data generation.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ParentServiceTests.cs
```diff
@@ -48,9 +53,170 @@ public void SetUp()
...
+            .Setup(r => r.GetById(user.Id))
+            .ReturnsAsync(user);
+
+        parentRepositoryMock
+            .Setup(r => r.Create(It.IsAny<Parent>()))
+            .ReturnsAsync(new Parent());
+
+        parentRepositoryMock
+            .SetupGet(r => r.UnitOfWork)
+            .Returns(new Mock<IUnitOfWork>().Object);
```
VadymLevkovskyi: (Test logic) There are few issues with selected code, originated from weird Sonar coverage calculation algorithm, violation of Demeter's law done in design of our repositories, and lack of testing. Please consider (in this and other tests) applying changes sugested below:

1. replace in `Arrange` section
```
        parentRepositoryMock
            .SetupGet(r => r.UnitOfWork)
            .Returns(new Mock<IUnitOfWork>().Object);
``` 
by
``` 
        parentRepositoryMock.Setup(r => r.UnitOfWork.CompleteAsync()) .Returns(Task.FromResult(1));`
``` 

2. and add in `Assert` section
``` 
        parentRepositoryMock.Verify(r => r.UnitOfWork.CompleteAsync(), Times.Once);`
``` 

---

More explanations

As I understand the selected code is used to support next line of code in `ParentService.Create()`
`        await repositoryParent.UnitOfWork.CompleteAsync();`

Imagine that `UnitOfWork` also implements `RollbackAsync()` method (now it doesn't and it's really strange, but that's other point). If code of `ParentService.Create()` by mistake calls `RollbackAsync()` instead of currently used `CompleteAsync()` your test will not fail. But it should as data are not created! Thus testing logic has gap.

Unfortunately Sonar thinks that that specific line is covered, thus you would never be forced to close this gap.

And yes, as without selected mocking code test fails with NRE, it's enough to provide mocks for your mocks, but this is usually a sign of a smell in design - viloation of Demeter's law.

Demeter's law states (among other statements) - "Only talk to your immediate friends". And since the time when we'd introduced public `UnitOfWork` to our repository API, and started to use it via call chain like `repo.UnitOfWork.Complete...()` (in total there are 11 places in our code) we are violating this statement. As a result tests are also "broken" - extra dependency is introduced that increases coupling and thus makes tests more fragile.

So, origin of the problem is design of repository API. But, as redesign is out of this PR scope, please at least make test more stable and "covering" as suggested (btw, this is already used in other tests).

VadymLevkovskyi (commented):
Logic seems ok. Test design has few minor issues, see comments below

OutOfSchool/OutOfSchool.WebApi.Tests/Services/ParentServiceTests.cs
vpustovoitov: > Demeter's law states (among other statements) - "Only talk to your immediate friends". And since the time when we'd introduced public `UnitOfWork` to our repository API, and started to use it via call chain like `repo.UnitOfWork.Complete...()` (in total there are 11 places in our code) we are violating this statement. As a result tests are also "broken" - extra dependency is introduced that increases coupling and thus makes tests more fragile.
> 
> So, origin of the problem is design of repository API. But, as redesign is out of this PR scope, please at least make test more stable and "covering" as suggested (btw, this is already used in other tests).

Thank you for the explanation! This information is really helpful. Although I heard about Demeter’s law, I never encountered a violation of it in practice.

VadymLevkovskyi (approved):

PR #1482, opened 2024-05-21
Kymachynskyi/ Fix workshop status and seats availability #2450

DKymachynskyi:
- Add check if a new value of workshop `AvailableSeats` is equal to or greater than the workshop's `TakenSeats` value.
- Set the workshop's status to open when the workshop is updated to unlimited seats by the workshops admin.
- Set the workshop's status to closed when the workshop is updated to limited seats and `AvailableSeats` equal to `TakenSeats`.
- A workshop's admin may suspend recruitment of children to the workshop with unlimited seats.

### DIFF
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
@@ -2,15 +2,15 @@
 using AutoMapper;
 using H3Lib;
 using H3Lib.Extensions;
-using OutOfSchool.Common.Enums;
-using OutOfSchool.Services.Enums;
-using OutOfSchool.Services.Models.Images;
 using OutOfSchool.BusinessLogic.Common;
 using OutOfSchool.BusinessLogic.Enums;
 using OutOfSchool.BusinessLogic.Models;
 using OutOfSchool.BusinessLogic.Models.Images;
 using OutOfSchool.BusinessLogic.Models.Workshops;
 using OutOfSchool.BusinessLogic.Services.AverageRatings;
+using OutOfSchool.Common.Enums;
+using OutOfSchool.Services.Enums;
+using OutOfSchool.Services.Models.Images;
 
 namespace OutOfSchool.BusinessLogic.Services;
 
@@ -339,6 +339,11 @@ public class WorkshopService : IWorkshopService
         {
             await UpdateDateTimeRanges(dto.DateTimeRanges, dto.Id).ConfigureAwait(false);
             var currentWorkshop = await workshopRepository.GetWithNavigations(dto!.Id).ConfigureAwait(false);
+            var currentWorkshopTakenSeats = currentWorkshop.Applications.TakenSeats();
+            if (dto.AvailableSeats < currentWorkshopTakenSeats)
+            {
+                throw new ArgumentException("The number of available seats must be equal or exceed the number of taken seats");
+            }
 
             // In case if AddressId was changed. AddresId is one and unique for workshop.
             dto.AddressId = currentWorkshop.AddressId;
@@ -349,6 +354,26 @@ public class WorkshopService : IWorkshopService
             if (dto.AvailableSeats is 0 or null)
             {
                 dto.AvailableSeats = uint.MaxValue;
+            }
+
+            if (dto.AvailableSeats == uint.MaxValue
+                && currentWorkshop.AvailableSeats == currentWorkshopTakenSeats)
+            {
+                await UpdateStatus(new()
+                {
+                    WorkshopId = currentWorkshop.Id,
+                    Status = WorkshopStatus.Open,
+                }).ConfigureAwait(false);
+            }
+
+            if (dto.AvailableSeats < uint.MaxValue
+                && dto.AvailableSeats == currentWorkshopTakenSeats)
+            {
+                await UpdateStatus(new()
+                {
+                    WorkshopId = currentWorkshop.Id,
+                    Status = WorkshopStatus.Closed,
+                }).ConfigureAwait(false);
             }
 
             mapper.Map(dto, currentWorkshop);
@@ -380,18 +405,7 @@ public class WorkshopService : IWorkshopService
 
         if (currentWorkshop.Status != dto.Status)
         {
-            if (currentWorkshop.AvailableSeats != uint.MaxValue)
-            {
-                currentWorkshop.Status = dto.Status;
-            }
-            else
-            {
-                logger.LogInformation(
-                    $"Unable to update status for workshop(id) {dto.WorkshopId}. Number of seats has not restriction.");
-                throw new ArgumentException(
-                    "Unable to update status for workshop because of number of seats has not restriction.");
-            }
-
+            currentWorkshop.Status = dto.Status;
             try
             {
                 await workshopRepository.Update(currentWorkshop).ConfigureAwait(false);
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
@@ -901,21 +901,27 @@ public class WorkshopService : IWorkshopService
         if (newAvailableSeats == uint.MaxValue
             && currentWorkshop.AvailableSeats == currentWorkshopTakenSeats)
         {
-            await UpdateStatus(new()
+            if (currentWorkshop.Status == WorkshopStatus.Closed)
             {
-                WorkshopId = currentWorkshop.Id,
-                Status = WorkshopStatus.Open,
-            }).ConfigureAwait(false);
+                await UpdateStatus(new()
+                {
+                    WorkshopId = currentWorkshop.Id,
+                    Status = WorkshopStatus.Open,
+                }).ConfigureAwait(false);
+            }
         }
 
         if (newAvailableSeats < uint.MaxValue
             && newAvailableSeats == currentWorkshopTakenSeats)
         {
-            await UpdateStatus(new()
+            if (currentWorkshop.Status == WorkshopStatus.Open)
             {
-                WorkshopId = currentWorkshop.Id,
-                Status = WorkshopStatus.Closed,
-            }).ConfigureAwait(false);
+                await UpdateStatus(new()
+                {
+                    WorkshopId = currentWorkshop.Id,
+                    Status = WorkshopStatus.Closed,
+                }).ConfigureAwait(false);
+            }
         }
     }
 }
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
@@ -899,29 +899,25 @@ public class WorkshopService : IWorkshopService
         var currentWorkshopTakenSeats = currentWorkshop.Applications.TakenSeats();
 
         if (newAvailableSeats == uint.MaxValue
-            && currentWorkshop.AvailableSeats == currentWorkshopTakenSeats)
+            && currentWorkshop.AvailableSeats == currentWorkshopTakenSeats
+            && currentWorkshop.Status == WorkshopStatus.Closed)
         {
-            if (currentWorkshop.Status == WorkshopStatus.Closed)
+            await UpdateStatus(new()
             {
-                await UpdateStatus(new()
-                {
-                    WorkshopId = currentWorkshop.Id,
-                    Status = WorkshopStatus.Open,
-                }).ConfigureAwait(false);
-            }
+                WorkshopId = currentWorkshop.Id,
+                Status = WorkshopStatus.Open,
+            }).ConfigureAwait(false);
         }
 
         if (newAvailableSeats < uint.MaxValue
-            && newAvailableSeats == currentWorkshopTakenSeats)
+            && newAvailableSeats == currentWorkshopTakenSeats
+            && currentWorkshop.Status == WorkshopStatus.Open)
         {
-            if (currentWorkshop.Status == WorkshopStatus.Open)
+            await UpdateStatus(new()
             {
-                await UpdateStatus(new()
-                {
-                    WorkshopId = currentWorkshop.Id,
-                    Status = WorkshopStatus.Closed,
-                }).ConfigureAwait(false);
-            }
+                WorkshopId = currentWorkshop.Id,
+                Status = WorkshopStatus.Closed,
+            }).ConfigureAwait(false);
         }
     }
 }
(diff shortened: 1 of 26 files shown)


OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -899,4 +893,29 @@ private async Task UpdateWorkshop()
...
+
+        if (newAvailableSeats < uint.MaxValue
+            && newAvailableSeats == currentWorkshopTakenSeats)
+        {
+            await UpdateStatus(new()
+            {
+                WorkshopId = currentWorkshop.Id,
+                Status = WorkshopStatus.Closed,
+            }).ConfigureAwait(false);
+        }
```
VadymLevkovskyi: (Logic) What if `newAvailableSeats > currentWorkshopTakenSeats`? Workshop will not be closed, though there is "overlimit" of participants. Please consider changing condition in this `if`.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -899,4 +893,29 @@ private async Task UpdateWorkshop()
     }
 
     private void ValidateOffsetFilter(OffsetFilter offsetFilter) => ModelValidationHelper.ValidateOffsetFilter(offsetFilter);
+
+    private async Task UpdateWorkshopStatusBySeatsLimitAndAvailability(uint newAvailableSeats, Workshop currentWorkshop)
+    {
+        var currentWorkshopTakenSeats = currentWorkshop.Applications.TakenSeats();
```
VadymLevkovskyi, line 898: (Logic) Related to the whole method. Here you change workshop status when some seats-related conditions are met. But you don't check the old workshop status. Thus you could set the same value for it again and again. Isn't it better to do changes only if they are really necessary?

VadymLevkovskyi (commented):
There are 3 comments about logic.

And I see no test changed to cover that changed logic (workshop status update) - could you point me any (or write a new one)?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
VadymLevkovskyi: Thanks for the change! 
One minor code style question - could you merge nested `if`s into one `if`?

DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
DKymachynskyi: If the admin switched the workshop to a limited number of seats `newAvailableSeats < uint.MaxValue` and at the same time set a new available number of seats greater than the current number of taken seats `newAvailableSeats > currentWorkshopTakenSeats`, we leave the current workshop status.

VadymLevkovskyi: I'm sorry, ofc I was asking about opppsite - `newAvailableSeats < currentWorkshopTakenSeats` case. 
If new available seats number are less than already taken we should at least block new registrations (so change status to Closed). 
Current code doesn't check this condition, instead only precise equality is checked `newAvailableSeats == currentWorkshopTakenSeats`, thus the status won't be set to `Closed` for the case when 5 seats are taken, but new available is 3.

DKymachynskyi: I did not consider this condition because I added the corresponding check in the controller. My mistake. Fixed. Thank you!

VadymLevkovskyi (approved):

VadymLevkovskyi (approved):

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopController.cs
```diff
@@ -361,7 +361,14 @@ public async Task<IActionResult> Update([FromBody] WorkshopBaseDto dto)
...
 
-        if (!await combinedWorkshopService.IsAvailableSeatsValid(dto.AvailableSeats, dto.Id))
+        var workshop = await combinedWorkshopService.GetById(dto.Id, true).ConfigureAwait(false);
+
+        if (workshop is null)
+        {
+            return BadRequest("Workshop does not exist.");
+        }
+
+        if (!combinedWorkshopService.IsAvailableSeatsValid(dto.AvailableSeats, workshop))
```
VadymLevkovskyi: (Design smell) Please consider moving such calls closer to the implementation, because now we have some sort of "jealous method" smell: controller action, that doesn't trust `combinedWorkshopService` and instead is calling each method (`.GetById()`, `.IsAvailableSeatsValid()`, `.Update()`) by itself.
Ideally, all logic related to data validation that is dependent on db data should be done in service.

OutOfSchool/OutOfSchool.WebApi/Controllers/V2/WorkshopController.cs
```diff
@@ -200,13 +200,25 @@ public async Task<IActionResult> Create([FromForm] WorkshopV2Dto dto)
...
+        }
+
         var userHasRights = await IsUserProvidersOwner(dto.ProviderId).ConfigureAwait(false);
         if (!userHasRights)
         {
             return StatusCode(StatusCodes.Status403Forbidden, "Forbidden to update workshops for another providers.");
         }
 
-        if (!await combinedWorkshopService.IsAvailableSeatsValid(dto.AvailableSeats, dto.Id))
+        var workshop = await combinedWorkshopService.GetById(dto.Id, true).ConfigureAwait(false);
```
VadymLevkovskyi: Similar "design issue" as already mentioned. Btw, code duplication, you have here and in V1 controller, should trigger your attention that something is going wrong/misplaced/etc

VadymLevkovskyi (changes requested):
Design and test coverage

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopControllerTests.cs
```diff
@@ -533,14 +533,35 @@ public async Task UpdateWorkshop_WhenDtoAvailableSeatsIsInvalid_ShouldReturnBadR
...
+
+    [Test]
+    public async Task UpdateWorkshop_WhenWorkshopEntityNotExist_ShouldReturnBadRequestObjectResult()
+    {
+        // Arrange
+        workshopUpdateDto.ProviderId = provider.Id;
+        var workshopDto = null as WorkshopDto;
+        providerServiceMoq.Setup(x => x.IsBlocked(It.IsAny<Guid>())).ReturnsAsync(false);
+        providerServiceMoq.Setup(x => x.GetByUserId(It.IsAny<string>(), It.IsAny<bool>())).ReturnsAsync(provider);
+        workshopServiceMoq.Setup(x => x.GetById(It.IsAny<Guid>(), It.IsAny<bool>())).ReturnsAsync(workshopDto);
```
VadymLevkovskyi: (Test design) Please avoid using `It.IsAny` when you know what data are expected. This way your test code is more clear (as instead of abstract `It.IsAny<Guid>()` you'll have specific `provider.Id`) and more precise in testing.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/WorkshopServiceTests.cs
```diff
@@ -864,6 +838,27 @@ private void SetupGetByFilter(IEnumerable<Workshop> workshops, IEnumerable<Avera
...
+        for (int i = 0; i < allApplications; i++)
+        {
+            if (i < approvedApplications)
+            {
+                applications[i].Status = ApplicationStatus.Approved;
+            }
+            else
+            {
+                applications[i].Status = ApplicationStatus.Rejected;
+            }
```
VadymLevkovskyi: (Code style, suggestion) Not insisting, but consider avoiding heavy `if / else` pattern, when you are doing same things in both `if` and `else` branches. Here only `.Status` is set - so it's possible to have one assignment like

`applications[i].Status = i < approvedApplications ? ApplicationStatus.Approved : ApplicationStatus.Rejected;`

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopControllerTests.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/WorkshopServiceTests.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopController.cs
DKymachynskyi: I made the correction if I understood you correctly. Thank you!

VadymLevkovskyi: @DKymachynskyi I was talking about all three methods  (`.GetById()`, `.IsAvailableSeatsValid()`, `.Update()`) , not only first two. 

I'm not insisting on next changes, it's up to you to decide do it or not, I'm just sharing my point.

So currently there is such code in controller
```
        var result = await combinedWorkshopService.IsAvailableSeatsValidForWorkshop(
            dto.AvailableSeats, dto.Id).ConfigureAwait(false);

        if (result is null)
        {
            return BadRequest("The workshop does not exist.");
        }

        if (result == false)
        {
            return BadRequest("The number of available seats must be equal or greater than the number of taken seats");
        }

        try
        {
            var updatingResult = await combinedWorkshopService.Update(dto).ConfigureAwait(false);
            return Ok(CreateUpdateResponse(updatingResult));
        }
        catch (ArgumentException e)
        {
            return BadRequest(e.Message);
        } 
```
my suggestion is to have this code reduced to simple
```
        var result = await combinedWorkshopService.Update(dto).ConfigureAwait(false);

        if (!result.Succeded)
        {
            return BadRequest(result.OperationResult.ToString());
        }

        return Ok(CreateUpdateResponse(result).Value);
```
where all previous checks are moved inside `combinedWorkshopService.Update()` that returns `Task<Result<WorkshopBaseDto>>` and all checks are done inside and errors are stored as `Result.OperationResult.Errors`.

(Minor note between lines - design of `Result` and `OperationResult` types is pretty hectic, but it works, at least)

Thus controller has no logic except dto mapping (that `CreateUpdateResponse()`) and you have better test locality for logic in your code.

OutOfSchool/OutOfSchool.BusinessLogic/Services/IWorkshopServicesCombiner.cs
```diff
@@ -130,4 +137,12 @@ public interface IWorkshopServicesCombiner
...
     /// <returns><see cref="IEnumerable{T}"/> of Workshops for the specified provider.</returns>
     Task<IEnumerable<ShortEntityDto>> UpdateProviderStatus(Guid providerId, ProviderStatus providerStatus);
+
+    /// <summary>
+    /// Checks if the given available seats value is valid for the specified workshop.
+    /// </summary>
+    /// <param name="availableSeats">The number of available seats to validate.</param>
+    /// <param name="workshop">The workshop for which the available seats value is being validated.</param>
+    /// <returns> A boolean value indicating whether the available seats value is valid for the workshop.</returns>
+    bool IsAvailableSeatsValidForWorkshop(uint? availableSeats, WorkshopDto workshop);
```
VadymLevkovskyi: (Design) This method seems to be never used in code. Only in tests. Could be removed from interface

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopControllerTests.cs
```diff
@@ -495,7 +500,12 @@ public async Task UpdateWorkshop_WhenModelIsValid_ShouldReturnOkObjectResult()
     public async Task UpdateWorkshop_WhenModelIsInvalid_ShouldReturnBadRequestObjectResult()
     {
         // Arrange
-        workshopServiceMoq.Setup(x => x.Update(workshopUpdateDto)).ReturnsAsync(workshopUpdateDto);
+        //workshopServiceMoq.Setup(x => x.Update(workshopUpdateDto))
+        //    .ReturnsAsync(Result<WorkshopBaseDto>.Failed(new OperationError
+        //    {
+        //        Code = HttpStatusCode.BadRequest.ToString(),
+        //        Description = "Workshop not found.",
+        //    }));
```
VadymLevkovskyi: (Commented code) Please don't leave commented out code. For history you could use git. For knowledge sharing - you should add more explanations than just a bare code.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopControllerTests.cs
```diff
@@ -523,6 +534,64 @@ public async Task UpdateWorkshop_WhenIdProviderHasNoRights_ShouldReturn403Object
...
+
+    //    // Act
+    //    var result = await controller.Update(workshopUpdateDto).ConfigureAwait(false) as ObjectResult;
+
+    //    // Assert
+    //    providerServiceMoq.VerifyAll();
+    //    workshopServiceMoq.Verify(x => x.Update(It.IsAny<WorkshopBaseDto>()), Times.Never);
+    //    Assert.IsNotNull(result);
+    //    Assert.AreEqual(BadRequest, result.StatusCode);
+    //}
```
VadymLevkovskyi: (Commented code) Please don't leave commented out code.

VadymLevkovskyi (changes requested):

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopControllerTests.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopControllerTests.cs
DKymachynskyi: Fixed. Thank you!

VadymLevkovskyi (commented):
The final step left - test coverage

DKymachynskyi:
@VadymLevkovskyi, I've covered by tests all new code that refers to V1. Should I write tests for V2?

VadymLevkovskyi:
> @VadymLevkovskyi, I've covered by tests all new code that refers to V1. Should I write tests for V2?

It would be a good for the project.

VadymLevkovskyi (approved):

PR #1489, opened 2024-06-04
Fix retrieving parent personal info

MrVenik:


OutOfSchool/OutOfSchool.BusinessLogic/Services/UserService.cs
```diff
@@ -48,23 +48,20 @@ public async Task<IEnumerable<ShortUserDto>> GetAll()
...
-    // TODO: use repository.GetById() method
     public async Task<ShortUserDto> GetById(string id)
     {
-        logger.LogInformation($"Getting User by Id started. Looking Id = {id}.");
+        logger.LogInformation("Getting User by Id started. Looking Id = {Id}.", id);
 
-        Expression<Func<User, bool>> filter = p => p.Id == id;
-
-        var users = await repository.GetByFilter(filter).ConfigureAwait(false);
+        var user = await repository.GetById(id).ConfigureAwait(false);
```
VadymLevkovskyi: (Performance) Interesting change of code. 
I presume it changes nothing in logic, but as you've touched this place I  have to check `.AsNoTracking()` moment (as retrieved `user` is then mapped to `Dto` we seems to need no tracking, right?). 

Could you confirm that `repository.GetById()` doesn't use tracking?
If not - could you rewrite it to version without tracking?

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/PersonalInfoController.cs
```diff
@@ -42,17 +42,14 @@ public class PersonalInfoController : ControllerBase
...
         {
             info = await parentService.GetPersonalInfoByUserId(currentUserService.UserId);
         }
-        else
-        {
-            info = await userService.GetById(currentUserService.UserId);
-        }
 
-        return info is null ? NoContent() : Ok(info);
+        return Ok(info ?? (await userService.GetById(currentUserService.UserId)));
```
VadymLevkovskyi: (Logic) You've slightly changed logic of this method for case when `parentService.GetPersonalInfoByUserId` returns `null`. I'm not sure if it was intended. Please clarify.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/PersonalInfoController.cs
```diff
@@ -42,17 +42,14 @@ public class PersonalInfoController : ControllerBase
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     public async Task<IActionResult> GetPersonalInfo()
     {
-        ShortUserDto info;
+        var info = null as ShortUserDto;
+
         if (currentUserService.IsInRole(Role.Parent))
         {
             info = await parentService.GetPersonalInfoByUserId(currentUserService.UserId);
         }
```
VadymLevkovskyi: (Code style, IMHO, don't insist on changes) Please consider usage of ternary operator here, instead of defining variable, setting it to `null` and then immediately in next line conditionally setting it to other value.

My preferred way to write such code is:
```
        var info = currentUserService.IsInRole(Role.Parent)
            ? await parentService.GetPersonalInfoByUserId(currentUserService.UserId)
            : null;
        info ??= await userService.GetById(currentUserService.UserId);
``` 
Benefits:
- you don't need that weird `null as ShortUserDto`;
- you immediately see that `if` is needed only for setting the same variable;
- less code.

VadymLevkovskyi (commented):
Minor comments, might be skipped, but I'd like to get answer on comment re changed logic.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/PersonalInfoController.cs
vpustovoitov: In our recent daily meeting, we discussed the importance of personal information for the proper functioning of the app’s frontend. As part of our registration logic update, we now create users without parents initially, allowing them to complete their registration later. However, we need to adjust this process. When a user in the parent role, who is not fully registered, attempts to retrieve their personal information, the current endpoint returns a 403 forbidden error. This behavior is incorrect, so I removed the call to the `HasRights` method in the parent service. Additionally, the frontend assumes that all users have some related information (such as name and role), even if they are not fully registered.

VadymLevkovskyi: Thanks for explanation!

VadymLevkovskyi (approved):

PR #1493, opened 2024-06-08
Glebshpuryk/1491

ShpurykGleb:
Now, we have an endpoint, that can gives you a collection of ApplicationDto with status "Pending".

Firstly, logic searching a standard provider with given id, if this provider does not exist - logic is going to find an admin provider. If there is no provider with give id - endpoint returns 400, else, if all is ok - 200 (collection of ApplicationDto with pending status)

Example path - /api/v1/provider/{providerId}/applications/pending

Additions:
- Writed 3 tests for the endpoint
- Added new extension method for SearchResult, that checks it for being null and his entities for being empty
- Added new extension method fot ControllerBase, that checks instance of SearchResult for being null and his entities for being empty, and, after this, decides, what to return - NoContent() or Ok()

### DIFF
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Extensions/SearchResultExtensions.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Extensions/SearchResultExtensions.cs
@@ -3,17 +3,6 @@
 namespace OutOfSchool.BusinessLogic.Extensions;
 public static class SearchResultExtensions
 {
-    /// <summary>
-    /// Method, that checks SearchResult for null and his entities for empty.
-    /// </summary>
-    /// <typeparam name="T">Typeof entity.</typeparam>
-    /// <param name="searchResult">Instance of SearchResult.</param>
-    /// <returns>
-    /// True - searchResult is null or searchResult.Entities is empty.
-    /// False - searchResult is not null or searchResult.Entities is not empty.
-    /// </returns>
     public static bool IsNullOrEntitiesEmpty<T>(this SearchResult<T> searchResult)
-    {
-        return searchResult == null || searchResult.Entities.Count == 0;
-    }
+        => searchResult == null || searchResult.Entities.Count == 0;
 }
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Extensions/SearchResultExtensions.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Extensions/SearchResultExtensions.cs
@@ -1,0 +1,19 @@
+﻿using OutOfSchool.BusinessLogic.Models;
+
+namespace OutOfSchool.BusinessLogic.Extensions;
+public static class SearchResultExtensions
+{
+    /// <summary>
+    /// Method, that checks SearchResult for null and his entities for empty.
+    /// </summary>
+    /// <typeparam name="T">Typeof entity.</typeparam>
+    /// <param name="searchResult">Instance of SearchResult.</param>
+    /// <returns>
+    /// True - searchResult is null or searchResult.Entities is empty.
+    /// False - searchResult is not null or searchResult.Entities is not empty.
+    /// </returns>
+    public static bool IsNullOrEntitiesEmpty<T>(this SearchResult<T> searchResult)
+    {
+        return searchResult == null || searchResult.Entities.Count == 0;
+    }
+}
--- a/OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
+++ b/OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
@@ -35,8 +35,6 @@ public class ApplicationControllerTests
     private string userId;
     private Guid providerId;
     private Mock<HttpContext> httpContext;
-    private ApplicationFilter filter;
-    private SearchResult<ApplicationDto> emptySearchResult;
 
 
     private List<ApplicationDto> applications;
@@ -83,9 +81,6 @@ public class ApplicationControllerTests
         provider.UserId = userId;
         provider.Id = providerId;
         applications = ApplicationDTOsGenerator.Generate(2).WithWorkshopCard(workshops.First()).WithParent(parent);
-
-        filter = new ApplicationFilter();
-        emptySearchResult = new SearchResult<ApplicationDto>() { TotalAmount = 0, Entities = new List<ApplicationDto>() };
     }
 
     [Test]
@@ -169,6 +164,7 @@ public class ApplicationControllerTests
     public void GetByParentId_WhenParentHasNoRights_ShouldThrowUnauthorizedAccess()
     {
         // Arrange
+        var filter = new ApplicationFilter();
         httpContext.Setup(c => c.User.IsInRole("parent")).Returns(true);
         List<ApplicationDto> app = applications.Where(a => a.ParentId == parent.Id).ToList();
         applicationService.Setup(s => s.GetAllByParent(parent.Id, It.IsAny<ApplicationFilter>()))
@@ -266,8 +262,10 @@ public class ApplicationControllerTests
     [Test]
     public async Task GetByWorkshopId_WhenIdIsNotValid_ShouldReturnBadRequest()
     {
-        // Act
+        // Arrange
+        var filter = new ApplicationFilter();
 
+        // Act
         var result = await controller.GetByWorkshopId(Guid.Parse("f25d1bfc-8ebc-4087-b1c3-8dbb7964222d"), filter).ConfigureAwait(false) as BadRequestObjectResult;
 
         // Assert
@@ -278,8 +276,10 @@ public class ApplicationControllerTests
     [Test]
     public async Task GetByProviderId_WhenIdIsNotValid_ShouldReturnBadRequest()
     {
-        // Act
+        // Arrange
+        var filter = new ApplicationFilter();
 
+        // Act
         var result = await controller.GetByProviderId(Guid.Parse("83caa2e6-902a-43b5-9744-8a9d66604666"), filter).ConfigureAwait(false) as BadRequestObjectResult;
 
         // Assert
@@ -292,6 +292,7 @@ public class ApplicationControllerTests
     {
         // Arrange
         var id = Guid.Parse("f25d1bfc-8ebc-4087-b1c3-8dbb7964222d");
+        var filter = new ApplicationFilter();
 
         var newProvider = new ProviderDto { Id = new Guid("83caa2e6-902a-43b5-9744-8a9d66604666"), UserId = userId };
         var newWorkshop = new WorkshopDto { Id = new Guid("94b81fa7-180f-4965-8aac-908a9f3ecb8d"), ProviderId = new Guid("83caa2e6-902a-43b5-9744-8a9d66604666") };
@@ -321,6 +322,7 @@ public class ApplicationControllerTests
     {
         // Arrange
         var newProviderStandard = new ProviderDto { Id = providerId, UserId = userId };
+        var filter = new ApplicationFilter();
 
         httpContext.Setup(c => c.User.IsInRole("provider")).Returns(true);
 
@@ -346,6 +348,9 @@ public class ApplicationControllerTests
     public async Task GetPendingApplicationsByProviderId_WhenProvidersNotExist_ShouldReturnBadRequest()
     {
         // Arrange
+        var filter = new ApplicationFilter();
+        var emptySearchResult = new SearchResult<ApplicationDto>() { TotalAmount = 0, Entities = new List<ApplicationDto>() };
+
         httpContext.Setup(c => c.User.IsInRole("provider")).Returns(true);
 
         providerService.Setup(s => s.GetById(providerId)).ReturnsAsync((ProviderDto)null);
@@ -388,6 +393,7 @@ public class ApplicationControllerTests
     {
         // Arrange
         var id = Guid.Parse("83caa2e6-902a-43b5-9744-8a9d66604666");
+        var filter = new ApplicationFilter();
 
         var newProvider = new ProviderDto { Id = new Guid("83caa2e6-902a-43b5-9744-8a9d66604666"), UserId = userId };
         var newWorkshop = new WorkshopDto { Id = new Guid("94b81fa7-180f-4965-8aac-908a9f3ecb8d"), ProviderId = new Guid("83caa2e6-902a-43b5-9744-8a9d66604666") };
@@ -417,6 +423,7 @@ public class ApplicationControllerTests
     {
         // Arrange
         var id = Guid.Parse("f25d1bfc-8ebc-4087-b1c3-8dbb7964222d");
+        var filter = new ApplicationFilter();
 
         httpContext.Setup(c => c.User.IsInRole("provider")).Returns(true);
 
@@ -433,6 +440,7 @@ public class ApplicationControllerTests
     {
         // Arrange
         var id = Guid.Parse("83caa2e6-902a-43b5-9744-8a9d66604666");
+        var filter = new ApplicationFilter();
         httpContext.Setup(c => c.User.IsInRole("provider")).Returns(true);
         providerService.Setup(s => s.GetById(id)).ReturnsAsync(new ProviderDto());
         applicationService.Setup(s => s.GetAllByProvider(id, filter))
@@ -446,6 +454,7 @@ public class ApplicationControllerTests
     public async Task GetByWorkshopId_WhenThereIsNoWorkshopWithId_ShouldReturnBadRequest()
     {
         // Arrange
+        var filter = new ApplicationFilter();
 
         // Act
         var result = await controller.GetByWorkshopId(Guid.Parse("f25d1bfc-8ebc-4087-b1c3-8dbb7964222d"), filter).ConfigureAwait(false) as BadRequestObjectResult;
--- a/OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
+++ b/OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
@@ -8,6 +8,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Moq;
 using NUnit.Framework;
+using OutOfSchool.AuthCommon.Services;
 using OutOfSchool.BusinessLogic.Models;
 using OutOfSchool.BusinessLogic.Models.Application;
 using OutOfSchool.BusinessLogic.Models.Providers;
@@ -322,20 +323,17 @@ public class ApplicationControllerTests
     {
         // Arrange
         var newProviderStandard = new ProviderDto { Id = providerId, UserId = userId };
-        var filter = new ApplicationFilter();
-
-        httpContext.Setup(c => c.User.IsInRole("provider")).Returns(true);
 
         providerService.Setup(s => s.GetById(providerId)).ReturnsAsync(newProviderStandard);
-
-        var providerApplications = applications.Where(a => a.Workshop.ProviderId == providerId).ToList();
-        applicationService.Setup(s => s.GetAllByProvider(providerId, filter))
-            .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = providerApplications.Count, Entities = providerApplications });
 
         // Act
         var result = await controller.GetPendingApplicationsByProviderId(providerId).ConfigureAwait(false);
 
         // Assert
+        providerService.VerifyAll();
+        applicationService.VerifyAll();
+
+
         result.Should().NotBeNull();
         result.Should()
               .BeOfType<NoContentResult>()
@@ -351,40 +349,41 @@ public class ApplicationControllerTests
         var filter = new ApplicationFilter();
         var emptySearchResult = new SearchResult<ApplicationDto>() { TotalAmount = 0, Entities = new List<ApplicationDto>() };
 
-        httpContext.Setup(c => c.User.IsInRole("provider")).Returns(true);
-
         providerService.Setup(s => s.GetById(providerId)).ReturnsAsync((ProviderDto)null);
         providerAdminService.Setup(s => s.GetById(providerId.ToString())).ReturnsAsync((ProviderAdminProviderRelationDto)null);
 
-        applicationService.Setup(s => s.GetAllByProvider(providerId, filter))
-            .ReturnsAsync(emptySearchResult);
-        applicationService.Setup(s => s.GetAllByProviderAdmin(providerId.ToString(), filter, providerId, It.IsAny<bool>()))
-            .ReturnsAsync(emptySearchResult);
-
         // Act
-        var result = await controller.GetPendingApplicationsByProviderId(providerId).ConfigureAwait(false) as BadRequestObjectResult;
+        var result = await controller.GetPendingApplicationsByProviderId(providerId).ConfigureAwait(false);
 
         // Assert
+        providerService.VerifyAll();
+        providerAdminService.VerifyAll();
+
         result.Should().NotBeNull();
         result.Should()
              .BeOfType<BadRequestObjectResult>()
              .Which.StatusCode
              .Should()
              .Be(StatusCodes.Status400BadRequest);
-        result.Value.Should().Be($"There is no any provider with given id - {providerId}.");
+        result.As<BadRequestObjectResult>()
+            .Value
+            .Should()
+            .Be($"There is no any provider with given id - {providerId}.");
     }
 
     [Test]
     public void GetPendingApplicationsByProviderId_WhenProviderHasNoRights_ShouldThrowUnauthorizedAccess()
     {
-        httpContext.Setup(c => c.User.IsInRole("provider")).Returns(true);
-
+        // Arrange
         providerAdminService.Setup(s => s.GetById(providerId.ToString())).ReturnsAsync(new ProviderAdminProviderRelationDto());
 
         applicationService.Setup(s => s.GetAllByProviderAdmin(It.IsAny<string>(), It.IsAny<ApplicationFilter>(), It.IsAny<Guid>(), It.IsAny<bool>()))
             .ThrowsAsync(new UnauthorizedAccessException());
 
         // Act & Assert
+        providerAdminService.Verify(s => s.GetById(providerId.ToString()), Times.Never);
+        applicationService.Verify(s => s.GetAllByProviderAdmin(It.IsAny<string>(), It.IsAny<ApplicationFilter>(), It.IsAny<Guid>(), It.IsAny<bool>()), Times.Never);
+
         Assert.ThrowsAsync<UnauthorizedAccessException>(async () => await controller.GetPendingApplicationsByProviderId(providerId));
     }
 
--- a/OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
+++ b/OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
@@ -328,9 +328,9 @@ public class ApplicationControllerTests
 
         providerService.Setup(s => s.GetById(providerId)).ReturnsAsync(newProviderStandard);
 
-        var app1 = applications.Where(a => a.Workshop.ProviderId == providerId).ToList();
+        var providerApplications = applications.Where(a => a.Workshop.ProviderId == providerId).ToList();
         applicationService.Setup(s => s.GetAllByProvider(providerId, filter))
-            .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = app1.Count, Entities = app1 });
+            .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = providerApplications.Count, Entities = providerApplications });
 
         // Act
         var result = await controller.GetPendingApplicationsByProviderId(providerId).ConfigureAwait(false);
--- a/OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
+++ b/OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
@@ -36,6 +36,7 @@ public class ApplicationControllerTests
     private Guid providerId;
     private Mock<HttpContext> httpContext;
 
+
     private List<ApplicationDto> applications;
     private IEnumerable<ChildDto> children;
     private IEnumerable<WorkshopCard> workshops;
@@ -317,30 +318,29 @@ public class ApplicationControllerTests
     [Test]
     public async Task GetPendingApplicationsByProviderId_WhenProviderDoesNotHaveAnyApplications_ShouldReturnNoContent()
     {
-        var id = Guid.Parse("83caa2e6-902a-43b5-9744-8a9d66604666");
         var filter = new ApplicationFilter();
 
-        var newProviderStandard = new ProviderDto { Id = new Guid("83caa2e6-902a-43b5-9744-8a9d66604666"), UserId = userId };
-        var newProviderAdmin = new ProviderAdminProviderRelationDto { ProviderId = new Guid("83caa2e6-902a-43b5-9744-8a9d66604666"), UserId = userId };
-        var newWorkshop = new WorkshopDto { Id = new Guid("94b81fa7-180f-4965-8aac-908a9f3ecb8d"), ProviderId = new Guid("83caa2e6-902a-43b5-9744-8a9d66604666") };
+        var newProviderStandard = new ProviderDto { Id = providerId, UserId = userId };
+        var newProviderAdmin = new ProviderAdminProviderRelationDto { ProviderId = providerId, UserId = userId };
+        var newWorkshop = new WorkshopDto { Id = Guid.NewGuid(), ProviderId = providerId };
 
         httpContext.Setup(c => c.User.IsInRole("provider")).Returns(true);
 
         providerService.Setup(s => s.GetByUserId(userId, It.IsAny<bool>())).ReturnsAsync(newProviderStandard);
-        providerAdminService.Setup(s => s.GetById(id.ToString())).ReturnsAsync(newProviderAdmin);
-        workshopService.Setup(s => s.GetById(id, It.IsAny<bool>())).ReturnsAsync(newWorkshop);
-        providerService.Setup(s => s.GetById(id)).ReturnsAsync(newProviderStandard);
-        List<ApplicationDto> app1 = applications.Where(a => a.Workshop.ProviderId == id).ToList();
-        applicationService.Setup(s => s.GetAllByProvider(id, filter))
+        providerAdminService.Setup(s => s.GetById(providerId.ToString())).ReturnsAsync(newProviderAdmin);
+        workshopService.Setup(s => s.GetById(providerId, It.IsAny<bool>())).ReturnsAsync(newWorkshop);
+        providerService.Setup(s => s.GetById(providerId)).ReturnsAsync(newProviderStandard);
+        var app1 = applications.Where(a => a.Workshop.ProviderId == providerId).ToList();
+        applicationService.Setup(s => s.GetAllByProvider(providerId, filter))
             .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = app1.Count(), Entities = app1 });
-        applicationService.Setup(s => s.GetAllByProviderAdmin(id.ToString(), filter, newProviderAdmin.ProviderId, newProviderAdmin.IsDeputy))
+        applicationService.Setup(s => s.GetAllByProviderAdmin(providerId.ToString(), filter, newProviderAdmin.ProviderId, newProviderAdmin.IsDeputy))
             .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = app1.Count(), Entities = app1 });
-        List<ApplicationDto> app2 = applications.Where(a => a.WorkshopId == id).ToList();
-        applicationService.Setup(s => s.GetAllByWorkshop(id, new Guid("83caa2e6-902a-43b5-9744-8a9d66604666"), filter))
+        var app2 = applications.Where(a => a.WorkshopId == providerId).ToList();
+        applicationService.Setup(s => s.GetAllByWorkshop(providerId, providerId, filter))
             .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = app2.Count(), Entities = app2 });
 
         // Act
-        var result = await controller.GetPendingApplicationsByProviderId(id).ConfigureAwait(false) as NoContentResult;
+        var result = await controller.GetPendingApplicationsByProviderId(providerId).ConfigureAwait(false) as NoContentResult;
 
         // Assert
         result.Should().NotBeNull();
@@ -351,13 +351,12 @@ public class ApplicationControllerTests
     public async Task GetPendingApplicationsByProviderId_WhenProvidersNotExist_ShouldReturnBadRequest()
     {
         // Arrange
-        var id = Guid.Parse("83caa2e6-1111-1111-9744-8a9d66604666");
         var filter = new ApplicationFilter();
 
         httpContext.Setup(c => c.User.IsInRole("provider")).Returns(true);
 
-        providerService.Setup(s => s.GetById(id)).ReturnsAsync((ProviderDto)null);
-        providerAdminService.Setup(s => s.GetById(id.ToString())).ReturnsAsync((ProviderAdminProviderRelationDto)null);
+        providerService.Setup(s => s.GetById(providerId)).ReturnsAsync((ProviderDto)null);
+        providerAdminService.Setup(s => s.GetById(providerId.ToString())).ReturnsAsync((ProviderAdminProviderRelationDto)null);
 
         applicationService.Setup(s => s.GetAllByProvider(It.IsAny<Guid>(), It.IsAny<ApplicationFilter>()))
             .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = 0, Entities = new List<ApplicationDto>() });
@@ -365,25 +364,24 @@ public class ApplicationControllerTests
             .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = 0, Entities = new List<ApplicationDto>() });
 
         // Act
-        var result = await controller.GetPendingApplicationsByProviderId(id).ConfigureAwait(false) as BadRequestObjectResult;
+        var result = await controller.GetPendingApplicationsByProviderId(providerId).ConfigureAwait(false) as BadRequestObjectResult;
 
         // Assert
         result.Should().NotBeNull();
         result.StatusCode.Should().Be(StatusCodes.Status400BadRequest);
-        result.Value.Should().Be($"There is no any provider with Id = {id}");
+        result.Value.Should().Be($"There is no any provider with given id - {providerId}.");
     }
 
     [Test]
     public void GetPendingApplicationsByProviderId_WhenProviderHasNoRights_ShouldThrowUnauthorizedAccess()
     {
         // Arrange
-        var id = Guid.Parse("83caa2e6-1111-1111-9744-8a9d66604666");
         var filter = new ApplicationFilter();
 
         httpContext.Setup(c => c.User.IsInRole("provider")).Returns(true);
 
-        providerService.Setup(s => s.GetById(id)).ReturnsAsync(new ProviderDto());
-        providerAdminService.Setup(s => s.GetById(id.ToString())).ReturnsAsync(new ProviderAdminProviderRelationDto());
+        providerService.Setup(s => s.GetById(providerId)).ReturnsAsync(new ProviderDto());
+        providerAdminService.Setup(s => s.GetById(providerId.ToString())).ReturnsAsync(new ProviderAdminProviderRelationDto());
 
         applicationService.Setup(s => s.GetAllByProvider(It.IsAny<Guid>(), It.IsAny<ApplicationFilter>()))
            .ThrowsAsync(new UnauthorizedAccessException());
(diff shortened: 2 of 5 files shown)


OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
```diff
@@ -82,7 +82,7 @@ public void Setup()
         applications = ApplicationDTOsGenerator.Generate(2).WithWorkshopCard(workshops.First()).WithParent(parent);
     }
 
-    
+
```
VadymLevkovskyi: (Code style) Please avoid adding extra new lines. 1 empty line is sufficient spacer.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
```diff
@@ -316,12 +316,92 @@ public async Task GetByWorkshopId_WhenProviderHasNoApplications_ShouldReturnNoCo
         result.StatusCode.Should().Be(StatusCodes.Status204NoContent);
     }
 
+    [Test]
+    public async Task GetPendingApplicationsByProviderId_WhenProviderDoesNotHaveAnyApplications_ShouldReturnNoContent()
+    {
+        var id = Guid.Parse("83caa2e6-902a-43b5-9744-8a9d66604666");
```
VadymLevkovskyi: (Code style) Magic number/string literals are usually not very usable. 
Please consider writing using it only once. I.e. - guid `83caa2e6...` is used near 6 times. And currently tthere are two ways to ensure that it is the same one: to use search functionality, or break your eyes comparing them symbol-by-symbol. Neither is good, neither is checked by compiler, neither gives me understanding what this guid is used for.  If you set it only once to const/var with a good name and reuse it - it would be much better.

(Test logic) Please avoid using hard-coded test data (as this and other guids). You could either use `new Guid()` that gives you unique guid every time, or use Bogus library.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
```diff
@@ -316,12 +316,92 @@ public async Task GetByWorkshopId_WhenProviderHasNoApplications_ShouldReturnNoCo
...
+        providerService.Setup(s => s.GetByUserId(userId, It.IsAny<bool>())).ReturnsAsync(newProviderStandard);
+        providerAdminService.Setup(s => s.GetById(id.ToString())).ReturnsAsync(newProviderAdmin);
+        workshopService.Setup(s => s.GetById(id, It.IsAny<bool>())).ReturnsAsync(newWorkshop);
+        providerService.Setup(s => s.GetById(id)).ReturnsAsync(newProviderStandard);
+        List<ApplicationDto> app1 = applications.Where(a => a.Workshop.ProviderId == id).ToList();
+        applicationService.Setup(s => s.GetAllByProvider(id, filter))
+            .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = app1.Count(), Entities = app1 });
+        applicationService.Setup(s => s.GetAllByProviderAdmin(id.ToString(), filter, newProviderAdmin.ProviderId, newProviderAdmin.IsDeputy))
+            .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = app1.Count(), Entities = app1 });
+        List<ApplicationDto> app2 = applications.Where(a => a.WorkshopId == id).ToList();
```
VadymLevkovskyi: (Code style) Please avoid explicitely typed variable definition. Instead consider usage of `var`. This way you'll get less code, code is easier to modify, no hidden unexpected cast.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
```diff
@@ -316,12 +316,92 @@ public async Task GetByWorkshopId_WhenProviderHasNoApplications_ShouldReturnNoCo
...
+        applicationService.Setup(s => s.GetAllByWorkshop(id, new Guid("83caa2e6-902a-43b5-9744-8a9d66604666"), filter))
+            .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = app2.Count(), Entities = app2 });
+
+        // Act
+        var result = await controller.GetPendingApplicationsByProviderId(id).ConfigureAwait(false) as NoContentResult;
+
+        // Assert
```
VadymLevkovskyi, line 331: (Test code) You have lot of `someMoq.Setup(...)` in `Arrange` part. Please check that they all are used in `Assert` part using `someMoq.VerifyAll()`. This way you'll immediately see excessive `.Setup`s and could remove them

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
```diff
@@ -316,12 +316,92 @@ public async Task GetByWorkshopId_WhenProviderHasNoApplications_ShouldReturnNoCo
...
+        applicationService.Setup(s => s.GetAllByProvider(id, filter))
+            .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = app1.Count(), Entities = app1 });
+        applicationService.Setup(s => s.GetAllByProviderAdmin(id.ToString(), filter, newProviderAdmin.ProviderId, newProviderAdmin.IsDeputy))
+            .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = app1.Count(), Entities = app1 });
+        List<ApplicationDto> app2 = applications.Where(a => a.WorkshopId == id).ToList();
+        applicationService.Setup(s => s.GetAllByWorkshop(id, new Guid("83caa2e6-902a-43b5-9744-8a9d66604666"), filter))
+            .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = app2.Count(), Entities = app2 });
+
+        // Act
+        var result = await controller.GetPendingApplicationsByProviderId(id).ConfigureAwait(false) as NoContentResult;
```
VadymLevkovskyi: (Test code) Type casting shouldn't be a part of `Act`, because it's not how we use it in real code.
Instead of type casting consider usage of fluent assertion to check both type and content:
 ```
   result.Should().BeOfType<NoContentResult>()
        .Which.StatusCode.Should().Be...
```

File: OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs [hunk 6]
```diff
@@ -316,12 +316,92 @@ public async Task GetByWorkshopId_WhenProviderHasNoApplications_ShouldReturnNoCo
...
+        result.Should().NotBeNull();
+        result.StatusCode.Should().Be(StatusCodes.Status400BadRequest);
+        result.Value.Should().Be($"There is no any provider with Id = {id}");
+    }
+
+    [Test]
+    public void GetPendingApplicationsByProviderId_WhenProviderHasNoRights_ShouldThrowUnauthorizedAccess()
+    {
+        // Arrange
+        var id = Guid.Parse("83caa2e6-1111-1111-9744-8a9d66604666");
```
VadymLevkovskyi: (Naming) `id` here iseems to be `providerId`. Could name it appropriately?

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
VadymLevkovskyi: Same issue in test below

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
VadymLevkovskyi: Same issue in test below

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
```diff
@@ -316,12 +316,92 @@ public async Task GetByWorkshopId_WhenProviderHasNoApplications_ShouldReturnNoCo
...
+        var filter = new ApplicationFilter();
+
+        httpContext.Setup(c => c.User.IsInRole("provider")).Returns(true);
+
+        providerService.Setup(s => s.GetById(id)).ReturnsAsync((ProviderDto)null);
+        providerAdminService.Setup(s => s.GetById(id.ToString())).ReturnsAsync((ProviderAdminProviderRelationDto)null);
+
+        applicationService.Setup(s => s.GetAllByProvider(It.IsAny<Guid>(), It.IsAny<ApplicationFilter>()))
+            .ReturnsAsync(new SearchResult<ApplicationDto>() { TotalAmount = 0, Entities = new List<ApplicationDto>() });
```
VadymLevkovskyi: (Code duplication) This `new SearchResult<ApplicationDto>() { TotalAmount = 0, Entities = new List<ApplicationDto>() }` could be introduced as `emptyApplications` (or give it better name) variable and reused. Less code, more self-explanatory.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
```diff
@@ -316,12 +316,92 @@ public async Task GetByWorkshopId_WhenProviderHasNoApplications_ShouldReturnNoCo
...
+        // Arrange
+        var id = Guid.Parse("83caa2e6-1111-1111-9744-8a9d66604666");
+        var filter = new ApplicationFilter();
+
+        httpContext.Setup(c => c.User.IsInRole("provider")).Returns(true);
+
+        providerService.Setup(s => s.GetById(id)).ReturnsAsync((ProviderDto)null);
+        providerAdminService.Setup(s => s.GetById(id.ToString())).ReturnsAsync((ProviderAdminProviderRelationDto)null);
+
+        applicationService.Setup(s => s.GetAllByProvider(It.IsAny<Guid>(), It.IsAny<ApplicationFilter>()))
```
VadymLevkovskyi: (Test logic) Please avoid overusing of `It.IsAny<...>()`. It gives no understanding what test is supposed to check. 
Here we are checking specific provider id - please pass it to `Setup` instead of `It.IsAny<Guid>()`. 
Use this rule wherever possible.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -409,6 +470,24 @@ public async Task<IActionResult> AllowedToReview(Guid parentId, Guid workshopId)
...
 
+    /// <summary>
+    /// Method, that checks applications of SearchResult for 204 status code.
+    /// </summary>
+    /// <param name="applications">Applications of SearchResult</param>
+    /// <returns>
+    /// True - applications is null or applications is empty.
+    /// False - applications is not null and applications collection has minimum 1 application.
+    /// </returns>
+    private static bool CheckApplicationsFor204(SearchResult<ApplicationDto> applications)
```
VadymLevkovskyi: (Code style, naming, placement) This method body doesn't check for 204. It checks for empty data in `SearchResult`. Thus it could be named `IsEmpty`. 
It also doesn't check for specific type `SearchResult<ApplicationDto>`, thus it could be generic 
`bool IsEmpty<T>(SearchResult<T> data)`.
It is not specific to `ApplicationController`, instead it is more related to `SearchResult`. Thus it could either be `SearchResult` method or `SearchResult` extension method.
It is pretty simple, thus it should be analyzed if method extraction is good at all.
As I can see there is constant pattern in this controller actions - "if(isEmpty(x)) return NoContent(); return Ok()". Maybe it could be extracted as method? Maybe other controllers have such pattern as well, and this method could be moved out of `ApplicationController` to `ControllerBase` or `Controller` extension method?
Please consider refactoring

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -409,6 +470,24 @@ public async Task<IActionResult> AllowedToReview(Guid parentId, Guid workshopId)
...
+    {
+        if (applications == null || applications.Entities.Count == 0)
+        {
+            return true;
+        }
+
+        return false;
```
VadymLevkovskyi: (Code style) These 6 lines of code could be rewritten in 1. Please consider refactoring.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -167,14 +168,74 @@ public async Task<IActionResult> GetByProviderId(Guid providerId, [FromQuery] Ap
...
+                                     .ConfigureAwait(false);
+        }
+
+        // If applications is null or empty - return 204
+        if (CheckApplicationsFor204(applications))
+        {
+            return NoContent();
+        }
+
+        return Ok(applications.TotalAmount);
```
VadymLevkovskyi: (Logic) Action name and comment promise to return applications. Instead it returns count. Please fix something.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -167,14 +168,74 @@ public async Task<IActionResult> GetByProviderId(Guid providerId, [FromQuery] Ap
...
+    /// <param name="providerId">Provider id.</param>
+    /// <returns>List of applications.</returns>
+    /// <response code="200">Entities were found by given Id.</response>
+    /// <response code="204">No entity with given Id was found.</response>
+    /// <response code="500">If any server error occurs.</response>
+    [HasPermission(Permissions.ApplicationRead)]
+    [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(SearchResult<ApplicationDto>))]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
```
VadymLevkovskyi: (Comments, OpenApi documentation) List of responses provided in XmlDoc section (`///` comments) is different form list provided in OpenApi section (`ProduceResponseType` attributes). Please align them

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -167,14 +168,74 @@ public async Task<IActionResult> GetByProviderId(Guid providerId, [FromQuery] Ap
...
+    [HttpGet("/api/v{version:apiVersion}/provider/{providerId}/applications/pending")]
+    public async Task<IActionResult> GetPendingApplicationsByProviderId(Guid providerId)
+    {
+        // Find a provider with given id in admins and standard providers
+        var providerStandard = await providerService.GetById(providerId).ConfigureAwait(false);
+
+        var providerAdminIdStringVersion = providerId.ToString();
+        var providerAdmin = await providerAdminService.GetById(providerAdminIdStringVersion).ConfigureAwait(false);
```
VadymLevkovskyi: (Performance) Your code doesn't need `providerAdmin` value when `providerStandard != null`. Consider refactoring.

VadymLevkovskyi (changes requested):
The major question is related to new action logic and there is a pack of minor comments re code style, performance, testing code... (see in separate comments below).

Also there is suggestion to spent extra time on adding some info into PR (current PR's name and description are not very usable).

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
ShpurykGleb: My bad, fixed.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
ShpurykGleb: Fixed.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
ShpurykGleb: Fixed.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
ShpurykGleb: Fixed

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
ShpurykGleb: We already have this variable as class field (providerId). I replaced old versions with this field.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
ShpurykGleb: Refactored.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
ShpurykGleb: Now returns collection. Fixed.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
ShpurykGleb: Refactored.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
ShpurykGleb: I dont understand your request too much for this. But, I think, i done what you want me to do.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
ShpurykGleb: Now method appears as extension method for SearchResult.

Also, i really like your idea, that we can extract construction of "return NoContent(); return Ok();", now this construction is in ControllerExtensions. I saw, that a lot of controllers have this construction too, if you want, I can change this construction in another controllers as another task. I think, it is a good idea.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
ShpurykGleb: Fixed.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
ShpurykGleb: Fixed. But, test, when user has no rights, wont work without It.IsAny. So, it is the only test, that i left without changes in this plan.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
ShpurykGleb: Fixed.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
ShpurykGleb: P.S.

Method is extension for SearchResult, because, if it was like usual method, it will throw an ArgumentNullException, if instance of SearchResult is null.

OutOfSchool/OutOfSchool.BusinessLogic/Extensions/SearchResultExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+public static class SearchResultExtensions
+{
+    /// <summary>
+    /// Method, that checks SearchResult for null and his entities for empty.
+    /// </summary>
+    /// <typeparam name="T">Typeof entity.</typeparam>
+    /// <param name="searchResult">Instance of SearchResult.</param>
+    /// <returns>
+    /// True - searchResult is null or searchResult.Entities is empty.
+    /// False - searchResult is not null or searchResult.Entities is not empty.
```
VadymLevkovskyi: (Comment wording) Please change comment or drop it. It says `False - searchResult is not null or searchResult.Entities is not empty.` but should be `...not null AND ... is not empty` - "and" instead of "or". 

In general - I don't see necessity to explain such code in comments. Current method name is self-explanatory, IMHO.

OutOfSchool/OutOfSchool.BusinessLogic/Extensions/SearchResultExtensions.cs
```diff
@@ -0,0 +1,19 @@
+﻿using OutOfSchool.BusinessLogic.Models;
+
+namespace OutOfSchool.BusinessLogic.Extensions;
+public static class SearchResultExtensions
+{
+    /// <summary>
+    /// Method, that checks SearchResult for null and his entities for empty.
+    /// </summary>
+    /// <typeparam name="T">Typeof entity.</typeparam>
```
VadymLevkovskyi: (Comment wording) Please consider changing comment or dropping it. It says `<typeparam name="T">Typeof entity.</typeparam>` that honestly explains nothing.

OutOfSchool/OutOfSchool.BusinessLogic/Extensions/SearchResultExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// <param name="searchResult">Instance of SearchResult.</param>
+    /// <returns>
+    /// True - searchResult is null or searchResult.Entities is empty.
+    /// False - searchResult is not null or searchResult.Entities is not empty.
+    /// </returns>
+    public static bool IsNullOrEntitiesEmpty<T>(this SearchResult<T> searchResult)
+    {
+        return searchResult == null || searchResult.Entities.Count == 0;
+    }
```
VadymLevkovskyi: (Code style, I don't insist) Consider usage of expression-bodied method syntax. IMHO it fits perfectly for such small methods.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
```diff
@@ -7,6 +7,7 @@
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Moq;
+using Nest;
```
VadymLevkovskyi: (Excessive using) Please check if this `using Nest` is required. Drop it if not.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
```diff
@@ -35,6 +36,9 @@ public class ApplicationControllerTests
     private string userId;
     private Guid providerId;
     private Mock<HttpContext> httpContext;
+    private ApplicationFilter filter;
+    private SearchResult<ApplicationDto> emptySearchResult;
```
VadymLevkovskyi: (Locals as a field) Please consider refactoring. 
This `emptySearchResult` is used only in 1 test. Why can't it be a local variable? Introducing locals as fields you add risk by (future) reusing them in other methods, thus changes done outside might affect the method. For test methods it's even more important, as they should be run isolated.

As for me, I'll do the same for `filter` though it is reused in many tests. The same reason applies: tests locality. 

And you can say that both variables are reset in `.Setup()` but I've already seen problem with this in parallel tests execution case.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
```diff
@@ -316,12 +317,78 @@ public async Task GetByWorkshopId_WhenProviderHasNoApplications_ShouldReturnNoCo
...
+    public async Task GetPendingApplicationsByProviderId_WhenProviderDoesNotHaveAnyApplications_ShouldReturnNoContent()
+    {
+        // Arrange
+        var newProviderStandard = new ProviderDto { Id = providerId, UserId = userId };
+
+        httpContext.Setup(c => c.User.IsInRole("provider")).Returns(true);
+
+        providerService.Setup(s => s.GetById(providerId)).ReturnsAsync(newProviderStandard);
+
+        var app1 = applications.Where(a => a.Workshop.ProviderId == providerId).ToList();
```
VadymLevkovskyi: (Code style) Name `app1`..., oh well. Please avoid usage of number suffix in variables name, especially when domain doesn't require it (and there is no any "app2", "app3" ...)

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
VadymLevkovskyi: I see no usage of `VerifyAll()`. Where was it fixed?

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
VadymLevkovskyi: If yoo are talking about `GetPendingApplicationsByProviderId_WhenProviderHasNoRights_ShouldThrowUnauthorizedAccess` where `Setup` is used to specify `.ThrowsException()` - then I agree. It's right case to use `It.IsAny<>` And thanks for changes

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
VadymLevkovskyi: Thanks for refactoring.

> ... if you want, I can change this construction in another controllers as another task

That's good that you don't increase scope of current task. Would be good to create separate ticket.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
VadymLevkovskyi: Well? some changes were done, but we need more. The 204 status is not mentioned in `[ProducesResponseType]`. But I see it in xmldoc comment `<response code="204"...>` and in code.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -167,12 +168,64 @@ public async Task<IActionResult> GetByProviderId(Guid providerId, [FromQuery] Ap
...
+    /// <response code="500">If any server error occurs.</response>
+    [HasPermission(Permissions.ApplicationRead)]
+    [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(SearchResult<ApplicationDto>))]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpGet("/api/v{version:apiVersion}/provider/{providerId}/applications/pending")]
+    public async Task<IActionResult> GetPendingApplicationsByProviderId(Guid providerId)
+    {
+        // Find a standard provider by given id
```
VadymLevkovskyi: (Comments) Please avoid writing comments that duplicate code. I mean next example
`i += 1; // here we increase our counter by 1`
- it is really obvious and doesn't need to be explained in comment. 

There are several issues with all comments: 
- they increase code "reading" complexity (yes, you have to read more), 
- they become obsolete pretty fast (code is changed, comment is not - have seen such situaton many times in this project),
- they make refactoring more hard (or welcome obsolete comments),
- they are never checked by compiler or unit test and thus might be simply wrong (seen even in this PR).

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
```diff
@@ -167,12 +168,64 @@ public async Task<IActionResult> GetByProviderId(Guid providerId, [FromQuery] Ap
...
+
+            // Standard and admin providers were not found by given id
+            if (providerAdmin is null)
+            {
+                return BadRequest($"There is no any provider with given id - {providerId}.");
+            }
+
+            applications = await applicationService
+                                     .GetAllByProviderAdmin(providerAdminIdStringVersion, filter, providerAdmin.ProviderId, providerAdmin.IsDeputy)
+                                     .ConfigureAwait(false);
```
VadymLevkovskyi: (Code style, code outlining) Please use standard outlining - 1 tab for wrapped line of code, not n+1 tab.
I.e. for the current code it should be
```
            applications = await applicationService
                .GetAllByProviderAdmin(....)
                .ConfigureAwait(false);
```
instead of current 
```
            applications = await applicationService
                                     .GetAllByProviderAdmin(....)
                                     .ConfigureAwait(false);
```
Idk, maybe it's your IDE configuration - please check it.

OutOfSchool/OutOfSchool.WebApi/Extensions/ControllerExtension.cs
```diff
@@ -1,10 +1,33 @@
 ﻿using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Localization;
+using OutOfSchool.BusinessLogic.Models;
 
 namespace OutOfSchool.WebApi.Extensions;
 
 public static class ControllerExtension
 {
+    /// <summary>
+    /// Extension method, that checks searchResult for null of entities empty and returns result of it.
```
VadymLevkovskyi: (Naming) Both comment and method name are clumsy. 

Comment: 
- `<summary> ...returns result of it` - well, If I see this comment then I also see method signature and method return type - `IAsyncResult`, thus your "returns result" doesn't provide more details, and could be just removed
- `<typeparam>Typeof entity...` - sorry, what entity, what is "typeof"? Is it ignored in the method body? If yes, why not state it directly?
- `<returns>NoContent()... Ok()...` - does it really return methods? No, it returns special types like `NoContentResult` and `OkObjectResult`. 

Method name - `GetResultOfGetOperation`
- `Get...` - where it gets? It converts or maps or "casts". "Get" is usually used to state fact of retrieval from something (from http channel, from repo, from db, etc)
- `...Result...` - as mentioned above it says nothing, because everything is result
- `...GetOperation` - please show me the get operation in this method parameters or body. We don't use it, so we shouldn't mention it.

My suggestion (and I don't insist, you might love comments/your way of naming more) - to remove xmldoc  comments for such simple methods at all, and to name it `MapToOkOrNoContent` or even simpler (using asp.net way of naming) `OkOrNoContent` (`OkOrNoContentWhenEmpty` is more descriptive, though I don't like it's length).

OutOfSchool/OutOfSchool.WebApi/Extensions/ControllerExtension.cs
```diff
@@ -1,10 +1,33 @@
...
+    /// </summary>
+    /// <typeparam name="T">Typeof entity.</typeparam>
+    /// <param name="controller">Controller base.</param>
+    /// <param name="searchResult">Instance of SearchResult.</param>
+    /// <returns>
+    /// NoContent() - searchResult is null or his entities is empty.
+    /// Ok() searchResult is not null and his entities is not empty.
+    /// </returns>
+    public static IActionResult GetResultOfGetOperation<T>(this ControllerBase controller, SearchResult<T> searchResult)
```
VadymLevkovskyi: (Test coverage) Could you show me the unit test checking this new method?

VadymLevkovskyi (changes requested):
More comments provided. Some of comments you've resolved are unresolved. 

Technical: please don't resolve not your comments. Because on n-th review iteration I have to check all resolved comment to find out which one were resolved by me (and treated as fixed) and which one were resolved by you "in advance", and thus have to be reverified. Also this gives wrong picture to other reviewers, as they think that everything was fixed and could be reviewed.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
ShpurykGleb: Yes, i am talking about this test.

OutOfSchool/OutOfSchool.WebApi/Extensions/ControllerExtension.cs
ShpurykGleb: Now you can find a tests to this method in MapSearchResultToOkOrNoContentExtensionTests.

But, i am not entirely sure, that i done this correctly. Can you check this and write your opinion?

OutOfSchool/OutOfSchool.WebApi/Extensions/ControllerExtension.cs
ShpurykGleb: Really, you are right. I renamed this method to MapSearchResultToOkOrNoContent.

'SearchResult' is necessary, in my opinion, because we need to know what we are mapping.

Comment - droped.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
ShpurykGleb: It is my bad, i was thinking, that IDE outlining is not correctly, so i added one more tab.

Now it is fixed.

OutOfSchool/OutOfSchool.BusinessLogic/Extensions/SearchResultExtensions.cs
ShpurykGleb: Comment - deleted.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
ShpurykGleb: Fixed. Now it is providerApplications.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
ShpurykGleb: Fixed. For filter, i was found, that all tests have the same filter, so, i extract this like a field. Now it is created locally for every test, like in previous versions.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
ShpurykGleb: Fixed.

OutOfSchool/OutOfSchool.BusinessLogic/Extensions/SearchResultExtensions.cs
ShpurykGleb: Fixed. Now it is lambda.

OutOfSchool/OutOfSchool.BusinessLogic/Extensions/SearchResultExtensions.cs
ShpurykGleb: Comment - deleted.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
ShpurykGleb: Added 204 response. Fixed.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
ShpurykGleb: After we will finish this task, you can write this ticket and assign it to me.

ShpurykGleb:
> More comments provided. Some of comments you've resolved are unresolved.
> 
> Technical: please don't resolve not your comments. Because on n-th review iteration I have to check all resolved comment to find out which one were resolved by me (and treated as fixed) and which one were resolved by you "in advance", and thus have to be reverified. Also this gives wrong picture to other reviewers, as they think that everything was fixed and could be reviewed.

Sorry, it is my first time, when i working with comments, like this. I was thinking, that i need to close this comments, after my changes, by myself. This will not happen again)

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ApplicationController.cs
ShpurykGleb: Fixed.

OutOfSchool/OutOfSchool.WebApi/Extensions/ControllerExtension.cs
VadymLevkovskyi: Looks good though bit long. Thanks.

OutOfSchool/OutOfSchool.WebApi/Extensions/ControllerExtension.cs
VadymLevkovskyi: Thanks for tests. Will comment them separately, current comment is resolved

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/MapSearchResultToOkOrNoContentExtensionTests.cs
```diff
@@ -0,0 +1,90 @@
...
+using OutOfSchool.WebApi.Extensions;
+
+namespace OutOfSchool.WebApi.Tests.Extensions;
+
+[TestFixture]
+public class MapSearchResultToOkOrNoContentExtensionTests
+{
+    private class TestController : ControllerBase
+    {
+    }
```
VadymLevkovskyi: (Excessive class) `TestController` don't change `ControllerBase`, thus latter could be used, and `TestController` is not needed. 
(Naming) JFYI, as class will be deleted: This test checks extension, not the controller. Thus controller is needed as parameter for the extension method. Thus it should be named not `TestController` but `Fake...` or `Mock...`

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/MapSearchResultToOkOrNoContentExtensionTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+    private TestController controller;
+
+    [SetUp]
+    public void SetUp()
+    {
+        controller = new TestController();
+    }
```
VadymLevkovskyi: (Test locality, not insisting) `controller` could be local variable in. We use it only three times and in some test mocking will be one-liner. So from code lines count the field + `Setup()` method it's not really effective.

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/MapSearchResultToOkOrNoContentExtensionTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+    [Test]
+    public void MapSearchResultToOkOrNoContentExtension_SearchResultIsNotNullAndNotEmpty_ReturnOk()
+    {
+        // Arrange
+        var applications = new List<ApplicationDto>() {
+            new ApplicationDto() { RejectionMessage = "Test 1" },
+            new ApplicationDto() { RejectionMessage = "Test 2", },
+            new ApplicationDto() { RejectionMessage = "Test 3" },
+        };
```
VadymLevkovskyi: (Test data generation) For test data you should use principle of YAGNI. 
I.e. here `RejectionMessage` value doesn't affect the test, thus you shouldn't fill it explicitely.
For efficient test data generation you could use Bogus library. I.e. to generate list of 3 applications you could use
`var applications = new Faker<ApplicationDto>().Generate(3);`

Please consider refactoring.

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/MapSearchResultToOkOrNoContentExtensionTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        var searchResult = new SearchResult<ApplicationDto>() { TotalAmount = applications.Count, Entities = applications };
```
VadymLevkovskyi: (Test data) YAGNI again. `SearchResult.TotalAmount` is not used in real code, thus could be not filled in testing code. Please drop setting of this property.

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/MapSearchResultToOkOrNoContentExtensionTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        // Act
+        var result = controller.MapSearchResultToOkOrNoContent(searchResult);
```
VadymLevkovskyi: (Test mocks) As mentioned previously `controller` here is not the unit we are testing (not the system-under-test, or SUT), it's just a dependency for the real SUT - extension method. As all dependencies it could be effectively mocked using Mock library. Please rewrite

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/MapSearchResultToOkOrNoContentExtensionTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        // Assert
+        result.Should().NotBeNull();
+        result.Should()
+              .BeOfType<OkObjectResult>()
+              .Which.StatusCode
+              .Should()
+              .Be(StatusCodes.Status200OK);
```
VadymLevkovskyi, line 41: (Test assertions) Please don't test dependencies, test only SUT code.
Here you check that result is of `OkObjectResult` type. Is this type generation part of our code? No, it's generated in `controller.Ok()` that is not SUT and thus should be mocked. Instead, you should check that `Ok()` was called, but `NoContent()` - wasn't. See example
```
        // Arrange
        var applications = new Faker<ApplicationDto>().Generate(3);
        var searchResult = new SearchResult<ApplicationDto>() { Entities = applications };
        var expectedResult = new OkObjectResult(new { });

        var mockedController = new Mock<ControllerBase>();
        mockedController.Setup(c => c.Ok(searchResult)).Returns(expectedResult);

        // Act
        var result = mockedController.Object.MapSearchResultToOkOrNoContent(searchResult);

        // Assert
        result.Should().Be(expectedResult);

        mockedController.VerifyAll();
        mockedController.Verify(c => c.Ok(searchResult), Times.Once);
        mockedController.Verify(c => c.NoContent(), Times.Never);
```
Here we check that result of `Ok()` is passed without any transformation outside, `Ok()` is called once, `NoContent()` is never called, and (by using `.VerifyAll()`) that not any other controller method is called.

VadymLevkovskyi (commented):
Tests could be redesigned. Also there is one other old comment not addressed in your recent changes - please respond there or fix it.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
ShpurykGleb: Added.

I saw, that "httpContext.Setup(c => c.User.IsInRole("provider")).Returns(true);" is not needed in tests, they are wotking without it, is it a bug?

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/MapSearchResultToOkOrNoContentExtensionTests.cs
ShpurykGleb: I was thinking, that i can not create a mock of ControllerBase, because it is abstract (you can not create instance of  abstract class - you can not create a mock of it). My bad.

ShpurykGleb: Fixed.

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/MapSearchResultToOkOrNoContentExtensionTests.cs
ShpurykGleb: Fixed.

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/MapSearchResultToOkOrNoContentExtensionTests.cs
ShpurykGleb: Fixed with project generator.

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/MapSearchResultToOkOrNoContentExtensionTests.cs
ShpurykGleb: Fixed.

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/MapSearchResultToOkOrNoContentExtensionTests.cs
ShpurykGleb: Fixed.

OutOfSchool/OutOfSchool.WebApi.Tests/Extensions/MapSearchResultToOkOrNoContentExtensionTests.cs
ShpurykGleb: Thanks for critic. Fixed.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/ApplicationControllerTests.cs
VadymLevkovskyi: Not sure. I don't know curent security model. Please clarify it with @DmyMi .

VadymLevkovskyi (approved):

PR #1495, opened 2024-06-12
Refactored return construction in controllers.

ShpurykGleb:
Now, all controllers methods, that returns SearchResult, is refactored according to new extension method for controllers.


VadymLevkovskyi (commented):
Most of introduced changes are changin existing logic. `MapSearchResultToOkOrNoContent()` extension method doesn't check for TotalAmount field, while old code of many controllers was checking it. 

Please 
- either investigate TotalAmount field necessity and incase of no necessity drop it and add this explanation to the PR 
- or refactor your code.

ShpurykGleb:
> Most of introduced changes are changin existing logic. `MapSearchResultToOkOrNoContent()` extension method doesn't check for TotalAmount field, while old code of many controllers was checking it.
> 
> Please
> 
> * either investigate TotalAmount field necessity and incase of no necessity drop it and add this explanation to the PR
> * or refactor your code.

Can we add a new OR to SearchResult extension method for solving this?
I think, with this method we will cover all the possible cases, or this OR will violate the SRP?
image

VadymLevkovskyi:
> > ...
> 
> Can we add a new OR ... or this OR will violate the SRP? 

About SRP. It always depends on scope of that "responsibility". If you name that method appropriately (i.e. `IsNullOrEmpty`) - then it still might be the "one" responsibility.

But there is other and bigger issue with adding this OR.

Imagine front-end has asked some existing data using page number that is so big that there is no data. This case you'll get no entities, but TotalAmount will be > 0 (we still have some data for other page numbers). Your current code, as well as suggested addition of `|| .TotalAmount == 0` will return `NoDataResult` for such case. So front-end will think that there is no data at all, istead of knowing that there are no data for specific page number. (And I dunno how front-end works, but such behavior is pretty expected.)

That's why TotalAmount might be important and you should take `TotalAmount>0` into account (but in other way than suggested!) and that's why I've asked you in previous comment to investigate its usage.

Summing up. Before changing logic, please check if you understand all use-cases. If you are changing existing unit-tests - that means that you might be breaking the logic and introducing regression. Refactoring shouldn't break existing unit-tests!

ShpurykGleb:
> > > ...
> > 
> > 
> > Can we add a new OR ... or this OR will violate the SRP?
> 
> About SRP. It always depends on scope of that "responsibility". If you name that method appropriately (i.e. `IsNullOrEmpty`) - then it still might be the "one" responsibility.
> 
> But there is other and bigger issue with adding this OR.
> 
> Imagine front-end has asked some existing data using page number that is so big that there is no data. This case you'll get no entities, but TotalAmount will be > 0 (we still have some data for other page numbers). Your current code, as well as suggested addition of `|| .TotalAmount == 0` will return `NoDataResult` for such case. So front-end will think that there is no data at all, istead of knowing that there are no data for specific page number. (And I dunno how front-end works, but such behavior is pretty expected.)
> 
> That's why TotalAmount might be important and you should take `TotalAmount>0` into account (but in other way than suggested!) and that's why I've asked you in previous comment to investigate its usage.
> 
> Summing up. Before changing logic, please check if you understand all use-cases. If you are changing existing unit-tests - that means that you might be breaking the logic and introducing regression. Refactoring shouldn't break existing unit-tests!

Good evening, Vadym, thank's for critic.

Here is what i did after I saw your message:
- Added new extensions methods for this task + write tests for new methods
- Replace checking total amount with new extension method
- Rolled back my changes to tests (my bad, will know)

One thing, do not sure about naming to new extensions methods, is it good for you? I want to hear your opinion, if it is not a problem.

OutOfSchool/OutOfSchool.BusinessLogic/Extensions/SearchResultExtensions.cs
```diff
@@ -4,5 +4,10 @@ namespace OutOfSchool.BusinessLogic.Extensions;
 public static class SearchResultExtensions
 {
     public static bool IsNullOrEntitiesEmpty<T>(this SearchResult<T> searchResult)
-        => searchResult == null || searchResult.Entities.Count == 0;
+        => IsNull(searchResult) || searchResult.Entities is null || searchResult.Entities.Count == 0;
+
+    public static bool IsNullOrTotalAmountIsZero<T>(this SearchResult<T> searchResult)
+        => IsNull(searchResult) || searchResult.TotalAmount == 0;
+
+    private static bool IsNull<T>(SearchResult<T> searchResult) => searchResult == null;
```
VadymLevkovskyi: (Overengineering) Please avoid extracting so simple code into methods.

Reasons:
1. Method that does only 1 simple operation, doesn't decrease cognitive complexity.
2. This specific method is overcomplicated, it could be replaced by simple `bool IsNull(object o) => o is null;` 
- we don't need input argument specific type as we don't use it inside method;
- we don't need this method to be generic (and thus clutter the generated code with many method specifications) as generic type is not used.
3. Due to private nature - this method is never reused outside method, so doesn't help to "improve" whole project code style
4. Last but not least - there is a "rule of 3 duplicates". And here it is violated as well

OutOfSchool/OutOfSchool.BusinessLogic/Extensions/SearchResultExtensions.cs
```diff
@@ -4,5 +4,10 @@ namespace OutOfSchool.BusinessLogic.Extensions;
 public static class SearchResultExtensions
 {
     public static bool IsNullOrEntitiesEmpty<T>(this SearchResult<T> searchResult)
-        => searchResult == null || searchResult.Entities.Count == 0;
+        => IsNull(searchResult) || searchResult.Entities is null || searchResult.Entities.Count == 0;
+
+    public static bool IsNullOrTotalAmountIsZero<T>(this SearchResult<T> searchResult)
+        => IsNull(searchResult) || searchResult.TotalAmount == 0;
```
VadymLevkovskyi: (Logic) Please leave only one extension method checking for the `TotalAmount`. Subsequent changes in controller extensions should be done as well.

Explanations:

Providing 2 extension methods you (implicitly) specify that there could be such `SearchResult`, where `TotalAmount == 0` but `Entities` are not empty. 
I've checked all places where `SearchResult` is created and returned and it seems to be not the case.
Also, from semantic perspective, it should always `TotalAmount >= Entities.Count()`, otherwise the name `TotalAmount` is "totally" (pun intended) wrong.

Therefore there is no necessity to additionally check the count of Entities if you already checked `TotalAmount`.

From previos discussion, we know, that if `TotalAmount > 0 && Entities.Count() == 0` it's better to not return `NoContentResult`, because for paging case you loose important information that data (on other pages) exist.
That's why `IsNullOrEntitiesEmpty` shouldn't be ever used.

VadymLevkovskyi (changes requested):

OutOfSchool/OutOfSchool.BusinessLogic/Extensions/SearchResultExtensions.cs
ShpurykGleb: Good evening, Vadym. thanks for your thoughts.

I am investigating usage of TotalAmount at the moment, and are going to continue doing this task tomorrow.

And, also, i decided to attract Dmytryi thoughts for our conversation, because, i am not sure, that we (you and I) are going in right direction, to finish this task. 

I can say, that i am a little bit confused at the moment, about TotalAmount naming and using.

So, good night, I will write to you tomorrow, after i push the new changes in branch. (also i will add Dmytriy thoughts about it, to new comment)

ShpurykGleb: Hello, Vadym.

As I promised:

You was right, your thoughts and Dmytryi explanations are match.

Dmytryi explanation:
The front end sends a request asking to show X records from the array with some filter, where X can be any number from 1 to 100 (as we agreed that this is our maximum).
We return X records and indicate that there are a total of Total records in the database that match this filter, where Total can be anything such as Total > X, Total = X, or Total < X.
Then the front end calculates how many pages with X records need to be displayed to show the Total records.
When the user clicks on a new page, the front end calculates how many records need to be skipped, like page_number * X, so that the back end knows how many records to skip for the new selection and can provide data from the required page.
Total amount is the total number of records matching the WHERE clause. Entity count is the number of records we retrieved from the database using the "take" parameter in the query.

So, my changes are:
- now we have extension method IsNullOrEmpty for SearchResult
- now we have extension method SearchResultToOkOrNoContent in ControllerExtensions
- fixed some tests (added TotalAmount to test, because, when TotalAmount = null, tests throws NullReferenceException)
- also, we have a bug checking in SearchResultToOkOrNoContent , when TotalAmount can be less than Entities.Count (Dmytriy said, that that case is real, and happened minimum one time)

VadymLevkovskyi (approved):

Issue #1492, opened 2024-06-07
[Update] Add full and short english title to Provider


DmyMi:
Update to user stories #236 and #507

As a service provider/service provider's deputy I want to edit Provider's full and short title in english so that it can be searched by english speaking applicants.

### Acceptance Criteria
GIVEN user (service provider/service providers deputy) wants to edit provider's full and/or short title
AND opens provider edit page
WHEN scrolls to input "Повна назва англійською" and/or "Скорочена назва англійською"
THEN editable input field shall be displayed similar to the following
![image](https://github.com/ita-social-projects/OoS-Backend/assets/24808838/a54138c5-33f1-4650-a8f6-09107069a043)

### Business rules
- Both fields are **optional**, maxLength=60
- Both fields should be searchable using the text input field with full text search on the search page.

PR #1498, opened 2024-06-14
Add English full and short titles to Provider

MrVenik:
Related issue: #1492


VadymLevkovskyi (approved):
See no issues in current code. Also thanks for the `ExecuteUpdate` usage!

Though there are 2 things worth mentioning. 
1. (Not need to be changed now, just keep it in mind) In current situation, having only 2 languages, it might be added as extra parameter, but usually it's sign of not a good design.
2. Tests! At least you could ensure that `..En` is saved.

PR #1502, opened 2024-06-20
Ydzys/specify kind of date in changes logs to utc

YaroslavDZ:
Related issue #2543 (frontend)

https://github.com/ita-social-projects/OoS-Frontend/issues/2543

# Problem
For the endpoint api/v1/ChangesLog/ProviderAdmin, the OperationDate is returned without specifying the time zone, and therefore it is perceived by the system as local.

# What was changed
- Added additional assertions in unit tests.
- Specified the kind of DateTime in ChangesLogService.
- In ParentBlockedByAdminLogService the operation date property was set to UtcNow


VadymLevkovskyi (commented):
(Design) I'd suggest to cover all issues with Date Kind on presentation level (api/controllers/...). The business layer should work with right kind (`Utc` only, I guess) and never decide which dates to be converted and which should not.
Treat this as part of efforts for mapping data received from front-end - that mapping code shouldn't be propagated to BL level.

MrVenik (commented):
As @VadymLevkovskyi already mentioned (https://github.com/ita-social-projects/OoS-Backend/pull/1502#pullrequestreview-2130523779), the issue lies not only in mapping existing entities; rather, it lies in the creation logic. Therefore, I recommend updating the logic for creating a new `ChangesLog` entity in the repository. Currently, it relies on system time rather than UTC time

Code link:
https://github.com/ita-social-projects/OoS-Backend/blob/7fb0460d91b01efc0c45e1eda3fee34e7552645e/OutOfSchool/OutOfSchool.DataAccess/Repository/ChangesLogRepository.cs#L147-L165

HlibBondarev:
I agree with https://github.com/ita-social-projects/OoS-Backend/pull/1502#pullrequestreview-2130523779 and https://github.com/ita-social-projects/OoS-Backend/pull/1502#pullrequestreview-2139929041 -
"The business layer should work with right kind (Utc only) and never decide which dates to be converted and which should not".
So we have to set the "operation date" property to UtcNow instead of Now in 3 classes ([ProviderAdminChangesLogService.cs](https://github.com/ita-social-projects/OoS-Backend/pull/1502/files#diff-2115b151ea567ca0d2ee096c47f76a9ea2dcb8f60f878bd47f7fdb907ed09e17), [ParentBlockedByAdminLogService.cs](https://github.com/ita-social-projects/OoS-Backend/pull/1502/files#diff-ad7e80038fc539b27dea534b8dbccabd2bc0d9a4e08420eae009e1517aba242b), [ChangesLogRepository.cs](https://github.com/ita-social-projects/OoS-Backend/pull/1502/files#diff-fe8480f0d6f21b22617c72f90bb81a52675c37f5fb7d81e7c16695f7743570a5)).
Also, there is no need to change the DateTimeKind for property "operation date" in the [ChangesLogService.cs](https://github.com/ita-social-projects/OoS-Backend/pull/1502/files#diff-2e0ab8398a3eec1037c8b4b7b190ff1703a9db1e7fc172a0fa348adbe060df66). It has already had the type "DateTimeKind.Utc".

How the date and time for "operation date" property will be displayed must be determined on the front-end, based on the fact that the client always receives a date with DateTimeKind.Utc from the server.

HlibBondarev (approved):
We set the "operation date" property to UtcNow instead of Now in 3 classes ([ProviderAdminChangesLogService.cs](https://github.com/ita-social-projects/OoS-Backend/pull/1502/files#diff-2115b151ea567ca0d2ee096c47f76a9ea2dcb8f60f878bd47f7fdb907ed09e17), [ParentBlockedByAdminLogService.cs](https://github.com/ita-social-projects/OoS-Backend/pull/1502/files#diff-ad7e80038fc539b27dea534b8dbccabd2bc0d9a4e08420eae009e1517aba242b), [ChangesLogRepository.cs](https://github.com/ita-social-projects/OoS-Backend/pull/1502/files#diff-fe8480f0d6f21b22617c72f90bb81a52675c37f5fb7d81e7c16695f7743570a5)), but in MySql we store this information in the format - DATETIME, which does not contain the time-zone. Using another TIMESTAMP format to store the date, which stores the date with a time zone, has its own disadvantages: the range of this format is 1970-01-01 00:00:00 to 2038-01-19 03:14:17, in addition, there are other problems with this.
It is important for the front-end to have the time-zone, so we have to restore time-zone for property  "operation date" in [ProviderAdminChangesLogService.cs](https://github.com/ita-social-projects/OoS-Backend/pull/1502/files#diff-2115b151ea567ca0d2ee096c47f76a9ea2dcb8f60f878bd47f7fdb907ed09e17) and ALL OTHER classes which will contain datetime properties used by the front-end

h4wk13 (approved):

VadymLevkovskyi (approved):

PR #1516, opened 2024-07-23
Ratings showing fixed

ZdorovenkoKyrylo:


OutOfSchool/OutOfSchool.WebApi/Extensions/ElasticsearchExtension.cs
```diff
@@ -31,7 +31,8 @@ public static void AddElasticsearch(this IServiceCollection services, ElasticCon
         var settings = new ConnectionSettings(pool)
             .DefaultIndex(config.WorkshopIndexName)
             .BasicAuthentication(config.User, config.Password)
-            .EnableApiVersioningHeader();
+            .EnableApiVersioningHeader()
+            .ServerCertificateValidationCallback((a, b, c, d) => true);
```
VadymLevkovskyi: I wonder how this change is related to the ratings. Could you provide more details?

OutOfSchool/OutOfSchool.BusinessLogic/Services/AverageRatings/AverageRatingService.cs
```diff
@@ -43,6 +43,7 @@ public async Task<AverageRatingDto> GetByEntityIdAsync(Guid entityId)
         logger.LogInformation("Getting the average rating by workshop's or provider's id started.");
 
         var rating = (await averageRatingRepository.GetByFilter(r => r.EntityId == entityId).ConfigureAwait(false)).SingleOrDefault();
+        rating.Rate = (float)Math.Round(rating.Rate);
```
VadymLevkovskyi, line 46: Sorry, but when I see "showing" bug, I expect it to be fixed somewhere in presentation layer (not business logic) using something as simple as formatting. Could it be done there?

ZdorovenkoKyrylo: Thank you for the comment. As far as I understand this rounding must be working either in elastic or in database. So, I found that like the only way to implement that requirement. Maybe it will be reasonable to move somewhere this logic but I assume it won`t, because the presentation in both elastic and database seems pretty similar to changing logic to me.

OutOfSchool/OutOfSchool.WebApi/Extensions/ElasticsearchExtension.cs
ZdorovenkoKyrylo: I am sorry. I think that is left since I could not run elastic. I`ll remove it now

OutOfSchool/OutOfSchool.BusinessLogic/Services/AverageRatings/AverageRatingService.cs
ZdorovenkoKyrylo: @VadymLevkovskyi I also had some changes in mapping lines, you probably saw it when reviewed. That is how I tried to fix the problem with ratings. There were no visible ratings in elastic. Then the version of the elastic was changed and I do not know what was the reason but now ratings work even without those changes in mapping. that is why PR is still draft. And maybe you have some thoughts on this issue. I do not deny I might forget something while checking that is why I am still thinking about that.

OutOfSchool/OutOfSchool.BusinessLogic/Util/Mapping/ElasticProfile.cs
```diff
@@ -1,3 +1,4 @@
+using AutoMapper;
```
VadymLevkovskyi, line 1: Is this using needed?

OutOfSchool/OutOfSchool.BusinessLogic/Services/AverageRatings/AverageRatingService.cs
VadymLevkovskyi: Sorry, can't give any suggestion why it works/or doesn`t. 
Now I'm resolving this comment, and asking you to either undraft (proceed with fixing the code) or close the PR

PR #1541, opened 2024-08-04
Created Contact Information entity

### LINKED ISSUE #1525: [Data model] Contacts
Information that describes in a standardized way a set of information about means of communication and other contact information (address, telephones, websites, e-mails) that may belong to an organization, a separate division or an individual, or other business objects that may have a contact information. Can contain several contacts of the same type.

**Is NOT a bussiness entity**

Parent.Name | Name | Data Type | Mandatory | Comment
-- | -- | -- | -- | --
Contact information | address, type |`Address` #1542 | TRUE | 1..* can be several. Type - фактична/юридична. _Edited 12.08.2024_
Contact information | phone number, type | json | TRUE | 1..* can be several
 | | | | 
Contact information | facebook | http | FALSE | 0..* can be several. _Edited 12.08.2024_
Contact information | instagram | http | FALSE | 0..* can be several. _Edited 12.08.2024_
Contact information | website, type | json | FALSE | 0..* can be several
Contact information | email, type | json | FALSE | 0..* can be several

olsavchenk:


OutOfSchool/OutOfSchool.DataAccess/Models/PhoneNumber.cs
```diff
@@ -0,0 +1,19 @@
...
+    public long Id { get; set; }
+
+    [DataType(DataType.PhoneNumber)]
+    [Required(ErrorMessage = "Phone number is required")]
+    [CustomPhoneNumber(ErrorMessage = Constants.PhoneErrorMessage)]
+    [DisplayFormat(DataFormatString = Constants.PhoneNumberFormat)]
+    [MaxLength(Constants.MaxPhoneNumberLengthWithPlusSign)]
+    public string Number { get; set; }
+
+    public string Type { get; set; }
```
VadymLevkovskyi: Just curios: this `string Type` property - what it describes? Why `string` (not `enum`), why it is presetn for `Email` and `PhoneNumber`, but absent for `Website`

olsavchenk: The type property is something like a user note for Email, PhoneNumber, or Website (I missed adding it, but it's fixed now). For example, an entity can have many phone numbers associated with it, such as the owner's number, manager's number, etc.

VadymLevkovskyi: (Don't insist, just recommend) Then probably name it like `UserNotes` or `UserDescription`?

VadymLevkovskyi:
Please remove conmmented-out code

VadymLevkovskyi (approved):

olsavchenk:
Changes based on solution provided by Anatolii. The idea is to store optional contacts in serialised view. Here is how migration would look like:
```migrationBuilder.CreateTable(
    name: "Contacts",
    columns: table => new
    {
        Id = table.Column<Guid>(type: "binary(16)", nullable: false),
        UserNote = table.Column<string>(type: "longtext", nullable: true)
            .Annotation("MySql:CharSet", "utf8mb4"),
        OptionalContacts = table.Column<string>(type: "varchar", nullable: true)
            .Annotation("MySql:CharSet", "utf8mb4")
    },
    constraints: table =>
    {
        table.PrimaryKey("PK_Contacts", x => x.Id);
    })
    .Annotation("MySql:CharSet", "utf8mb4");

migrationBuilder.CreateTable(
    name: "PhoneNumbers",
    columns: table => new
    {
        Id = table.Column<Guid>(type: "binary(16)", nullable: false),
        Number = table.Column<string>(type: "varchar(16)", maxLength: 16, nullable: false)
            .Annotation("MySql:CharSet", "utf8mb4"),
        ContactId = table.Column<Guid>(type: "binary(16)", nullable: true)
    },
    constraints: table =>
    {
        table.PrimaryKey("PK_PhoneNumbers", x => x.Id);
        table.ForeignKey(
            name: "FK_PhoneNumbers_Contacts_ContactId",
            column: x => x.ContactId,
            principalTable: "Contacts",
            principalColumn: "Id");
    })
    .Annotation("MySql:CharSet", "utf8mb4");

migrationBuilder.CreateIndex(
    name: "IX_PhoneNumbers_ContactId",
    table: "PhoneNumbers",
    column: "ContactId");```

File: OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/ContactConfiguration.cs [hunk 2]
```diff
@@ -0,0 +1,20 @@
...
+
+internal class ContactConfiguration : IEntityTypeConfiguration<Contact>
+{
+    public void Configure(EntityTypeBuilder<Contact> builder)
+    {
+        builder.HasKey(x => x.Id);
+
+        builder.Property(e => e.OptionalContacts)
+                .HasConversion(
+                    v => JsonSerializer.Serialize(v, (JsonSerializerOptions)null),
```
h4wk13: Does `JsonSerializer` have Serialize/Deserialize methods without options ?
If no - you can use `JsonConvert` Serialize/Deserialize.

OutOfSchool/OutOfSchool.DataAccess/Models/Contact.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace OutOfSchool.Services.Models;
+
+public class Contact : IKeyedEntity<Guid>
```
h4wk13, line 6: xml-doc ?

OutOfSchool/OutOfSchool.DataAccess/Models/ContactInformation.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+
+namespace OutOfSchool.Services.Models;
+
+public class ContactInformation : IKeyedEntity<long>
+{
+    public long Id { get; set; }
+
+    public List<Address> Addresses { get; set; }
```
h4wk13, line 9: Is that possible that contact will have, hmmm, 1000 addresses ?

OutOfSchool/OutOfSchool.DataAccess/Models/Facebook.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.ComponentModel.DataAnnotations;
+using OutOfSchool.Common;
+
+namespace OutOfSchool.Services.Models;
+
+public class Facebook
```
h4wk13: Suggestion: As for me it is better to make all such contacts like `facebook`, `instagram` and other similar like a one `SocialNetworkContact` or `UrlContact` or better name, since they all will have an `Url`, and add a `enum` type for them so you will need a simple table to store it

OutOfSchool/OutOfSchool.DataAccess/Models/PhoneNumber.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+using OutOfSchool.Common;
+using OutOfSchool.Common.Validators;
+
+namespace OutOfSchool.Services.Models;
+
+public class PhoneNumber : IKeyedEntity<Guid>
```
h4wk13, line 8: why do you need that as a separate class ?

OutOfSchool/OutOfSchool.DataAccess/Models/Website.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.ComponentModel.DataAnnotations;
+using OutOfSchool.Common;
+
+namespace OutOfSchool.Services.Models;
+
+public class Website
```
h4wk13: please take a look at `facebook` contact comment

OutOfSchool/OutOfSchool.DataAccess/Models/Email.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace OutOfSchool.Services.Models;
+
+public class Email
+{
+    [DataType(DataType.EmailAddress)]
+    [MaxLength(256)]
```
h4wk13: can you move that `256` to `Constants` ?

OutOfSchool/OutOfSchool.DataAccess/Models/Instagram.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.ComponentModel.DataAnnotations;
+using OutOfSchool.Common;
+
+namespace OutOfSchool.Services.Models;
+
+public class Instagram
+{
+    [DataType(DataType.Url)]
+    [MaxLength(Constants.MaxUnifiedUrlLength)]
+    public string Url { get; set; } = string.Empty;
```
h4wk13: why it cannot be null ?

h4wk13 (changes requested):

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/ContactConfiguration.cs
olsavchenk: JsonSerializer has Serialize/Deserialize methods with null options, I use implicitly typed (JsonSerializerOptions)null to avoid the “Invocation tree...” error.

OutOfSchool/OutOfSchool.DataAccess/Models/ContactInformation.cs
olsavchenk: As far as I know we do not have any restictions regarding it. But I believe we can add them if needed :)

OutOfSchool/OutOfSchool.DataAccess/Models/Facebook.cs
olsavchenk: We don't store them directly in the database tables, they are just entities that we use to serialize/deserialize the data. I can add comments on this. Also, I'm not sure if it would be better to use a single entity as it would add code complexity and unnecessary enum checks throughout the code.

OutOfSchool/OutOfSchool.DataAccess/Models/Instagram.cs
olsavchenk: I added ```= string.Empty`` only to fix the null reference warning, and I can remove it if necessary

OutOfSchool/OutOfSchool.DataAccess/Models/PhoneNumber.cs
olsavchenk: Because we can have 1..* phoneNumbers linked to the single Contact

OutOfSchool/OutOfSchool.DataAccess/Models/Facebook.cs
olsavchenk: I actually misunderstood the question, sorry. If we use one class for all of this, we'll have to move field validation into the code, which can add complexity and lead to new bugs in the future

OutOfSchool/OutOfSchool.DataAccess/Models/Contact.cs
olsavchenk: I am not sure what comment should be added to this entity. Something like “An entity that contains the person's contact information”?

OutOfSchool/OutOfSchool.DataAccess/Models/Email.cs
olsavchenk: Updated

OutOfSchool/OutOfSchool.DataAccess/Models/Contact.cs
h4wk13: Yes, documentation matters.
If we don't add documentation for all the entities - the ok, you can skip it.

OutOfSchool/OutOfSchool.DataAccess/Models/ContactInformation.cs
h4wk13: Ok, do we have a search for whatever it is by address ? How we gonna do that in such case ?
Is that field required for contact ?
Where we validating that ?

OutOfSchool/OutOfSchool.DataAccess/Models/Facebook.cs
h4wk13: what validation ? ) contacts are optional, doesn't them ?
As far as I understood all you need to validate that this is valid `Uri`, meaning there is no check whether that contact exist.

OutOfSchool/OutOfSchool.DataAccess/Models/Instagram.cs
h4wk13: If we are using the `nullable` feature - we have to handle that in some way.
Like `string? Url { get; set; }`
By the way, why it is a `string` but not an `Uri` ?

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/ContactConfiguration.cs
h4wk13: Ok, can we fix it in any other way instead of this ?
if it is impossible - could you please add a comment that will explain why this code is here

olsavchenk: I changed the part to JsonConvert, so we don't need these “magic” options

OutOfSchool/OutOfSchool.DataAccess/Models/Contact.cs
olsavchenk: We do not add documentation to entities

OutOfSchool/OutOfSchool.DataAccess/Models/ContactInformation.cs
olsavchenk: The address is a mandatory part of the contact information. Hovewer I am not sure about the other 2 questions, I will ask them during the next daily status and provide you with an answer

OutOfSchool/OutOfSchool.DataAccess/Models/Facebook.cs
olsavchenk: I was referring to validating field attributes, but they are all the same so it doesn't matter. Hovewer I still don't think moving the OptionalContact's contacts into one class with an enum is a good idea as it will surely add complexity to the code, for example if we need to get the contact's Instagram it would look like this: contact.OptionalContact.Where(x => x.Type == ContactType.Instagram).ToList(). I can create something like OptionalContactEntityBase which will be base class for Instagram, Facebook and Website and will look like this:
```
public class OptionalContactEntityBase
{
    [DataType(DataType.Url)]
    [MaxLength(Constants.MaxUnifiedUrlLength)]
    public string Url { get; set; }
}
```
to reduce repeating properties but not sure if it is needed.
I'm not that experienced in code design, so if you insist on changing the OptionalContact contacts into a single object, I'll do it.

h4wk13: I'll rephrase my question: 
1.while we're gonna keep these contacts in single string it doesn't really matter how many types are there, but imagine you'll have to store them separately, how it will look like in the database ?
2. Imagine more complex case than filter just instagram contacts, for example instagram and facebook, how will you do that when all of them will be different types ?

olsavchenk: You're right, that looks better. I've updated the entities

h4wk13 (approved):

PR #1549, opened 2024-08-15
reorganize repositories

DmyMi:
- removed unit of work
- added save changes methods to base repository
- reorganised namespaces for easier navigation (up for discussion)

alot of files changes are due to new `using` imports


OutOfSchool/OutOfSchool.BusinessLogic/Services/ChildService.cs
```diff
@@ -539,7 +541,7 @@ private async Task CompleteChildChangesAsync()
     {
         try
         {
-            await childRepository.UnitOfWork.CompleteAsync();
+            await childRepository.SaveChangesAsync();
```
h4wk13, line 544: We have to make a choice whether to use or not `ConfigureAwait` bring it to a single form.
I suggest to use them everywhere where it is possible. But we can have a discussion about it.

OutOfSchool/OutOfSchool.DataAccess/Repository/Base/EntityRepositoryBase.cs
```diff
@@ -235,4 +234,16 @@ public virtual IQueryable<TEntity> Get(
...
     }
+
+    public Task<int> SaveChangesAsync(
+        bool acceptAllChangesOnSuccess = true,
+        CancellationToken cancellationToken = default)
+    {
+        return dbContext.SaveChangesAsync(acceptAllChangesOnSuccess, cancellationToken);
+    }
+
+    public int SaveChanges(bool acceptAllChangesOnSuccess = true)
```
h4wk13, line 245: why do we need `acceptAllChangesOnSuccess` here ?

h4wk13 (approved):

PR #1551, opened 2024-08-27
Hlib bondarev/store draft in cache

HlibBondarev:
**Roles:** provider, ministryadmin, areaadmin, regionadmin, techadmin.

**Story:** As a User with the above roles, I want to be able to save the field values before saving the Workshop (Provider, etc.) in the DB. So that if the  Workshop (Provider, etc.) creation is terminated, I can restore the values of previously filled fields.

**Solution**: Created DraftStorageService.  It captures and exports the state of an object (entity) outside of it so that the state of the original object can be restored later. Redis was used to store intermediate values of entity fields.

1) Created IDraftStorageService<T> interface and implemented it in StorageService<T> class .
2) Created IReadWriteCacheService interface with methods - ReadAsync(), WriteAsync() and RemoveAsync().
3) Implemented IReadWriteCacheService interface in CacheService class.
4) Refactored  CacheService class - used ReadAsync() and WriteAsync()  methods in GetOrAddAsync<T>() method.
5) Fixed RemoveAsync() method - instead of cache.RemoveAsync(key) used cache.Remove(key).
6) Created abstract generic controller DraftStorageController<T> with actions StoreDraft, RestoreDraft and RemoveDraft.
7) Created WorkshopDraftStorageController inherited from DraftStorageController<T> with T = WorkshopBaseDto.
8) Added tests to CacheServiceTests class for methods - ReadAsync(), WriteAsync() and RemoveAsync().
9) Created DraftStorageServiceTests class with NUnit tests for DraftStorageService class methods - RestoreAsync(), CreateAsync() and RemoveAsync().


OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/Storage.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    /// <summary>Gets the memento value asynchronous.</summary>
+    /// <param name="key">The key.</param>
+    /// <returns>
+    /// Representing the asynchronous operation with result KeyValuePair{string, string?} type.
+    /// </returns>
+    public async Task<KeyValuePair<string, string?>> GetMementoValueAsync(string key)
+    {
+        string logMsg = $"Getting memento with key = {key} from cache has started.";
+        logger.LogDebug(logMsg);
```
VadymLevkovskyi: (Logging too verbose due to wrong level) Not sure if this rather technical info should pe part of the "Debug" logs. It looks more like Information/Trace level

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/Storage.cs
```diff
@@ -0,0 +1,81 @@
...
+        var value = await crudCacheService.GetValueFromCacheAsync(key);
+
+        if (value == null)
+        {
+            logMsg = $"Memento with key = {key} not found in cache.";
+            logger.LogWarning(logMsg);
```
VadymLevkovskyi: (Log level) Is empty cache an issue or a normal case? I believe it's latter. Then why show warning? It's Debug/Info/Trace level, not higher

OutOfSchool/OutOfSchool.Redis/CacheService.cs
```diff
@@ -8,26 +8,25 @@
...
     private readonly ReaderWriterLockSlim cacheLock = new ReaderWriterLockSlim();
     private readonly IDistributedCache cache;
     private readonly RedisConfig redisConfig;
 
     private bool isWorking = true;
     private readonly bool isEnabled = false;
-
     private readonly object lockObject = new object();
-    
+    bool isExists = false;
```
VadymLevkovskyi: (State in srvices) Please avoid usage of any kind of state in services, as they should support shared access. 
The issue is next: serivce state (`isExists`) is shared for all service methods but is related to specific method input (`string key`). Thus I can in parallel call the `GetOrAddAsync()` twice or together with `GetValueFromCacheAsync()` (both are working with `isExists`), passing there different keys. What `key` this `isExists` will be related as a result, I dunno

OutOfSchool/OutOfSchool.Redis/CacheService.cs
```diff
@@ -46,73 +45,92 @@ public async Task<T> GetOrAddAsync<T>(
...
+
+        if (!isExists)
+        {
+            returnValue = await newValueFactory();
+            var value = JsonConvert.SerializeObject(returnValue);
+            await SetValueToCacheAsync(key, value, absoluteExpirationRelativeToNowInterval, slidingExpirationInterval);
+        }
+
+        return returnValue;
+    }
```
VadymLevkovskyi: Here you've broken previously existing logic of not adding same data twice.
Imagine `GetOrAddAsync` is called at same time for for same `key`, that is absent in cache.
You'll get `SetValueToCacheAsync` called twice in result. 
It's 
- either ok, then you don't need `cacheLock.EnterReadLock()` in `GetValueFromCacheAsync()`
- or wrong, then your code should be rewritten

OutOfSchool/OutOfSchool.Redis/CacheService.cs
```diff
@@ -46,73 +45,92 @@ public async Task<T> GetOrAddAsync<T>(
...
+                cache.SetString(key, value, options);
             }
             finally
             {
                 cacheLock.ExitWriteLock();
             }
         });
+    }
+
+    public Task RemoveFromCacheAsync(string key) => RemoveAsync(key);
```
VadymLevkovskyi: Please clarify what is the purpose of this method.
If it is just to provide other name for same logic as `RemoveAsync` do, than it's bit excessive, as for **Cache**Service.**Remove**Async(key) it's obvious to remove from cache, and you don't need specify it as **Cache**Service.Remove**FromCache**Async(key)

OutOfSchool/OutOfSchool.Redis/ICrudCacheService.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace OutOfSchool.Redis;
+
+public interface ICrudCacheService
+{
+    Task<string?> GetValueFromCacheAsync(string key);
```
VadymLevkovskyi: ( Naming again) ICrud**Cache**Service already has `Cache` in name, don't need to repeat it in all methods, like GetValue**FromCache**Async(), SetValue**ToCache**Async(), Remove**FromCache**Async(). 
It's basic property of encapsulation - to specify your main object of interest on class/interface name level and work with it on methods level.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopMementoControllerTests.cs
```diff
@@ -0,0 +1,204 @@
...
+        memento = FakeMemento();
+    }
+
+    //[Test]
+    //public async Task GetByFilter_WhenSearchResultIsNotNullOrEmpty_ReturnOkObjectResult()
+    //{
+    //    // Arrange
+    //    var data = new SearchResult<DirectionDto>()
+    //    {
+    //        Entities = new List<DirectionDto>() { new DirectionDto() },
```
VadymLevkovskyi: (Commented code) Please don't commit commented code. Usually it's useless later (due to subsequent code/requirement changes), thus could be removed in initial commit

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/Interfaces/IStorage.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace OutOfSchool.BusinessLogic.Services.Memento.Interfaces;
+
+public interface IStorage
```
VadymLevkovskyi: (Naming) Interface pretend to be a **Storage**. All methods are working with **Memento**. Looks bit irrelevant.
My recommendation - remove `Memento` part from all methods' names.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/Interfaces/IMementoService.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace OutOfSchool.BusinessLogic.Services.Memento.Interfaces;
+public interface IMementoService<T>
+{
+    T? State { get; set; }
+
+    IMemento CreateMemento(string key, T value);
```
VadymLevkovskyi: (Naming) Too many **Memento**s. Suggestion is to not duplicate information that is already set on class/interface name level (that it works with memento) in names of methods

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// <param name="memento">The memento.</param>
+    /// <exception cref="ArgumentNullException">Memento.</exception>
+    public MementoService(IMemento memento)
+    {
+        this.memento = memento ?? throw new ArgumentNullException(nameof(memento));
+    }
+
+    /// <summary>Gets or sets the State of the memento.</summary>
+    /// <value>The state of the memento.</value>
+    public T? State { get; set; }
```
VadymLevkovskyi: (Shared state in services) Issue similar to `isExists` mentioned in other comments. You could change it by parallel calling `RestoreMemento(memento)` with different input and get unpredictable result in `State`.

Suggestion - remove `State` propertty at all. Instead return the `State` data from `RestoreMemento()`.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// <summary>Restores the memento.</summary>
+    /// <param name="memento">The memento.</param>
+    public void RestoreMemento(KeyValuePair<string, string?> memento)
```
VadymLevkovskyi: (Probably wrong type) `KeyValuePair<string, string?> memento` - this inut type is very similar to `IMemento`. What is the reason to not use it?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
```diff
@@ -0,0 +1,61 @@
...
+        memento.State = new KeyValuePair<string, string?>(GetMementoKey(key), JsonConvert.SerializeObject(value));
+        return memento;
+    }
+
+    /// <summary>Gets the memento key.</summary>
+    /// <param name="key">The key.</param>
+    /// <returns>
+    /// Memento key of string type.
+    /// </returns>
+    public string GetMementoKey(string key)
```
VadymLevkovskyi: (Class design) I don't see reason to have this GetMementoKey() as a public method. In each call site you call it an then some other method form the same `MementoService`. Why don't make `GetMementoKey()` private one and call it inside `MementoService` public methods (like you do in `CreateMemento()`)?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/Storage.cs
```diff
@@ -0,0 +1,81 @@
+﻿using OutOfSchool.BusinessLogic.Services.Memento.Interfaces;
+
+namespace OutOfSchool.BusinessLogic.Services.Memento;
+
+/// <summary>
+/// Implements the IStorage interface to get, set, and remove memento in the cache.
+/// </summary>
+public class Storage : IStorage
```
VadymLevkovskyi: (Occam's razor) This whole `Storage` implementation looks like logging wrapper over `ICrudCacheService`. Why don't do all this things inside `ICrudCacheService` and don't introduce another entity with limited logic?

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/BaseMementoController.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// or
+    /// mementoService
+    /// or
+    /// storage.</exception>
+    protected BaseMementoController(
+        IMementoService<T> mementoService,
+        IStorage storage)
+    {
+        this.mementoService = mementoService ?? throw new ArgumentNullException(nameof(mementoService));
+        this.storage = storage ?? throw new ArgumentNullException(nameof(storage));
```
VadymLevkovskyi: (Logic) Controllers should usually avoid containing logic. In this controller you have 2 services and introduce logic of calling both of them in specific order. As for me it looks like they are highly coupled, that shouldn't be (remember, _high cohesion, low coupling_). 
Probably something is wrong on design level and these 2 services should be the one (or at least hidden by the facade of one additional service)?

VadymLevkovskyi (changes requested):
Naming, logic, design - please see other comments. All tests weren't reviewed, as they probably will be rewritten.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/Interfaces/IStorage.cs
VadymLevkovskyi: Also, see other comment in `BaseMementoController`. Looks like this `IStorage` could be joined with memento service.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/Storage.cs
HlibBondarev: IStorage and Storage has been removed from code.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/Storage.cs
HlibBondarev: IStorage and Storage has been removed from code.

OutOfSchool/OutOfSchool.Redis/CacheService.cs
HlibBondarev: isExists  field is removed from CacheService class.

OutOfSchool/OutOfSchool.Redis/CacheService.cs
HlibBondarev: I returned the logic that was before the changes in the code.

OutOfSchool/OutOfSchool.Redis/CacheService.cs
HlibBondarev: RemoveFromCacheAsync() is removed from CacheService class.
The RemoveAsync() method has been added to ICrudCacheService so that it can be used in MementoService where the ICrudCacheService interface is used.

OutOfSchool/OutOfSchool.Redis/ICrudCacheService.cs
HlibBondarev: Renamed methods in ICrudCacheService - removed the word 'Cache' from method names.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/Interfaces/IStorage.cs
HlibBondarev: IStorage and Storage has been removed from code.
Also IMemento and Memento has been removed from code.

OutOfSchool/OutOfSchool.WebApi.Tests/Controllers/WorkshopMementoControllerTests.cs
HlibBondarev: This code is removed

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/Interfaces/IMementoService.cs
HlibBondarev: Renamed methods in MementoService - removed the word 'memento' from method names.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
HlibBondarev: IMemento and Memento has been removed from code.
Method signature changed to Task<T> RestoreAsync(string key).

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
HlibBondarev: State property is removed from MementoService class.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/Storage.cs
HlibBondarev: It's done. 
Storage has been removed from code.
All CRUD logic of cache is relocated to ICrudCacheService.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/BaseMementoController.cs
HlibBondarev: It's done. 
Storage has been removed from code.
All CRUD logic of cache is relocated to ICrudCacheService.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
HlibBondarev: It's done. 
GetMementoKey() method has been made private.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        await crudCacheService.SetValueAsync(GetMementoKey(key), value);
+    }
+
+    /// <summary>Asynchronously removes a memento from the cache.</summary>
+    /// <param name="key">The key.</param>
+    /// <returns>Representation of an asynchronous operation - removing memento from the cache.</returns>
+    public async Task RemoveAsync(string key)
+    {
+        string logMessage;
```
VadymLevkovskyi: (Variable declaration placement) There are 2 suggestions I have about variable declarations:
1. Put them as close to usage as possible (not at the method beginning, at least)
2. Declare variable with immediate initialization - thus you could use 'var' instead of explicit type.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
```diff
@@ -0,0 +1,81 @@
...
+        string mementoKey = GetMementoKey(key);
+        var valueToRemove = await crudCacheService.GetValueAsync(mementoKey);
+
+        if (valueToRemove == null)
+        {
+            logMessage = $"Memento with key = {mementoKey} was not found in the cache.";
```
VadymLevkovskyi: What is the reason to not pass this string into `LogInformation()` directly without variable?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
```diff
@@ -0,0 +1,81 @@
...
+        string logMsg = $"Removing memento with key = {mementoKey} from cache has started.";
+        logger.LogInformation(logMsg);
+        await crudCacheService.RemoveAsync(mementoKey);
+        logMsg = $"Memento with key = {mementoKey} has been removed from the cache.";
+        logger.LogInformation(logMsg);
+    }
+
+    private static string GetMementoKey(string key)
+    {
+        return string.Format("{0}_{1}", key, typeof(T).Name);
```
VadymLevkovskyi: (Code style) Why not use string interpolation here? (That $"" syntax)

OutOfSchool/OutOfSchool.Redis/CacheService.cs
```diff
@@ -95,24 +93,71 @@ await ExecuteRedisMethod(() =>
...
+    }
+
+    public async Task SetValueAsync<T>(
+        string key,
+        T value,
+        TimeSpan? absoluteExpirationRelativeToNowInterval = null,
+        TimeSpan? slidingExpirationInterval = null)
+    {
+        await ExecuteRedisMethod(() =>
+        {
```
VadymLevkovskyi, line 121: (Code duplication) This method looks like a copy of part of the `GetOrAdd()` implementation. 
If it is so and to avoid duplication could it be called in `GetOrAdd()` instead of existing code?
If it is not a duplication - then why? For me it seems to do same thing

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopMementoController.cs
```diff
@@ -0,0 +1,20 @@
...
+
+/// <summary>Controller with CRUD operations for RequiredWorkshopMemento entity.</summary>
+[ApiController]
+[ApiVersion("1.0")]
+[Route("api/v{version:apiVersion}/[controller]/[action]")]
+public class WorkshopMementoController : BaseMementoController<RequiredWorkshopMemento>
+{
+    /// <summary>Initializes a new instance of the <see cref="WorkshopMementoController" /> class.</summary>
+    /// <param name="mementoService">The memento service.</param>
+    public WorkshopMementoController(
```
VadymLevkovskyi: (Code style) Primary constructor could be used here

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /// <summary>Asynchronously removes a memento from the cache.</summary>
+    /// <param name="key">The key.</param>
+    /// <returns>Representation of an asynchronous operation - removing memento from the cache.</returns>
+    public async Task RemoveAsync(string key)
+    {
+        string logMessage;
+        string mementoKey = GetMementoKey(key);
+        var valueToRemove = await crudCacheService.GetValueAsync(mementoKey);
```
VadymLevkovskyi: (Code style) Please be consistent about `var` usage. Either don't use it at all or use it everywhere. Here in 2 lines of code variables are defined in 2 different ways

VadymLevkovskyi (commented):
Code style comments

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
HlibBondarev: It's done.
Declared a variable with immediate initialization in the RemoveAsync() method of the MementoService<T> class.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
HlibBondarev: It's done.
Replaced String.Format with interpolation in the GetMementoKey() method of the MementoService<T> class.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopMementoController.cs
HlibBondarev: It's done.
Used the primary constructor in the WorkshopMementoController.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
HlibBondarev: It's done.
Replaced the variable type 'string' with the keyword 'var'.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
HlibBondarev: I passed the interpolated string to LogInformation() directly, without a variable. But Sonar wrote this warning - Don't use string interpolation in logging message templates.

https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&sinceLeakPeriod=true&pullRequest=1551&id=ita-social-projects-oos-backend

OutOfSchool/OutOfSchool.Redis/CacheService.cs
HlibBondarev: Today, I've had a discussion about MementoService and CacheService with @h4wk13.
He suggested not to create a new entity (RequiredWorkshopMemento) and then serialize it and pass it to Cache. Instead, work directly with the Json-string and store it into the cache. Then all work will be done in CacheService class.
I asked this question to @DmyMi and now I'm wating the answer.

**UPD**
I had a talk with  @DmyMi on Friday.
He explained how the code should work, and why you need to pass a DTO entity to the Controller Action, not just a string. I want to discuss this with @h4wk13 on Monday

HlibBondarev: > (Code duplication) This method looks like a copy of part of the `GetOrAdd()` implementation. If it is so and to avoid duplication could it be called in `GetOrAdd()` instead of existing code? If it is not a duplication - then why? For me it seems to do same thing

I agree with you that the code in GetOrAddAsync<T>() method is similar to the code in new methods.
In my opinion, this happened because this method should not be related to the ICacheService interface, since it has a query to the database. 
I think that a new interface should be created (I don’t know what to call it yet) in which the GetOrAddAsync<T>() method will be included. And all other methods (RemoveAsync(), GetValueAsync() and UpsertValueAsync<T>()) have to be in ICasheService interface. Then ICrudCacheService does not make sense and should be removed.

VadymLevkovskyi: Sorry, where we have query to db in this code?

HlibBondarev: parameter - Func<Task<T>> newValueFactory,
if (!isExists)
{
    returnValue = await newValueFactory();
...

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/Interfaces/IMementoService.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace OutOfSchool.BusinessLogic.Services.Memento.Interfaces;
+public interface IMementoService<T>
```
h4wk13: Why do we want to state that this is a Memento ?
Is that business logic related ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
```diff
@@ -0,0 +1,82 @@
...
+        this.crudCacheService = crudCacheService;
+        this.logger = logger;
+    }
+
+    /// <summary>Restores the memento.</summary>
+    /// <param name="key">The key.</param>
+    /// <returns> Representing the asynchronous operation with result of T type.</returns>
+    public async Task<T> RestoreAsync([NotNull] string key)
+    {
+        var mementoKey = key is not null ? GetMementoKey(key) : throw new ArgumentNullException(nameof(key));
```
h4wk13: 1. please move arguments validation logic out from the business logic.
2. Please use `ArgumentException.ThrowIfNullOrEmpty(nameof(key))` instead.

It will bring more readability to the code.
Update all other places.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        var memento = await crudCacheService.GetValueAsync(mementoKey).ConfigureAwait(false);
```
h4wk13: What exactly `memento` means here ?
We're storing instance of the memento pattern in the storage ? Why ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
```diff
@@ -0,0 +1,82 @@
...
+
+namespace OutOfSchool.BusinessLogic.Services.Memento;
+
+/// <summary>
+/// Implements the IMementoService{T} interface with CRUD functionality to store an entity of type T in a cache.
+/// </summary>
+/// <typeparam name="T">T is the entity type that should be stored in the cache.</typeparam>
+public class MementoService<T> : IMementoService<T>
+{
+    private readonly ICrudCacheService crudCacheService;
```
h4wk13: `Crud` ? in name ? why ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
```diff
@@ -0,0 +1,82 @@
...
+        var valueToRemove = await crudCacheService.GetValueAsync(mementoKey);
+
+        if (valueToRemove == null)
+        {
+            logger.LogInformation($"Memento with key = {mementoKey} was not found in the cache.");
+            return;
+        }
+
+        logger.LogInformation($"Removing memento with key = {mementoKey} from cache has started.");
+        await crudCacheService.RemoveAsync(mementoKey).ConfigureAwait(false);
```
h4wk13: is it possible that value will be removed from the cache between that `if and this line ?

OutOfSchool/OutOfSchool.Redis/CacheService.cs
```diff
@@ -95,24 +95,71 @@ await ExecuteRedisMethod(() =>
     }
 
     public Task RemoveAsync(string key)
-        => ExecuteRedisMethod(async () => {
+        => this.ExecuteRedisMethod(() =>
```
h4wk13: Could you please look inside this method and check if we can stuck in the internal `while` for a wile ?
Also please, invert the `if` inside it,

OutOfSchool/OutOfSchool.Redis/CacheService.cs
```diff
@@ -95,24 +95,71 @@ await ExecuteRedisMethod(() =>
...
             {
                 cacheLock.ExitWriteLock();
             }
         });
 
+    public async Task<string> GetValueAsync(string key)
+    {
+        string returnValue = string.Empty;
+
+        await ExecuteRedisMethod(() =>
```
h4wk13, line 99: I see that Enter/Exit... in many places.
Is that possible to create 2 methods to wrap that logic ?
One for read and one for write locks ?

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/BaseMementoController.cs
```diff
@@ -0,0 +1,61 @@
+﻿using Microsoft.AspNetCore.Mvc;
+using OutOfSchool.BusinessLogic.Common;
+using OutOfSchool.BusinessLogic.Services.Memento.Interfaces;
+
+namespace OutOfSchool.WebApi.Controllers.V1;
+
+/// <summary>Base Controller with operations for storing data in cache.</summary>
+/// <typeparam name="T">T is the entity type that should be stored in the cache.</typeparam>
+public abstract class BaseMementoController<T> : ControllerBase
```
h4wk13: Please rename it!

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopMementoController.cs
```diff
@@ -0,0 +1,14 @@
...
+using OutOfSchool.BusinessLogic.Models.Workshops.IncompletedWorkshops;
+using OutOfSchool.BusinessLogic.Services.Memento.Interfaces;
+
+namespace OutOfSchool.WebApi.Controllers.V1;
+
+/// <summary>Controller with CRUD operations for WorkshopWithRequiredPropertiesDto entity.</summary>
+[ApiController]
+[ApiVersion("1.0")]
+[Route("api/v{version:apiVersion}/[controller]/[action]")]
+public class WorkshopMementoController(IMementoService<WorkshopWithRequiredPropertiesDto> mementoService)
```
h4wk13: Are you sure it should contain a `memento` as a part of the name ?

h4wk13 (changes requested):

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/Interfaces/IMementoService.cs
HlibBondarev: Yes, I think that we store in Redis not data but some notes (memento)

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
HlibBondarev: > 1. please move arguments validation logic out from the business logic.
> 2. Please use `ArgumentException.ThrowIfNullOrEmpty(nameof(key))` instead.
> 
> It will bring more readability to the code. Update all other places.

1) When we discussed last time this code I understood that I had to check all arguments that input to service methods.

2) Ok, I changed code

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
HlibBondarev: I think that we store in Redis not data but some notes (memento)

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
HlibBondarev: There are all CRUD operations in ICrudCacheService - Create and Update (UpsertValueAsync), Read (GetValueAsync) and Delete (RemoveAsync).


Add my discussion with [VadymLevkovskyi](https://github.com/VadymLevkovskyi) about Code duplication in CasheService class:

> > (Code duplication) This method looks like a copy of part of the `GetOrAdd()` implementation. If it is so and to avoid duplication could it be called in `GetOrAdd()` instead of existing code? If it is not a duplication - then why? For me it seems to do same thing
> 
> I agree with you that the code in GetOrAddAsync() method is similar to the code in new methods. In my opinion, this happened because this method should not be related to the ICacheService interface, since it has a query to the database. I think that a new interface should be created (I don’t know what to call it yet) in which the GetOrAddAsync() method will be included. And all other methods (RemoveAsync(), GetValueAsync() and UpsertValueAsync()) have to be in ICasheService interface. Then ICrudCacheService does not make sense and should be removed.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/MementoService.cs
HlibBondarev: Thanks, I agree.
I deleted code line - 
logger.LogInformation($"Memento with key = {mementoKey} has been removed from the cache.");

OutOfSchool/OutOfSchool.Redis/CacheService.cs
HlibBondarev: I can't find _if_ in this method

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/BaseMementoController.cs
HlibBondarev: Is the name MementoController<T> appropriated?

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopMementoController.cs
HlibBondarev: I think controller name should contain **memento** in it, because in API may contain another controller with **memento**, for example - ProviderMementoController

OutOfSchool/OutOfSchool.Redis/CacheService.cs
VadymLevkovskyi: I get it finally. It's not a "query to db" its just a "value factory". And I see the difference now.
But I still don't see why it's impossible to reduce code duplication. You can use `Upsert` inside `GetOrAdd`.

OutOfSchool/OutOfSchool.Redis/CacheService.cs
vpustovoitov: Is locking necessary within this service? Since `CacheService` is either a **Scoped** or **Transient** service, concurrency shouldn’t be an issue.

In my opinion, for these requirements, we could create a service that isn’t coupled with the database, acting as a temporary data storage solution.

Alternatively, we could create a **Facade** for interacting with **Redis** and a **Decorator** for caching database responses. This way, we could use the **Facade** service to store temporary data.

UPD: Oh, as I see, `CacheService` is injected as a **Singleton** service (but why ?).

OutOfSchool/OutOfSchool.Redis/CacheService.cs
HlibBondarev: Ok, I used ReadAsync() and WriteAsync inside GetOrAdd.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/JSONConverter/JSONCreationConverter.cs
```diff
@@ -0,0 +1,44 @@
+﻿using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace OutOfSchool.BusinessLogic.Services.Memento.JSONConverter;
+
+public abstract class JsonCreationConverter<T> : JsonConverter
+{
+    public override bool CanWrite
```
olsavchenk: (Code style) I think it will be more readable if you make it
`public override bool CanWrite { get => false; }`

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/JSONConverter/JSONCreationConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        JObject jObject = JObject.Load(reader);
+        T target = Create(objectType, jObject);
+        serializer.Populate(jObject.CreateReader(), target);
+
+        return target;
+    }
+
+    public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
```
olsavchenk: (Code style) Isn't it better to just not add this method if you don't implement it?

OutOfSchool/OutOfSchool.Redis/IReadWriteCacheService.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace OutOfSchool.Redis;
+
+public interface IReadWriteCacheService
+{
+    Task<string> ReadAsync(string key);
+
+    public Task WriteAsync(
```
olsavchenk: (Code style) Please remove “public” as it is public anyway

OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/IncompletedWorkshops/WorkshopWithContactsDto.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.ComponentModel.DataAnnotations;
+using Microsoft.AspNetCore.Mvc;
+using OutOfSchool.BusinessLogic.Util.JsonTools;
+
+namespace OutOfSchool.BusinessLogic.Models.Workshops.IncompletedWorkshops;
+public class WorkshopWithContactsDto : WorkshopWithDescriptionDto
```
olsavchenk: (Code style) Please add a blank line between the namespace and the class. And in other places too (see WorkshopWithDescriptionDto, WorkshopWithTeachersDto etc.)

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/JSONConverter/JSONCreationConverter.cs
HlibBondarev: O.k. I changed code - used an exspression body for this method and CanConvert().

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/JSONConverter/JSONCreationConverter.cs
HlibBondarev: WriteJson() is an abstract method of JsonConverter class, so I have to implemented it.

OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/IncompletedWorkshops/WorkshopWithContactsDto.cs
HlibBondarev: Ok, I've done it

OutOfSchool/OutOfSchool.Redis/IReadWriteCacheService.cs
HlibBondarev: Ok, I've done it

OutOfSchool/OutOfSchool.BusinessLogic/Services/Memento/JSONConverter/JSONCreationConverter.cs
olsavchenk: Yes, you must implement abstract methods when inheriting abstract classes, but JsonCreationConverter is also abstract, so there is no need to override it (especially with a simple throwing error). JsonCreationConverter simply extends the JsonConverter's abstaction(if I may put it that way)

HlibBondarev: If I don't implement WriteJson() method in JSONCreationConverter class, then I have to implement it in WorkshopConverter class, which is not very good in my opinion. The only task of WorkshopConverter class is to implement Create() method.

OutOfSchool/OutOfSchool.Redis/CacheService.cs
VadymLevkovskyi: Ok, we have `DraftStorageService` and `IDistributedCache` and `CacheService`. "Not that we needed all that for the trip" ([Fear and Loathing in Las Vegas](https://www.goodreads.com/quotes/59592-we-had-two-bags-of-grass-seventy-five-pellets-of-mescaline)) but obviously we could add another Facade

OutOfSchool/OutOfSchool.WebApi.Tests/Redis/CacheServiceTests.cs
```diff
@@ -85,14 +86,71 @@ public async Task GetOrAddAsync_WhenDataNotExistsOrExpired_ShouldSaveNewData()
...
+
+    [Test]
+    public async Task ReadAsync_WhenDataExistsInCacheAndNotExpired_ShouldReturnData()
+    {
+        // Arrange
+        var expected = new Dictionary<string, string>()
+        {
+            {"ExpectedKey", "ExpectedValue"},
+        };
+        distributedCacheMock.Setup(c => c.Get(It.IsAny<string>()))
```
VadymLevkovskyi: (Testing wrong data) Please avoid usage of `It.IsAny<>()` whenever possible. Here you know the key thus could specify it in `Setup`. Please do the same in other tests as well

For this specific test it just adds better understanding of the logic. But for some tests it might be crucial.

OutOfSchool/OutOfSchool.WebApi.Tests/Redis/CacheServiceTests.cs
```diff
@@ -85,14 +86,71 @@ public async Task GetOrAddAsync_WhenDataNotExistsOrExpired_ShouldSaveNewData()
...
+    }
+
+    [Test]
+    public async Task ReadAsync_WhenDataExistsInCacheAndNotExpired_ShouldReturnData()
+    {
+        // Arrange
+        var expected = new Dictionary<string, string>()
+        {
+            {"ExpectedKey", "ExpectedValue"},
+        };
```
VadymLevkovskyi: (Complicated test) Please pay attention what you are testing.

For this specific test (and few next) - sorry, but `expected` variable:
1. It is not "expected" (expected is serialized presentation of this dictionary)
2. It has wrong type - as `ReadAsync` returns just string without extra processing - why not check strings without that serialization / Contains magic?
3. It is hardcoded - why not use Bogus random string?

OutOfSchool/OutOfSchool.WebApi.Tests/Redis/CacheServiceTests.cs
```diff
@@ -85,14 +86,71 @@ public async Task GetOrAddAsync_WhenDataNotExistsOrExpired_ShouldSaveNewData()
...
+            Times.Once);
+    }
+
+    [Test]
+    public async Task ReadAsync_WhenDataExistsInCacheAndNotExpired_ShouldReturnData()
+    {
+        // Arrange
+        var expected = new Dictionary<string, string>()
+        {
+            {"ExpectedKey", "ExpectedValue"},
```
VadymLevkovskyi: (Magic strings) `"ExpectedKey"` is used twice. Please think about using const. Same for next tests

OutOfSchool/OutOfSchool.WebApi.Tests/Redis/CacheServiceTests.cs
```diff
@@ -85,14 +86,71 @@ public async Task GetOrAddAsync_WhenDataNotExistsOrExpired_ShouldSaveNewData()
...
+        };
+        distributedCacheMock.Setup(c => c.Get(It.IsAny<string>()))
+            .Returns(Encoding.UTF8.GetBytes(JsonConvert.SerializeObject(expected)));
+
+        // Act
+        var result = await readWriteCacheService.ReadAsync("ExpectedKey");
+
+        // Assert
+        result.Should().Contain("ExpectedValue");
+        distributedCacheMock.Verify(
```
VadymLevkovskyi: (Mock calls verification) In general, please prefer `VerifyAll` over specific `Verify(..., Times.Once)`. Former lets you check that absence of excessive `Setup`s

OutOfSchool/OutOfSchool.WebApi.Tests/Services/DraftStorage/DraftStorageServiceTests.cs
```diff
@@ -0,0 +1,157 @@
...
+        loggerMock = new Mock<ILogger<DraftStorageService<WorkshopBaseDto>>>();
+        readWriteCacheServiceMock = new Mock<IReadWriteCacheService>();
+        draftStorageService = new DraftStorageService<WorkshopBaseDto>(readWriteCacheServiceMock.Object, loggerMock.Object);
+    }
+
+    [Test]
+    public async Task RestoreAsync_WhenDraftExistsInCache_ShouldRestoreAppropriatedEntity()
+    {
+        // Arrange
+        var workshopDraft = new WorkshopBaseDto()
```
VadymLevkovskyi: (Fake data) Test data could and should be randomly generated. See other tests using Faker.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/DraftStorage/DraftStorageServiceTests.cs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            Title = "title",
+            Email = "myemail@gmail.com",
+            Phone = "+380670000000",
+        };
+        var expected = new Dictionary<string, string>()
+            {
+                {"ExpectedKey", "{\"Title\":\"title\",\"Email\":\"myemail@gmail.com\",\"Phone\":\"+380670000000\"}"},
```
VadymLevkovskyi: Please cosider refactoring (`JsonConvert.SerializeObject` and usage of `const`)

There are two issues:
1. Manual json serialization. Wish a good luck for next developer trying to change `workshopDraft`
2. Magic string - `"ExpectedKey"` - used here and in `Setup` and in `RestoreAsync`. Another "good luck wishes" for changing this test.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/DraftStorage/DraftStorageServiceTests.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        // Act
+        var result = await draftStorageService.RestoreAsync("ExpectedKey");
+
+        // Assert
+        readWriteCacheServiceMock.Verify(
+            c => c.ReadAsync(It.IsAny<string>()),
+            Times.Once);
+        result.Should().BeOfType<WorkshopBaseDto>();
+        Assert.AreEqual(workshopDraft.Title, result.Title);
```
VadymLevkovskyi: (Code style) `Should()` and `Assert()` in one test. Consider usage of `Should....Which` or `Should...Where` or `BeEquivalentTo`

OutOfSchool/OutOfSchool.Redis/CacheService.cs
HlibBondarev: > `IDistributedCache`

To implement the DraftStorageService<T> logic, I only need two methods - ReadAsync() and WriteAsync(). I could create this class as a facade over IDistributedCache. But in this case, in order for correctly reading and writing to the cache, I will need to implement the locking logic, which is already implemented in the methods of the CacheService class.
Therefore, I added the IReadWriteCacheService interface and implemented its methods in CacheService.
DraftStorageService<T> class is designed  as a facade over CacheService. This class works with entities of type T, unlike CacheService, which saves and restores information in string format.

OutOfSchool/OutOfSchool.WebApi.Tests/Redis/CacheServiceTests.cs
HlibBondarev: Ok, done. I replaced It.IsAny<> with specific known values.

OutOfSchool/OutOfSchool.WebApi.Tests/Redis/CacheServiceTests.cs
HlibBondarev: Ok, done.
1) Used Bogus random string for generating data.
2) Removed dictionary.

OutOfSchool/OutOfSchool.WebApi.Tests/Redis/CacheServiceTests.cs
HlibBondarev: Ok, done.

Used Bogus random string for generating data.
Removed dictionary.

OutOfSchool/OutOfSchool.WebApi.Tests/Redis/CacheServiceTests.cs
HlibBondarev: Ok, done.
Used VerifyAll() method for tests instead of Verify().

OutOfSchool/OutOfSchool.WebApi.Tests/Services/DraftStorage/DraftStorageServiceTests.cs
HlibBondarev: Ok, done.
Used Bogus.Facker for creating Workshop draft with random properties.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/DraftStorage/DraftStorageServiceTests.cs
HlibBondarev: Ok, done.
Used Bogus random string for generating key and cacheKey instead hardcore string.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/DraftStorage/DraftStorageServiceTests.cs
HlibBondarev: Ok, done. 
Use Should() in all places instead of Assert().

PR #1552, opened 2024-08-28
Nakoneshniuk/add moderator role

### LINKED ISSUE #1548: Add moderator role and admin/moderator workshops endpoint
- [ ] Add new set of permissions for approving workshops in [permissions list](https://github.com/ita-social-projects/OoS-Backend/blob/develop/OutOfSchool/OutOfSchool.Common/PermissionModule/PermissionManagement/Permissions.cs). Check provider permissions for reference. Also have a separate permission for moderators to read providers and workshops.
- [ ] Add a new role named `moderator` using [hosted service](https://github.com/ita-social-projects/OoS-Backend/blob/develop/OutOfSchool/OutOfSchool.AuthCommon/Services/IdentityRolesInitializerHostedService.cs#L8)
- [ ] Add this permission to every role that already was able to [approve providers](https://github.com/ita-social-projects/OoS-Backend/blob/develop/OutOfSchool/OutOfSchool.Common/PermissionModule/PermissionManagement/PermissionsSeeder.cs). Also create a new permissions pack for `moderator` role to be able to read his info and have access to approve provider/workshop. Make sure that a new migration with packed permissions exists.
- [ ] In `AdminController` create a new endpoint `GetWorkshopsByFilter` that is similar to [GetProviderByFilter](GetProviderByFilter) and returns a full `WokrshopDTO` list. Make this endpoint accessible only to moderator workshop list permission. 

> Note: You may need to modify other components aswell.

IrynaNakoneshniuk:


OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceTests.cs
```diff
@@ -0,0 +1,224 @@
...
+
+    private void SetUpWorkshopsRepository(List<Workshop> workshopsReturned)
+    {
+        workshopRepository.Setup(
+                w => w.Get(
+                    It.IsAny<int>(),
+                    It.IsAny<int>(),
+                    It.IsAny<string>(),
+                    It.IsAny<Expression<Func<Workshop, bool>>>(),
+                    It.IsAny<Dictionary<Expression<Func<Workshop, object>>, SortDirection>>(),
```
VadymLevkovskyi: (Testing wrong logic) Please avoid wherever possible usage of `It.IsAny()`, as it doesn't ensure that input parameters passed as expected. I.e. there're two first arguments of `int` type and they can be switched by mistake in real code, but your test will never catch this case.

IrynaNakoneshniuk: Thank you, i will fix this.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -631,6 +648,121 @@ public async Task<SearchResult<WorkshopCard>> GetNearestByFilter(WorkshopFilter
         return result;
     }
 
+    /// <inheritdoc/>
```
h4wk13, line 651: please add closing part after the docs

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -631,6 +648,121 @@ public async Task<SearchResult<WorkshopCard>> GetNearestByFilter(WorkshopFilter
...
+    /// <exception cref="InvalidOperationException">If the region admin is not found in the database.</exception>
+    public async Task<SearchResult<WorkshopDto>> FetchByFilterForAdmins(WorkshopFilterAdministration filter = null)
+    {
+        logger.LogInformation("Started retrieving Workshops by filter for admins.");
+        filter = filter ?? new WorkshopFilterAdministration();
```
h4wk13: could you please add a `Debug` log that value is null and we continue with the default (WorkshopFilterAdministration) filter

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -631,6 +648,121 @@ public async Task<SearchResult<WorkshopCard>> GetNearestByFilter(WorkshopFilter
...
+                includeProperties: includingPropertiesForMappingDtoModel,
+                whereExpression: predicate,
+                asNoTracking: true)
+            .ToListAsync()
+            .ConfigureAwait(false);
+
+        var workshopsDTO = mapper.Map<List<WorkshopDto>>(workshops);
+        var workshopsCount = workshops.Count;
+
+        logger.LogInformation("Retrieved {WorkshopsCount} matching recods by filter for admins.", workshopsCount);
```
h4wk13, line 674: In case when mapper fail - this log will be lost. I'd suggest to move it to the 668 line

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -631,6 +648,121 @@ public async Task<SearchResult<WorkshopCard>> GetNearestByFilter(WorkshopFilter
...
+    }
+
+    private async Task<Expression<Func<Workshop, bool>>> PredicateBuildForAdminds(WorkshopFilterAdministration filter)
+    {
+        var predicate = PredicateBuilder.True<Workshop>();
+
+        // TODO: add condition for united territorial community admin
+        if (currentUserService.IsMinistryAdmin())
+        {
+           await ApplyFilterForMinistryAdmin(filter);
```
h4wk13: You've added `ConfigureAwait` in the previous method, why didn't here ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -631,6 +648,121 @@ public async Task<SearchResult<WorkshopCard>> GetNearestByFilter(WorkshopFilter
...
+        {
+            var tempPredicate = await ApplyFilterForRegionAdmin(filter);
+            predicate = predicate.And(tempPredicate);
+        }
+
+        if (!string.IsNullOrEmpty(filter.SearchString))
+        {
+            var tempPredicate = PredicateBuilder.False<Workshop>();
+
+            foreach (var word in filter.SearchString.Split(' ', ',', StringSplitOptions.RemoveEmptyEntries))
```
h4wk13: I suppose these splitters should not be here.
It looks like a part of a 'Search' business logic.
Why dot is not here ? Are there other word separators like `-` or `;` ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -631,6 +648,121 @@ public async Task<SearchResult<WorkshopCard>> GetNearestByFilter(WorkshopFilter
...
+            {
+                tempPredicate = tempPredicate.Or(x => EF.Functions.Like(x.Keywords, $"%{word}%"));
```
h4wk13: I suppose answer is yes but - does this `EF.Functions.Like` protects itself from the injections ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -631,6 +648,121 @@ public async Task<SearchResult<WorkshopCard>> GetNearestByFilter(WorkshopFilter
...
+
+        if (filter.InstitutionId != Guid.Empty)
+        {
+            predicate = predicate.And(x => x.InstitutionHierarchy.InstitutionId == filter.InstitutionId);
+        }
+
+        if (filter.CATOTTGId > 0)
+        {
+            var subSettlementsIds = await codeficatorService
+            .GetAllChildrenIdsByParentIdAsync(filter.CATOTTGId).ConfigureAwait(false);
```
h4wk13: please format: settle `ConfigureAwait` to it's own line

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
IrynaNakoneshniuk: All EF methods are converted to command paramerers and as such should be safe from SQL injection. But in our case all queries built using the ```LIKE ``` operator are not parameterized.  Perhabs, issues with the MySql provider?) I am looking for a solution.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
IrynaNakoneshniuk: I built the filtering based on the analogy with the filtering logic of providers , where a search string is also used. Thank you, I indeed didn't consider other separators, I will fix it.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
IrynaNakoneshniuk: fixed

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
IrynaNakoneshniuk: fixed

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
IrynaNakoneshniuk: fixed

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
IrynaNakoneshniuk: added  ```ConfigureAwait```

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
IrynaNakoneshniuk: I will create a  service for handling the search string in a separate pr.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
IrynaNakoneshniuk: I changed  function to  `Contains`.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
IrynaNakoneshniuk: fixed

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -631,6 +648,132 @@ public async Task<SearchResult<WorkshopCard>> GetNearestByFilter(WorkshopFilter
...
+
+    private async Task<Expression<Func<Workshop, bool>>> PredicateBuildForAdminds(WorkshopFilterAdministration filter)
+    {
+        var predicate = PredicateBuilder.True<Workshop>();
+
+        // TODO: add condition for united territorial community admin
+        if (currentUserService.IsMinistryAdmin())
+        {
+            await ApplyFilterForMinistryAdmin(filter).
+                ConfigureAwait(false);
```
VadymLevkovskyi, line 695: Please check if it's necessary to change this code.

Right now `predicate` is not changed.
I believe it should (like in next `if`: `predicate = predicate.And(tempPredicate)`, where `tempPredicate` is result of `ApplyFilterForMinistryAdmin`)

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -631,6 +648,132 @@ public async Task<SearchResult<WorkshopCard>> GetNearestByFilter(WorkshopFilter
...
+            var tempPredicate = await ApplyFilterForRegionAdmin(filter).
+                ConfigureAwait(false);
+
+            predicate = predicate.And(tempPredicate);
+        }
+
+        if (!string.IsNullOrEmpty(filter.SearchString))
+        {
+            var tempPredicate = PredicateBuilder.False<Workshop>();
+            var keyWords = filter.SearchString.Split(' ', ',', StringSplitOptions.RemoveEmptyEntries);
```
VadymLevkovskyi: I worry, that `Split` arguments are incorrect. Could you double-check them?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -631,6 +648,132 @@ public async Task<SearchResult<WorkshopCard>> GetNearestByFilter(WorkshopFilter
...
+        }
+        else if (currentUserService.IsRegionAdmin())
+        {
+            var tempPredicate = await ApplyFilterForRegionAdmin(filter).
+                ConfigureAwait(false);
+
+            predicate = predicate.And(tempPredicate);
+        }
+
+        if (!string.IsNullOrEmpty(filter.SearchString))
```
VadymLevkovskyi: (Logic) `string.IsNullOrEmpty` will pass the string with spaces or commas (like `"   "` or `"   , ,  ,"`). Theses strings are bad, because in result you'll have `tempPredicate == PredicateBuilder.False<Workshop>();` and never get any data.

Please rewrite if condition to check the count of `keyWords` as well

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
VadymLevkovskyi: Wrong alert, now I see that `filter.InstitutionId` is changed inside the `ApplyFilterForMinistryAdmin` method. Not very intuitive code, but let it be as is

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceDBTests.cs
```diff
@@ -0,0 +1,211 @@
...
+    private IMapper mapper;
+    private Mock<IImageDependentEntityImagesInteractionService<Workshop>> workshopImagesMediator;
+    private Mock<IProviderAdminRepository> providerAdminRepository;
+    private Mock<IAverageRatingService> averageRatingServiceMock;
+    private Mock<IProviderRepository> providerRepositoryMock;
+    private Mock<ICurrentUserService> currentUserServiceMock;
+    private Mock<IMinistryAdminService> ministryAdminServiceMock;
+    private Mock<IRegionAdminService> regionAdminServiceMock;
+    private Mock<ICodeficatorService> codeficatorServiceMock;
+    private Guid institutionId;
```
VadymLevkovskyi: (Unused fields) Hadn't checked all of them, but at least `logger` and `providerAdminRepository` are used only locally in `SetUp()`. Please consider moving them there as local variables (or probably even inline them). Please check other variables.

Reasoning: less class fields means you have less code, control less changes, know what is really needed for tests and what is needed only for instantiating `WorkshopService`

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceDBTests.cs
```diff
@@ -0,0 +1,211 @@
...
+    {
+        // Arrange
+        var workshopDto = await MapWorkshopsToDtos();
+        var filter = new WorkshopFilterAdministration()
+        {
+            InstitutionId = institutionId,
+            SearchString = "keyWord",
+            Size = 100,
+        };
+        var expectedEntities = new List<WorkshopDto> {workshopDto[0] };
```
VadymLevkovskyi: (Test logic) What am I missing? Why test expects 0 entities? As I can see at least one entity with `keyWord` is present in seeded data.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceDBTests.cs
```diff
@@ -0,0 +1,211 @@
...
+            Entities = expectedEntities,
+        };
+
+        // Act
+        var result = await sensitiveWorkshopService.FetchByFilterForAdmins(filter).ConfigureAwait(false);
+
+        // Assert
+        result.Should().BeEquivalentTo(expectedResult);
+    }
+
```
VadymLevkovskyi, line 179: (Tests set) Could you have few tests for different "keyword" cases? Imagine there are few entities with keywords "test", "skipped", "othertest", "other" and "test other".
Could you check that 
- filter for "test" returns only "test" and "other test"
- filter for "other" returns only "other" and "other test"
- filter for "other test" returns "test", "other" and "other test"
- filter for " " (one space) - returns all, as well as filter for null or empty string

Do the same with comma separator

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceDBTests.cs
```diff
@@ -0,0 +1,211 @@
...
+    private void Seed()
```
VadymLevkovskyi, line 274: (Tests set) Could you add few tests for different "CATOTGId" cases? Different "InstitutionId" cases?  Combinations of them?

At the moment I see only limited coverage of filtering logic.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceDBTests.cs
IrynaNakoneshniuk: The test expects the first element from the list of DTOs.

VadymLevkovskyi: Indeed. I need more sleep. Thank you.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceTests.cs
```diff
@@ -0,0 +1,227 @@
...
+    private Mock<IMapper> mapperMock;
+    private Mock<IImageDependentEntityImagesInteractionService<Workshop>> workshopImagesMediator;
+    private Mock<IProviderAdminRepository> providerAdminRepository;
+    private Mock<IAverageRatingService> averageRatingServiceMock;
+    private Mock<IProviderRepository> providerRepositoryMock;
+    private Mock<ICurrentUserService> currentUserServiceMock;
+    private Mock<IMinistryAdminService> ministryAdminServiceMock;
+    private Mock<IRegionAdminService> regionAdminServiceMock;
+    private Mock<ICodeficatorService> codeficatorServiceMock;
+    private string userId;
```
VadymLevkovskyi: (Excessive fields) Same comment as in previous test class. Please don't use something as a filed if it could be a local variable.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceTests.cs
```diff
@@ -0,0 +1,227 @@
...
+        var workshops = WithListOfWorkshops().ToList();
+        var workshopsDto = WithListOfWorkshopsDto().ToList();
+        IEnumerable<long> subSettlementsIds = new List<long>() { 1, 2, 3 };
+        SetUpCurrentUserService(userId, isRegionAdmin, isMinistryAdmin);
+        SetupRegionAdminService(userId, adminRegion);
+        SetupMinistryAdminService(userId, adminMinistry);
+        SetUpCodeficatorService(subSettlementsIds, parentCATOTTGId);
+        SetUpWorkshopsRepository(workshops, filter);
+        SetUpMapper(workshopsDto, workshops);
+        var resultExcpected = new SearchResult<WorkshopDto> { TotalAmount = workshopsDto.Count, Entities = workshopsDto };
```
VadymLevkovskyi: Please use some spell checker. `Expected` is misspelt here. In other file I've seen misspelt `address`.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceTests.cs
```diff
@@ -0,0 +1,227 @@
...
+        WorkshopFilterAdministration filter = new WorkshopFilterAdministration() { Size = 8, CATOTTGId = 1, From = 2, SearchString = "key word" };
+        var resultExcpected = SetupFetchByFilterForAdmins(userId, false, true, parentCATOTTGId, filter, null, admin);
+
+        // Act
+        var result = await sensitiveWorkshopService.FetchByFilterForAdmins(filter).ConfigureAwait(false);
+
+        // Assert
+        resultExcpected.Should().BeEquivalentTo(result);
+    }
+
```
VadymLevkovskyi, line 172: Same conmment about testing keywords filtering separately and in combination with other filters

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceTests.cs
```diff
@@ -0,0 +1,227 @@
...
+        SetUpMapper(workshopsDto, workshops);
+        var resultExcpected = new SearchResult<WorkshopDto> { TotalAmount = workshopsDto.Count, Entities = workshopsDto };
+        return resultExcpected;
+    }
+
+    private IEnumerable<Workshop> WithListOfWorkshops()
+    {
+        var workshops = WorkshopGenerator.Generate(5);
+        return workshops;
+    }
```
VadymLevkovskyi: (Design) What was the intention to extract this (and next) methods. They are one-liners, called ony once and can be inlined in that single call place.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
IrynaNakoneshniuk: The arguments in the Split function correspond to the arguments in other services where filtering by search string is present. As far as I understand, the approach to splitting the string into components before starting the search is the same for all services.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceTests.cs
IrynaNakoneshniuk: Thank you, I will correct it

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceDBTests.cs
IrynaNakoneshniuk: Thank you, I will correct it

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
VadymLevkovskyi: Ok, let me double check it for you. From the screenshot below, you see that second parameter in called `Split` function is `int count`. You have no compilation errors, because there is implicit cast between `char` and `int` type. But logic is still bad. 

Please, read what you copy-paste. 
Please, don't trust others code, even your own code. Sometimes even Microsoft devs have problems in their code.
Please, double-check concerned code, especially when it was asked so during code review.
Please, fix your copy of the code, cover it with unit-tests, and create tech-debt item in backlog to fix "other services where filtering by search string is present"

Thanks in advance

image

IrynaNakoneshniuk: Sorry, I was thinking about possible punctuation marks for splitting rather than the arguments of the Split function in general. Of course, after your comment, I checked the function's behavior and its possible arguments, and I found the mistake, which I will fix - ```filter.SearchString.Split([' ', ','], StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)```. I tested how the splitting works with the specified arguments and in no case did I ignore this.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceDBTests.cs
IrynaNakoneshniuk: Added various cases for search.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceTests.cs
IrynaNakoneshniuk: Added various cases for search

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceTests.cs
IrynaNakoneshniuk: Fixed

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceDBTests.cs
IrynaNakoneshniuk: Fixed: added  inline in the constructor.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
IrynaNakoneshniuk: Fixed: modified the method to check for null or whitespace, and added a word count check for the search.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
IrynaNakoneshniuk: Thank you, corrected the mistake in the arguments.

OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/SensitiveWorkshopsServiceTests.cs
IrynaNakoneshniuk: Fixed

h4wk13 (approved):

VadymLevkovskyi (approved):

PR #1553, opened 2024-08-28
Add CompetitiveEvents

alrom:


OutOfSchool/OutOfSchool.BusinessLogic/Services/CompetitiveEventService.cs
```diff
@@ -0,0 +1,105 @@
...
+        this.competitiveEventRepository = competitiveEventRepository ?? throw new ArgumentNullException(nameof(competitiveEventRepository));
+        this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        this.localizer = localizer ?? throw new ArgumentNullException(nameof(localizer));
+        this.mapper = mapper ?? throw new ArgumentNullException(nameof(mapper));
+    }
+
+    /// <inheritdoc/>
+    public async Task<CompetitiveEventDto> GetById(Guid id)
+    {
+        logger.LogInformation($"Getting CompetitiveEvent by Id started. Looking Id = {id}.");
```
VadymLevkovskyi: (Log level) Here and everywhere - such log message looks more like Trace level message, not Information.

OutOfSchool/OutOfSchool.BusinessLogic/Services/CompetitiveEventService.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        var competitiveEvent = await competitiveEventRepository.GetById(id).ConfigureAwait(false);
+
+        if (competitiveEvent == null)
+        {
+            throw new ArgumentOutOfRangeException(
+                nameof(id),
+                localizer["The id cannot be greater than number of table entities."]);
+        }
```
VadymLevkovskyi: Have 2 thoughts about this code snippet:
1. (logic) `null` could be returned not only because of "id cannot be greater than number of table entities", there might becase when some entity was deleted, right?
2. (code style) why not use "?? throw" approach? Like 
```
var competitiveEvent = (await competitiveEventRepository.GetById(id).ConfigureAwait(false)) 
        ?? throw new ArgumentOutOfRangeException(
                nameof(id),
                localizer["The id cannot be greater than number of table entities."]);
```
this way you are ensuring that `competitiveEvent` type is not nullable

OutOfSchool/OutOfSchool.BusinessLogic/Services/CompetitiveEventService.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        if (competitiveEvent is null)
+        {
+            var message = $"Updating failed. CompetitiveEvent with Id = {dto.Id} doesn't exist in the system.";
+            logger.LogError(message);
+            throw new DbUpdateConcurrencyException(message);
+        }
+
+        mapper.Map(dto, competitiveEvent);
+        competitiveEvent = await competitiveEventRepository.Update(competitiveEvent).ConfigureAwait(false);
```
VadymLevkovskyi, line 77: (Logic) You call db twice - `GetById` and then `Update`. I wonder what's the reason to not just `Update` it (less code, better performance) and (for absent entity) have that exception thrown by EF?

OutOfSchool/OutOfSchool.BusinessLogic/Services/CompetitiveEventService.cs
```diff
@@ -0,0 +1,105 @@
...
+        var favorite = await competitiveEventRepository.GetById(id).ConfigureAwait(false);
+
+        if (favorite == null)
+        {
+            throw new ArgumentOutOfRangeException(
+                nameof(id),
+                localizer[$"CompetitiveEvent with Id = {id} doesn't exist in the system"]);
+        }
+
+        await competitiveEventRepository.Delete(favorite).ConfigureAwait(false);
```
VadymLevkovskyi: (Logic) Similar question as above - 2 calls to db: `GetById` + `Delete` are used where one `Delete` is sufficient

OutOfSchool/OutOfSchool.DataAccess/Extensions/ModelBuilderExtension.cs
```diff
@@ -242,6 +243,84 @@ public static void Seed(this ModelBuilder builder)
...
+                Id = 3,
+                Title = "Районний",
+                TitleEn = "Regional",
+            },
+            new CompetitiveEventCoverage
+            {
+                Id = 4,
+                Title = "Обласний",
+                TitleEn = "Provincial",
+            },
```
VadymLevkovskyi: (Translation, not sure) Please double-check translations.

Here you've introduced 2 translations I'm not happy about:
"Районний" = "Regional", "Обласний" = "Provincial"

My expectations for english translations are "Raional" and "Regional", respectively. Have checked it in [wiki](https://en.wikipedia.org/wiki/Administrative_divisions_of_Ukraine)

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/CompetitiveEventCoverageConfiguration.cs
```diff
@@ -0,0 +1,17 @@
...
+
+internal class CompetitiveEventCoverageConfiguration : IEntityTypeConfiguration<CompetitiveEventCoverage>
+{
+    public void Configure(EntityTypeBuilder<CompetitiveEventCoverage> builder)
+    {
+        builder.HasKey(x => x.Id);
+
+        builder.HasIndex(x => x.IsDeleted);
+
+        builder.Property(x => x.IsDeleted).HasDefaultValue(false);
```
VadymLevkovskyi: (Code reuse) Same code for all `IKeyed` and `ISoftDeleted` entities. Why not introduce some helper method?

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/CompetitiveEventController.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// Delete a specific CompetitiveEvent from the database.
+    /// </summary>
+    /// <param name="id">CompetitiveEvent id.</param>
+    /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
+    //[HasPermission(Permissions.CompetitiveEventRemove)]
+    [Authorize]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpDelete("{id}")]
```
VadymLevkovskyi, line 93: (OpenApi doc) Seems missed `[ProducesResponseType(StatusCodes.Status400BadRequest)]`

VadymLevkovskyi (commented):
See no isssue with logic, just few comments

OutOfSchool/OutOfSchool.BusinessLogic/Services/CompetitiveEventService.cs
alrom: Done. We use LogTrace, LogInformation and LogDebug in the different services.

OutOfSchool/OutOfSchool.DataAccess/Extensions/ModelBuilderExtension.cs
alrom: Changed

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/CompetitiveEventCoverageConfiguration.cs
alrom: Done

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/CompetitiveEventController.cs
alrom: Fixed

OutOfSchool/OutOfSchool.BusinessLogic/Services/CompetitiveEventService.cs
alrom: Thanks. Done.

OutOfSchool/OutOfSchool.BusinessLogic/Services/CompetitiveEventService.cs
alrom: Was changed

OutOfSchool/OutOfSchool.BusinessLogic/Services/CompetitiveEventService.cs
alrom: It would be nice if the structure of the DTO and the base model were the same.

How will the method work if the DTO model is not equal to the model of an entity?
Required fields will throw exceptions and missed fields will be clear?

This is the common approach of Update in our solution.
In the nearest future, we will change this method to some different partial updates.

VadymLevkovskyi: Thnks man, I was inattentive and didn't noticed that we read current value of entity to perform correct update for not-mapped properties. Yes it looks weird for me to have different data in dto and entity, but for this specific PR it shouldn't be addressed. 
Resolved

VadymLevkovskyi:
See no issues in code. But test coverage could be improved

alrom:
> See no issues in code. But test coverage could be improved

Tests were added. Uncovered code - is code for models ))

VadymLevkovskyi (approved):

PR #1555, opened 2024-09-04
Initial distribution and adding new model of user

ZdorovenkoKyrylo:


VadymLevkovskyi (approved):

OutOfSchool/OutOfSchool.DataAccess/Models/User.cs
```diff
@@ -9,32 +9,32 @@ public class User : IdentityUser, IKeyedEntity<string>, ISoftDeleted
...
-    public bool IsBlocked { get; set; } = false;
-
     [MaxLength(50)]
     public string Role { get; set; }
 
     public bool IsRegistered { get; set; }
 
-    // for permissions managing at login and check if user is original provider or its admin
+    // If the flag is true, that user can no longer do anything to website.
+    public bool IsBlocked { get; set; } = false;
```
h4wk13, line 35: I suggest storing such users like blocked or not confirmed their registration yet in a separate table and use a separate service to handle them.

h4wk13 (approved):
Issues with users data that I mentioned are not critical for now but might impact performance and cause additional maintaining work in the future.

PR #1558, opened 2024-09-09
Added check is workshop is blocked

### LINKED ISSUE #1557: [Messages] The service provider blocked by the administrator can correspond with the parent
**Environment:** Windows 11, 64-bit, Google Chrome v.128.0.6613.114 
**Reproducible:** always. 
**Build found:** e6f0d4a

**Preconditions**

1. Go to https://oos.dmytrominochkin.cloud/;
2. In the correspondence history, there must be at least one message;
3. The service provider must be blocked by the administrator.

**Steps to reproduce**

1. Log in as a parent.
2. Go to the Messages page;
3. Select from the list of messages from the blocked service provider;
4. Send a message to the blocked service provider;
5. Log in as a service provider.
6. Open the Messages page;
7. Ensure that the message from the parent has been received by the service provider;
8. Open the message and write to the parent;
9. Note the result.

**Actual result**
The service provider and the parent can correspond when the service provider is blocked by the administrator.
image
image

**Expected result**
The parent and the blocked service provider cannot send messages to each other.

**Labels to be added**
"Bug", Priority ("Priority: Hight"), Severity ("Severity: Critical"), Type ("Functional").

olsavchenk:
#1557

### DIFF
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Hubs/ChatWorkshopHub.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Hubs/ChatWorkshopHub.cs
@@ -256,7 +256,7 @@ public class ChatWorkshopHub : Hub
 
         bool isParentBlocked = await blockedProviderParentService.IsBlocked(parentId, workshop.ProviderId);
 
-        if (isParentBlocked)
+        if (isParentBlocked || workshop.IsBlocked)
         {
             return false;
         }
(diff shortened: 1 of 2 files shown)


OutOfSchool/OutOfSchool.BusinessLogic/Hubs/ChatWorkshopHub.cs
```diff
@@ -256,7 +256,7 @@ private async Task<bool> UserHasRigtsForChatRoomAsync(Guid workshopId, Guid pare
```
h4wk13: Correct me if I wrong but:
1. there is a call to the database for the workshop, 
2. caller method makes almost the same call to the workshop service,

Why do we need 2 calls to the database here ?
Could you please check that and refactor it if possible ?

olsavchenk: I don't quite understand what you mean. UserHasRigtsForChatRoomAsync method has only one call to the workshop repository. Could you please provide more details?

PR #1563, opened 2024-09-18
Kymachynskyi/ Migrate ElasticSearch to version 8.15.6

DKymachynskyi:


OutOfSchool/OutOfSchool.BusinessLogic/Services/Elasticsearch/ElasticIndexEnsureCreatedHostedService.cs
```diff
@@ -0,0 +1,64 @@
...
+        using var scope = services.CreateScope();
+        var elasticHealthService = scope.ServiceProvider.GetService<IElasticsearchHealthService>();
+        while (!elasticHealthService.IsHealthy
+                && (DateTimeOffset.UtcNow.ToUnixTimeSeconds() - startTime < connectionWaitingTimeSec))
+        {
+            logger.LogInformation("Waiting for Elastic connection");
+            await Task.Delay(checkConnectivityDelayMs).ConfigureAwait(false);
+        }
+
+        if (elasticHealthService.IsHealthy)
```
h4wk13, line 52: should not we fail if there is no connection to elastic ?
If this service is not required - why do we wait for the connection ?
Logs ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Elasticsearch/ElasticIndexEnsureCreatedHostedService.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            var existsResponse = await client.Indices.ExistsAsync(indexName).ConfigureAwait(false);
+            if (existsResponse.ApiCallDetails.HttpStatusCode == StatusCodes.Status404NotFound)
```
h4wk13: does that client have a resiliency configured ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Elasticsearch/ElasticPinger.cs
```diff
@@ -26,7 +26,19 @@ public Task StartAsync(CancellationToken cancellationToken)
     {
         logger.LogInformation("Service started pinging");
         timer = new Timer(
-            cb => IsHealthy = elasticClient.Ping().IsValid,
+            async (cb) =>
```
h4wk13: what `cb` means here ?
I cannot see where we're using it? can it be changed to `_` ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Elasticsearch/ElasticsearchSynchronizationService.cs
```diff
@@ -141,7 +141,7 @@ private async Task<bool> Synchronize(
             {
                 var result = await esProvider.DeleteRangeOfEntitiesByIdsAsync(ids).ConfigureAwait(false);
 
-                if (result == Result.Error)
+                if (result != Result.Deleted)
```
h4wk13, line 145: why it is changed ?

h4wk13 (approved):

OutOfSchool/OutOfSchool.BusinessLogic/Services/Elasticsearch/ElasticPinger.cs
DKymachynskyi: Fixed. Thank you!

OutOfSchool/OutOfSchool.BusinessLogic/Services/Elasticsearch/ElasticsearchSynchronizationService.cs
DKymachynskyi: It has been changed because the Result enum in Elasticsearch 8.15 no longer contains an Error element.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Elasticsearch/ElasticIndexEnsureCreatedHostedService.cs
DKymachynskyi: No, we are using one node for now.
According to the Elasticsearch 8.15 documentation, `ExistsAsync` returns only codes 200 or 404, еxceptions are not used.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Elasticsearch/ElasticIndexEnsureCreatedHostedService.cs
DKymachynskyi: > should not we fail if there is no connection to elastic ?

I discussed this question with @DmyMi, and he suggested adding messages to the log without returning errors, since if Elasticsearch is not working, the application uses a different strategy for search.
I've added logging to indicate when an Elasticsearch index is created successfully and when Elasticsearch index failed to ensure due to a connection timeout.

> If this service is not required - why do we wait for the connection ?

Whether the service is required or not is determined by the options' settings.
If the service is enabled through the options, we wait for a connection in case, for example, Elasticsearch is temporarily unavailable.

PR #1564, opened 2024-09-18
Nakoneshniuk/fixed total amount workshops returned

IrynaNakoneshniuk:
### Summary of Changes:
- Fixed the filtering behavior by `InstitutionId` when the user has the `regionadmin` or `ministryadmin` role. 
- Now, if the `InstitutionId` in the filter differs from the `InstitutionId` associated with the user's role, an empty list is returned instead of the full list of workshops belonging to the  role.
- Fixed the `totalAmount` value, which now correctly reflects the total number of workshops found by the filter.

### Testing:
- Added tests to verify that an empty list is returned when the `InstitutionId` in the filter does not match the user's role.

### DIFF
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
@@ -694,15 +694,15 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
         // TODO: add condition for united territorial community admin
         if (currentUserService.IsMinistryAdmin())
         {
-            await ApplyFilterForMinistryAdmin(filter).
-                ConfigureAwait(false);
+            var predicateForMinAdmin = await PredicateForRoleMinistryAdmin()
+                .ConfigureAwait(false);
+            predicate = predicate.And(predicateForMinAdmin);
         }
         else if (currentUserService.IsRegionAdmin())
         {
-            var tempPredicate = await ApplyFilterForRegionAdmin(filter).
-                ConfigureAwait(false);
-
-            predicate = predicate.And(tempPredicate);
+            var predicateForRegionAdmin = await PredicateForRoleRegionAdmin()
+                .ConfigureAwait(false);
+            predicate = predicate.And(predicateForRegionAdmin);
         }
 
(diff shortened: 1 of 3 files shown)


OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -743,17 +746,20 @@ await ApplyFilterForMinistryAdmin(filter).
         return predicate;
     }
 
-    private async Task ApplyFilterForMinistryAdmin(WorkshopFilterAdministration filter)
+    private async Task<Expression<Func<Workshop, bool>>> PredicateForRoleMinistryAdmin()
     {
+        var predicate = PredicateBuilder.True<Workshop>();
```
h4wk13: why do you need it here ?
you can create it in single line.
Isa it possible to pass all the data as a parameters ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -767,20 +773,19 @@ private async Task<Expression<Func<Workshop, bool>>> ApplyFilterForRegionAdmin(W
...
-        var tempPredicate = PredicateBuilder.False<Workshop>();
+        predicate = predicate.And(x => subSettlementsIds.Contains(x.Address.CATOTTGId));
 
-        foreach (var item in subSettlementsIds)
-        {
-            tempPredicate = tempPredicate.Or(x => x.Address.CATOTTGId == item);
-        }
-
-        return tempPredicate;
+        return predicate;
```
h4wk13, line 798: could you please reorder the code so the predicate will be built and returned in single line ?

h4wk13 (approved):

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
IrynaNakoneshniuk: Thank you.  I separated the logic of predicate building and database queries into different methods, so the data is passed as parameters.

DmyMi (approved):

Issue #1534, opened 2024-07-30
[Data model] Workshop (Study group)


DmyMi:
# Summary

Unified information about a group of education seekers (class, study group, extended day group, workshop group, etc.), which is formed within a certain educational institution of any type or level in order to ensure the teaching of a certain set of educational subjects to this group, regardless of the peculiarities of the educational process.

Can include hierarchy of sections which are also a workshop (study group).

**Is a business entity as in #1524**

# Entity attributes

Name | Data Type | Mandatory | Comment
-- | -- | -- | --
identifier | GUID | TRUE |
CoverImage | string | TRUE |
Images | list | TRUE |
Minimum age | int | TRUE |
Maximum age | int | TRUE |
Total number of seats in the group | int | TRUE | 
educational shift | Dictionary to educational shift | TRUE | _Edited 21.09.2024_ Dictionary will be implemented on Stage II. Currently can be a stub field.
Short stay | boolean | TRUE |
Language of education | Dictionary of languages ​​of study | TRUE | The dictionary is implemented in the current AIKOM, we can reuse it for the 1st stage
schedule | Schedule | TRUE |
Works with funds from parents, benefactors | boolean | TRUE |
Type of group by age composition | Type of group by age composition | TRUE | see below
there is a competitive selection in the group | boolean | TRUE |
number of seats | int | TRUE | calculated from available seats minus taken
study from | Date | TRUE |
study to | Date | TRUE |
Educational disciplines that are taught | `Educational subject` #1533 | TRUE | = Subcategory of group direction. _Edited 21.09.2024_: Will be implemented on stage II
workshop title | String | TRUE |
direction | Dictionary of directions of the group | TRUE | see below. _Edited 21.09.2024_: Can re-use our current mechanism.
Category | `Direction` | TRUE | see below
formed in provider| `Provider` #1543 | TRUE |
group type | Classifier of types of educational groups | TRUE | _Edited 21.09.2024_: can be a stub, no dictionary at the moment.
member of the member of the study group (children) | TRUE | We will put it on BE, on the FE we will not add it until 2nd stage
is paid | boolean | TRUE |
is special | boolean | TRUE |
is inclusive | boolean | TRUE |
 |  |  | 
Number of occupied seats | int | FALSE | At the both workshop and section level
includes study groups | `Workshop (Study group)` | FALSE | links to other study groups, if this group is consolidated
group with special needs | Dictionary of types of special needs of children | FALSE | see below
[...]

PR #1576, opened 2024-10-15
Hlib bondarev/workshop entity with address

### LINKED ISSUE #1544: [Workshop] Create workshop
**Role**: User (provider's employee)

**Story**: I, as a User, want to create a study group (workshop) for further display of information about the workshop on the Out of School platform.

# Acceptance Criteria

### Access to the workshop creation interface

- GIVEN user wants to create a workshop
- WHEN the user selects the "Create" option in the main menu of the service provider's account
- THEN the workshop creation interface opens

### Entering basic information about the workshop

- GIVEN the user fills in basic information about the workshop
- WHEN the user enters the name of the workshop, description, age category, planned schedule and other necessary attributes
- THEN the system saves the entered information

### Adding a description of the workshop, category and subcategory, images

- GIVEN the user wants to describe the workshop in detail
- WHEN the user enters a description of the workshop, selects a category and subcategory, uploads images and other media files
- THEN system saves all entered information and media files

### Adding contact information

- GIVEN user wants to add contact information for feedback
- WHEN the user enters phone, email and other contact data
- THEN system stores contact information

### Adding a teacher

- GIVEN user wants to add teacher to workshop
- WHEN the user enters the relevant information about the teacher
- THEN system stores information about the teacher

> for `Stage I` re-use current logic, during `Stage II` there should be a full addition from the list of employees of the provider

### Saving workshop draft (_Edited 16.10.2024_)

- GIVEN the user has filled in all the necessary information about the workshop
- WHEN user clicks "Save the draft"
- THEN system saves the entered information in `Draft` status
- AND the draft of the workshop is saved and available for further editing

### Sending a draft for moderation (_Edited 16.10.2024_)

- GIVEN the workshop is in `DRAFT` status
- WHEN user clicks "Send for moderation" option
- THEN the system changes the status to `PENDING MODERATION`
- AND The workshop becomes available for review by the moderator

### Moderation & Publication (_Edited 16.10.2024_)

- GIVEN the workshop is in `PENDING MODERATION` status
- WHEN the moderator checks the information and approves the workshop for publication
- THEN the system changes the workshop status to `ACTIVE` and it becomes available for viewing in the System

## Business rules, validations, alternatives

- Data that the User must enter when creating a workshop in all mandatory fields of entities:
  - Workshop #1534
  - Contacts #1525
- A users can create a workshop only if they have the appropriate access rights.
- A user can edit a draft an unlimited number of times before publishing.
- The user can delete the draft of the workshop without its further saving and display in the system (hard delete) before its transition to the next status.
- All mandatory fields must be filled in to save the workshop.
- Publishing a workshop involves prior moderation to ensure compliance with established standards.
- If the user did not fill in the mandatory fields, the system displays an error message and does not save the workshop.
- If a moderator rejects a workshop, the system returns it to `Draft` status with comments for correction. _Edited 16.10.2024_
- _Each step before final "save" should be saved in cache (`stage I` - local), so even if users refresh the page - they will be notified they have a draft to finish._

## List of layouts
- "My workshops" tab in the provider's cabinet with the "Create" button.
- Workshop creation interface:
  - Interface for adding basic information
  - Interface for adding description, category and subcategory, images
  - Interface for adding contact information (phone, email, etc.)
  - The interface for adding a teacher
- Confirmation of intention to create a workshop (for example, a popup).
- Notification of successful workshop creation.
- Error message.

HlibBondarev:
1) Updated Workshop entity - added new properties according to ISSUE - https://github.com/ita-social-projects/OoS-Backend/issues/1534.
2) Updated WorkshopBaseDto - added new properties (corresponding to new properties of the Workshop entity).
3) Updated WorkshopService methods Create() and CreateV2 for filling new properties of Workshop entity according to ISSUE - https://github.com/ita-social-projects/OoS-Backend/issues/1544..
4) Updated MappingProfile class for mapping new added properties of Workshop and Workshop dtos..
5) Updated TeacherConfiguration and WorkshopConfiguration classes to set new relations between entities.
6) Updated Create actions for V1-WorkshopController and V2-WorkshopController.
7) Added new NUnit tests for check new code.


AndrD1406 (approved):
Look well!
At least everything related to the tags entity

OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/WorkshopBaseDto.cs
```diff
@@ -55,21 +55,25 @@ public class WorkshopBaseDto : IValidatableObject
     [Range(0, 120, ErrorMessage = "Max age should be a number from 0 to 120")]
     public int MaxAge { get; set; }
 
-    [Required]
     [ModelBinder(BinderType = typeof(JsonModelBinder))]
+    [CollectionNotEmpty(ErrorMessage = "At least one DateTime range is required")]
     public List<DateTimeRangeDto> DateTimeRanges { get; set; }
 
+    [Required]
```
h4wk13: does this attribute make sense for non nullable fields ?

OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/WorkshopBaseDto.cs
```diff
@@ -125,6 +126,64 @@ public class WorkshopBaseDto : IValidatableObject
...
+    public virtual ICollection<WorkshopBaseDto> IncludedStudyGroups { get; set; } // Navigation property to included study groups
+
+    [Required]
+    public long AddressId { get; set; }
+
+    [Required]
+    [ModelBinder(BinderType = typeof(JsonModelBinder))]
+    public AddressDto Address { get; set; }
+
     public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
```
h4wk13, line 182: I think it would be better to have it like extension method but not a part of the DTO

OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/WorkshopBaseDto.cs
```diff
@@ -125,6 +126,64 @@ public class WorkshopBaseDto : IValidatableObject
...
+    [ModelBinder(BinderType = typeof(JsonModelBinder))]
+    public virtual ICollection<WorkshopBaseDto> IncludedStudyGroups { get; set; } // Navigation property to included study groups
+
+    [Required]
+    public long AddressId { get; set; }
+
+    [Required]
+    [ModelBinder(BinderType = typeof(JsonModelBinder))]
+    public AddressDto Address { get; set; }
+
```
h4wk13, line 181: do we really need that DTO as a whole ?
I mean, 185 lines ? seriously ?
Can we split that to parts ?

OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/WorkshopDto.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace OutOfSchool.BusinessLogic.Models.Workshops;
 
-public class WorkshopDto : WorkshopBaseDto, IHasRating
+public class WorkshopDto : WorkshopCreateUpdateDto, IHasRating
 {
     public uint TakenSeats { get; set; } = 0;
```
h4wk13, line 13: why do we need this trailing `0`  ? Is it not a default value for `uint` ?
Why do we need default values in models at all ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/ITeacherService.cs
```diff
@@ -49,4 +49,11 @@ public interface ITeacherService
     /// <param name="teacherId">Teacher's key.</param>
     /// <returns>A <see cref="Task"/> representing the result of the asynchronous operation.</returns>
     Task<Guid> GetTeachersWorkshopId(Guid teacherId);
+
+    /// <summary>
+    /// Check if entity is exists by it's key.
+    /// </summary>
+    /// <param name="id">Key in the table.</param>
+    /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.
+    Task<bool> Exists(Guid id);
```
h4wk13: `ExistsAsync` ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/TeacherService.cs
```diff
@@ -176,6 +163,27 @@ public async Task<Guid> GetTeachersWorkshopId(Guid teacherId)
...
+
+    private async Task UpdateTeacher()
+    {
+        try
+        {
+            await teacherRepository.SaveChangesAsync().ConfigureAwait(false);
+        }
+        catch (DbUpdateException ex)
+        {
+            logger.LogError(ex, "Unreal to update teacher.");
```
h4wk13: Unreal -> Unable

OutOfSchool/OutOfSchool.BusinessLogic/Services/TeacherService.cs
```diff
@@ -176,6 +163,27 @@ public async Task<Guid> GetTeachersWorkshopId(Guid teacherId)
...
+            throw;
+        }
+    }
+
+    /// <inheritdoc/>
+    public Task<bool> Exists(Guid id)
+    {
+        logger.LogInformation("Checking if Teacher exists by Id started. Looking Id = {id}.", id);
```
h4wk13: this is Trace or Debug but not Info message

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/WorkshopConfiguration.cs
```diff
@@ -13,25 +11,35 @@ public override void Configure(EntityTypeBuilder<Workshop> builder)
...
+            .HasForeignKey<Workshop>(x => x.DefaultTeacherId) // Shadow property for One-to-One relationship
+            .IsRequired(false) // Optional relationship
+            .OnDelete(DeleteBehavior.Restrict);
+
+        // One-to-Many relationship (with shadow property in Workshop)
+        builder.HasMany(x => x.Teachers)
+            .WithOne()
+            .HasForeignKey(x => x.WorkshopId) // Shadow property for One-to-Many relationship
+            .IsRequired(false) // Optional relationship
+            .OnDelete(DeleteBehavior.Cascade);
```
h4wk13, line 31: Is it possible that teacher can be assigned to more than one workshop ?

OutOfSchool/OutOfSchool.DataAccess/Models/Workshop.cs
```diff
@@ -55,60 +31,113 @@ public class Workshop : BusinessEntity, IImageDependentEntity<Workshop>, IHasEnt
...
+    public uint SpecialNeedsId { get; set; } = uint.MinValue;
 
-    [MaxLength(200)]
-    public string Keywords { get; set; } = string.Empty;
+    [Required(ErrorMessage = "Property IsInclusive is required")]
+    public bool IsInclusive { get; set; } = default;
 
-    [Required]
-    public PayRateType PayRate { get; set; }
+    [MaxLength(500)]
```
h4wk13: I think it should be a const

OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/WorkshopBaseDto.cs
HlibBondarev: O.k. 
Removed [Required] attribute from the bool type properties - IsPaid, IsInclusive, IsSelfFinanced and IsSpecial of WorkshopBaseDto class.

HlibBondarev: WorkshopBaseDto class already uses data annotation attributes (like [Required], [MinLength], [MaxLength], [Range]) etc. Also we use checking  ModelState.IsValid property in Create() action of WorkshopControllerso so  implementing IValidatableObject provides a single, unified validation mechanism.

OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/WorkshopBaseDto.cs
HlibBondarev: WorkshopBaseDto class already uses data annotation attributes (like [Required], [MinLength], [MaxLength], [Range]) etc. Also we use checking ModelState.IsValid property in Create() action of WorkshopControllerso so implementing IValidatableObject provides a single, unified validation mechanism.

OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/WorkshopBaseDto.cs
HlibBondarev: I'm thinking of splitting Workshop BaseDto into parts when I work on implementing draft saving in the backend. Then I plan to use a sequence of model inheritance.

OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/WorkshopDto.cs
HlibBondarev: O.k.
Removed default value for TakenSeats property of WorkshopDto.

OutOfSchool/OutOfSchool.BusinessLogic/Services/ITeacherService.cs
HlibBondarev: O.k. 
Reworked Exists() method to async method => ExistsAsync() in ITeacherService interface and its implementation in TeacherService class.

OutOfSchool/OutOfSchool.BusinessLogic/Services/TeacherService.cs
HlibBondarev: O.k.
Fixed Unreal -> Unable in TeacherService class.

OutOfSchool/OutOfSchool.BusinessLogic/Services/TeacherService.cs
HlibBondarev: O.k.
Changed - used logger.LogDebug istead of logger.LogInformation in Exists() method of TeacherService class.

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/WorkshopConfiguration.cs
HlibBondarev: In this version - for each Workshop is creating new List of new Teachers.

OutOfSchool/OutOfSchool.DataAccess/Models/Workshop.cs
HlibBondarev: O.k. 
Create new constants - MaxAdditionalDescriptionLength, MaxPreferentialTermsOfParticipationLength and MaxKeywordsLength in Constants class. Used these constants for [MaxLength] attributes in Workshop class.

OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/WorkshopBaseDto.cs
```diff
@@ -125,6 +125,61 @@ public class WorkshopBaseDto : IValidatableObject
 
     public DateOnly ActiveTo { get; set; }
 
+    public bool ShortStay { get; set; } = false;
+
+    public bool IsSelfFinanced { get; set; } = false;
+
+    public bool IsSpecial { get; set; } = false;
+
+    public uint SpecialNeedsId { get; set; } = uint.MinValue;
```
h4wk13: why `uint.MinValue` is better than nothing (which is `0`) or `0` or `default` ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -157,23 +138,17 @@ public async Task<WorkshopResultDto> CreateV2(WorkshopV2Dto dto)
         _ = dto ?? throw new ArgumentNullException(nameof(dto));
         logger.LogInformation("Workshop creating was started.");
 
+        // TODO: after refactoring the DTOs for the Workshop entities, this method needs to be replaced with the correct mapping
+        await SetIdsToDefaultValue(dto); // This method sets the properties with the Id to the default value.
```
h4wk13, line 142: why do we need that ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -1165,4 +1140,77 @@ await UpdateStatus(new()
             }).ConfigureAwait(false);
         }
     }
+
+    private async Task<Workshop> CheckDtoAndPrepareCreatedWorkshop(WorkshopCreateUpdateDto dto)
+    {
+        if (dto.MemberOfWorkshopId.HasValue && !await Exists((Guid)dto.MemberOfWorkshopId).ConfigureAwait(false))
```
h4wk13, line 1146: `why do we need that (Guid)` cast here ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -1165,4 +1140,77 @@ await UpdateStatus(new()
...
+        createdWorkshop.ProviderOwnership = createdWorkshop.Provider.Ownership;
+        createdWorkshop.ProviderTitle = createdWorkshop.Provider.FullTitle;
+        createdWorkshop.ProviderTitleEn = createdWorkshop.Provider.FullTitleEn;
+
+        if (!dto.Teachers.IsNullOrEmpty())
+        {
+            createdWorkshop.Teachers = dto.Teachers.Select(mapper.Map<Teacher>).ToList();
+        }
+
+        createdWorkshop.Tags = (await tagRepository.GetAll()).Where(tag => dto.TagIds.Contains(tag.Id)).ToList();
```
h4wk13: this request looks strange.
So we are getting all tags and then filters them in memory ?
Can we add Where method to the repository instead ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -1165,4 +1140,77 @@ await UpdateStatus(new()
...
+        createdWorkshop.Status = WorkshopStatus.Open;
+
+        return createdWorkshop;
+    }
+
+    private async Task SetIdsToDefaultValue(WorkshopCreateUpdateDto dto)
+    {
+        dto.Id = Guid.Empty;
+
+        if (dto.Address is not null)
```
h4wk13, line 1190: why do we need all these checks if we set default values ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -1165,4 +1140,77 @@ await UpdateStatus(new()
...
+        {
+            dto.DefaultTeacher.Id = Guid.Empty;
+        }
+
+        if (dto.MemberOfWorkshop is not null)
+        {
+            dto.MemberOfWorkshop.Id = Guid.Empty;
+        }
+
+        dto.WorkshopDescriptionItems?.ToList().ForEach(e => e.Id = Guid.Empty);
```
h4wk13, line 1200: so in a worse case we will have a list of a billion items with default ID ? Why ?

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -1165,4 +1140,77 @@ await UpdateStatus(new()
...
+            createdWorkshop.Teachers = dto.Teachers.Select(mapper.Map<Teacher>).ToList();
+        }
+
+        createdWorkshop.Tags = (await tagRepository.GetAll()).Where(tag => dto.TagIds.Contains(tag.Id)).ToList();
+        createdWorkshop.Status = WorkshopStatus.Open;
+
+        return createdWorkshop;
+    }
+
+    private async Task SetIdsToDefaultValue(WorkshopCreateUpdateDto dto)
```
h4wk13: is there at least one good reason why we cannot create a new object instead of that ?

OutOfSchool/OutOfSchool.WebApi/Controllers/V2/WorkshopController.cs
```diff
@@ -146,18 +161,46 @@ public async Task<IActionResult> GetByFilter([FromQuery] WorkshopFilter filter,
...
+        {
+            return StatusCode(403, "Forbidden to create workshops at blocked providers");
+        }
+
+        if (await IsCurrentUserBlocked())
+        {
+            return StatusCode(403, "Forbidden to create the workshop by the blocked provider.");
+        }
+
+        if (!ModelState.IsValid)
```
h4wk13, line 179: should not this check be performed before checks for blocked users ?

OutOfSchool/OutOfSchool.WebApi/Controllers/V2/WorkshopController.cs
```diff
@@ -146,18 +161,46 @@ public async Task<IActionResult> GetByFilter([FromQuery] WorkshopFilter filter,
...
-        dto.Address.Id = default;
+        try
+        {
+            var creationResult = await combinedWorkshopService.Create(dto).ConfigureAwait(false);
 
-        var creationResult = await combinedWorkshopService.Create(dto).ConfigureAwait(false);
+            // here we will get "false" if workshop was created by assistant provider admin
+            // because user is not currently associated with new workshop
+            // so we can update information to allow assistant manage created workshop
+            if (!await IsUserProvidersOwnerOrAdmin(creationResult.Workshop.ProviderId, creationResult.Workshop.Id).ConfigureAwait(false))
```
h4wk13, line 197: it does not look like a controller logic, as for me it is more business logic and should be moved to the service

OutOfSchool/OutOfSchool.WebApi/Startup.cs
```diff
@@ -191,7 +192,12 @@ public static async Task AddApplicationServices(this WebApplicationBuilder build
                         Duration = cacheProfilesConfig.PublicDurationInSeconds,
                     });
             })
-            .AddNewtonsoftJson()
+
+            // to handle self-references during serialization
+            .AddNewtonsoftJson(options =>
```
h4wk13, line 197: are you sure it won't be an issue ?
I mean, we should find the root cause but not ignore the issue

OutOfSchool/OutOfSchool.WebApi/Startup.cs
```diff
@@ -191,7 +192,12 @@ public static async Task AddApplicationServices(this WebApplicationBuilder build
                         Duration = cacheProfilesConfig.PublicDurationInSeconds,
                     });
             })
-            .AddNewtonsoftJson()
+
+            // to handle self-references during serialization
```
h4wk13, line 196: This comment explain 'what' this code is doing which is bad, but does not explain 'why' this code it is here.

OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/WorkshopBaseDto.cs
HlibBondarev: Ok, replaced 'uint.MinValue' with '0' for uint properties of WorkshopBaseDto class.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
HlibBondarev: We get the WorkshopCreateUpdateDto object from Frontend. It may contain not only null value but also invalid values, so we will have to replace them with the default value in the SetIdsToDefaultValue() method.
To avoid this, we can do it using mapping.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
HlibBondarev: MemberOfWorkshopId property of WorkshopBaseDto class has to be 'Guid?' type. Because of created Workshop may not be a member of other Workshop.
The Exists(Guid id) method has an argument of type 'Guid', not 'Guid?'. I don't want to overload or change this method, so I cast 'Guid?' to 'Guid'.
It's cast safely because I check dto.MemberOfWorkshopId.HasValue before doing so.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
HlibBondarev: Ok, I used tagRepository.GetByFilterNoTracking(tag => dto.TagIds.Contains(tag.Id)) method instead of (await tagRepository.GetAll()).Where(tag => dto.TagIds.Contains(tag.Id)).

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
HlibBondarev: We get the WorkshopCreateUpdateDto object from Frontend. It may contain not only null value but also invalid values, so we will have to replace them with the default value in the SetIdsToDefaultValue() method.
The Address property  of WorkshopBaseDto may be empty, so I need to check the Address before setting the Id property in Address object.
The same situation with the DefaultTeacher and MemberOfWorkshop properties.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
HlibBondarev: We will have as many items as will be transferred to DTO from the frontend and we need to set all of them value - Guid.Empty. Then DB will be generate new Id for WorkshopDescriptionItems. If we don't set Guid.Empty DB will use passed through Dto Id value. This will throw an exception if we pass the same Id value multiple times.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
HlibBondarev: We create a new Workshop from WorkshopCreateUpdateDto which we get from frontend. It may have contain invalid Id value in passed Dto for properties - Address, DefaultTeacher, MemberOfWorkshop, WorkshopDescriptionItems, Teachers, etc...

OutOfSchool/OutOfSchool.WebApi/Controllers/V2/WorkshopController.cs
HlibBondarev: I assumed that check - 'is user blocked?' would take less time than validation of Model. That's why I chose this order. Is this wrong?

OutOfSchool/OutOfSchool.WebApi/Controllers/V2/WorkshopController.cs
HlibBondarev: I didn't do any WorkshopController logic (for V2), I just used the logic from WorkshopController (for V1) and put it in the Tray-Catch block.

OutOfSchool/OutOfSchool.WebApi/Startup.cs
HlibBondarev: I need to handle self-references during serialization for WorkshopCreateUpdateDto because of Workshop entity has self-references.   
Workshop can have Workshops inside itself, also Workshop can be part of another Workshop. For these relationships are used MemberOfWorkshop and IncludedStudyGroups properties.

OutOfSchool/OutOfSchool.WebApi/Startup.cs
HlibBondarev: See above

DmyMi (approved):

PR #1580, opened 2024-10-21
add encryption library service

DmyMi:


OutOfSchool/OutOfSchool.Encryption/Config/Proxy.cs
```diff
@@ -0,0 +1,14 @@
+namespace OutOfSchool.Encryption.Config;
+
+public class Proxy
+{
+    public bool Use { get; set; }
+
+    public string Address { get; set; }
+
+    public string Port { get; set; }
```
VadymLevkovskyi: Is it correct ot have `Port` as a string?

OutOfSchool/OutOfSchool.Encryption/Models/CASettings.cs
```diff
@@ -0,0 +1,20 @@
+namespace OutOfSchool.Encryption.Models;
+
+public class CASettings
+{
+    public string[] issuerCNs { get; set; }
```
VadymLevkovskyi: (Naming) Properties are not PascalCased

OutOfSchool/OutOfSchool.Encryption/Models/CASettings.cs
```diff
@@ -0,0 +1,20 @@
...
+
+    public string address { get; set; }
+
+    public string ocspAccessPointAddress { get; set; }
+
+    public string ocspAccessPointPort { get; set; }
```
VadymLevkovskyi: ("Stringly-typed" code) Why port (`ocspAccessPointPort` and `tspAddressPort`) has string type?

OutOfSchool/OutOfSchool.Encryption/Models/EnvelopedUserInfoResponse.cs
```diff
@@ -0,0 +1,10 @@
+namespace OutOfSchool.Encryption.Models;
+
+public class EnvelopedUserInfoResponse
+{
+    public string EncryptedUserInfo { get; set; }
+
+    public int Error { get; set; }
```
VadymLevkovskyi: Is this `int Error` something related to `HttpStatusCode` standard enum?

OutOfSchool/OutOfSchool.Encryption/Handlers/AppHandlers.cs
```diff
@@ -0,0 +1,55 @@
+using Asp.Versioning.Builder;
+using Microsoft.AspNetCore.Mvc;
+using OutOfSchool.Encryption.Models;
+using OutOfSchool.Encryption.Services;
+
+namespace OutOfSchool.Encryption.Handlers;
+
+public static class AppHandlers
+{
+    public static void Map(WebApplication app, ApiVersionSet apiVersionSet)
```
VadymLevkovskyi: Such methods are usually implemented as extension methods (`public static WebApplication MapEUSignOAuth(this WebApplication app...`), so that they can be chained and called in similar way as standard/other 3rd party implementations.
That's not strict rule, just informaly established convention

OutOfSchool/OutOfSchool.Encryption/Config/EUSignConfig.cs
```diff
@@ -0,0 +1,16 @@
+namespace OutOfSchool.Encryption.Config;
+
+public class EUSignConfig
+{
+    public static readonly string Name = "EUSign";
```
h4wk13: 1. const ?
2. what the `Name` actually means here ?

OutOfSchool/OutOfSchool.Encryption/Config/Proxy.cs
```diff
@@ -0,0 +1,14 @@
+namespace OutOfSchool.Encryption.Config;
+
+public class Proxy
+{
+    public bool Use { get; set; }
```
h4wk13: `Enabled` ?

OutOfSchool/OutOfSchool.Encryption/Config/EUSignConfig.cs
```diff
@@ -0,0 +1,16 @@
...
+
+    public string DefaultOCSPServer { get; set; }
+
+    public string DefaultTSPServer { get; set; }
+
+    public PrivateKey PrivateKey { get; set; }
+
+    public CA CA { get; set; }
+
+    public Proxy Proxy { get; set; }
```
h4wk13, line 25: This setting here means that the system could have more than one proxy setting, is it the case ?

OutOfSchool/OutOfSchool.Encryption/Config/Proxy.cs
```diff
@@ -0,0 +1,14 @@
...
+
+    public string Address { get; set; }
```
h4wk13: `Host` ?
And I haven't seen the validation if that host is valid.
 could you please add that somewhere where we load the configuration ?

OutOfSchool/OutOfSchool.Encryption/Handlers/AppHandlers.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        app.MapGet(
+                "api/{version:apiVersion}/certificate",
+                ([FromServices] EUSignOAuth2Service euSignOAuth2Service) =>
+                {
+                    var cert = euSignOAuth2Service.GetEnvelopCertificateBase64();
```
h4wk13: GetEnvelop**e**CertificateBase64

OutOfSchool/OutOfSchool.Encryption/Handlers/AppHandlers.cs
```diff
@@ -0,0 +1,55 @@
...
+                    if (cert != null)
+                    {
+                        return Results.Ok(cert);
+                    }
+
+                    return Results.Problem(
+                        title: "Certificate not found",
+                        detail: "Certificate was not present in the system.",
+                        statusCode: 500);
```
h4wk13, line 28: `404` ?

OutOfSchool/OutOfSchool.Encryption/Handlers/AppHandlers.cs
```diff
@@ -0,0 +1,55 @@
...
+                })
+            .WithApiVersionSet(apiVersionSet)
+            .MapToApiVersion(1);
```
h4wk13: please move this and all other `1` (api version) to settings or at least to const

OutOfSchool/OutOfSchool.Encryption/Handlers/AppHandlers.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        app.MapPost(
+                "api/{version:apiVersion}/decryption", (
```
h4wk13: `decryption` > `decrypt` ?

OutOfSchool/OutOfSchool.Encryption/Handlers/AppHandlers.cs
```diff
@@ -0,0 +1,55 @@
...
+                    EnvelopedUserInfoResponse encryptedUserInfo,
```
h4wk13: [NotNull] ?

OutOfSchool/OutOfSchool.Encryption/Models/CASettings.cs
```diff
@@ -0,0 +1,20 @@
+namespace OutOfSchool.Encryption.Models;
+
+public class CASettings
+{
+    public string[] issuerCNs { get; set; }
+
+    public string address { get; set; }
+
+    public string ocspAccessPointAddress { get; set; }
```
h4wk13: `Address` > `Host` ?

OutOfSchool/OutOfSchool.Encryption/Models/EnvelopedUserInfoResponse.cs
```diff
@@ -0,0 +1,10 @@
+namespace OutOfSchool.Encryption.Models;
+
+public class EnvelopedUserInfoResponse
+{
+    public string EncryptedUserInfo { get; set; }
```
h4wk13, line 5: should the model be here instead of string ?

OutOfSchool/OutOfSchool.Encryption/Models/UserInfoResponse.cs
```diff
@@ -0,0 +1,42 @@
+namespace OutOfSchool.Encryption.Models;
+
+public class UserInfoResponse
```
h4wk13, line 3: Pascal case for property names ?

OutOfSchool/OutOfSchool.Encryption/OutOfSchool.Encryption.csproj
```diff
@@ -0,0 +1,61 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Remove="stylecop.json" />
+    <AdditionalFiles Include="stylecop.json">
+      <ExcludeFromSingleFile>true</ExcludeFromSingleFile>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+    </AdditionalFiles>
+    <None Remove="appsettings.jsonc" />
```
h4wk13, line 33: `json` ?

OutOfSchool/OutOfSchool.Encryption/Program.cs
```diff
@@ -0,0 +1,73 @@
...
+using OutOfSchool.Encryption.Services;
+using Serilog;
+using Serilog.Context;
+using Serilog.Exceptions;
+using Serilog.Exceptions.Core;
+
+var builder = WebApplication.CreateBuilder(args);
+
+builder.Configuration
+    .AddJsonFile("appsettings.jsonc", false, true)
```
h4wk13, line 15: `json` ?

OutOfSchool/OutOfSchool.Encryption/Program.cs
```diff
@@ -0,0 +1,73 @@
...
+    builder.Services.AddSingleton<IEUSignOAuth2Service, DevEUSignOAuth2Service>();
+}
+else
+{
+    builder.Services.AddSingleton<IEUSignOAuth2Service, EUSignOAuth2Service>();
+}
+
+builder.Services.AddApiVersioning(options =>
+    {
+        options.DefaultApiVersion = new ApiVersion(1);
```
h4wk13: I've seen this `1` in many places.
please move this `1` to config
if this is default value - please move to the const

OutOfSchool/OutOfSchool.Encryption/Program.cs
```diff
@@ -0,0 +1,73 @@
...
+    })
+    .AddApiExplorer(options =>
+    {
+        options.GroupNameFormat = "'v'V";
+        options.SubstituteApiVersionInUrl = true;
+    });
+
+var app = builder.Build();
+
+app.UseSerilogRequestLogging();
```
h4wk13, line 59: Does that logger hide the sensitive data like certs or keys or... ?

OutOfSchool/OutOfSchool.Encryption/Services/DevEUSignOAuth2Service.cs
```diff
@@ -0,0 +1,19 @@
...
+using OutOfSchool.Encryption.Models;
+
+namespace OutOfSchool.Encryption.Services;
+
+public class DevEUSignOAuth2Service : IEUSignOAuth2Service
+{
+    /// <inheritdoc/>
+    public CertificateResponse GetEnvelopCertificateBase64() => new()
+    {
+        CertBase64 = Convert.ToBase64String("very_secret_certificate"u8.ToArray()),
```
h4wk13, line 13: should this name be configurable ?

OutOfSchool/OutOfSchool.Encryption/Services/DevEUSignOAuth2Service.cs
```diff
@@ -0,0 +1,19 @@
...
+    };
+
+    /// <inheritdoc/>
+    public UserInfoResponse DecryptUserInfo(EnvelopedUserInfoResponse encryptedUserInfo) =>
```
h4wk13: Since this is the public method could you please:
1. add the `[NotNull]` attribute to the argument
2. validate it is not null.

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Exception.cs
```diff
@@ -0,0 +1,21 @@
+namespace OutOfSchool.Encryption.Services;
+
+public class EUSignOAuth2Exception : Exception
+{
+    public EUSignOAuth2Exception()
```
h4wk13: ?

OutOfSchool/OutOfSchool.Encryption/Services/IEUSignOAuth2Service.cs
```diff
@@ -0,0 +1,24 @@
+using OutOfSchool.Encryption.Models;
+
+namespace OutOfSchool.Encryption.Services;
+
+public interface IEUSignOAuth2Service
```
h4wk13, line 9: I suppose we can add the documentation for interface as well )
And: we might use the `/// <inheritdoc/>` in all the implementations for each method so they will be documented.

OutOfSchool/OutOfSchool.Encryption/stylecop.json
```diff
@@ -0,0 +1,14 @@
+{
+  // ACTION REQUIRED: This file was automatically added to your project, but it
```
h4wk13: ?

OutOfSchool/OutOfSchool.Encryption/Services/DevEUSignOAuth2Service.cs
VadymLevkovskyi, line 13: (Security) `very_secret_certificate` )))

OutOfSchool/OutOfSchool.Encryption/Config/Proxy.cs
h4wk13: And another question:
I haven't seen the validation anywhere... 
Could you please add the check whether the port is valid ? (>=0 && < ushort.MaxValue ) ?

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+        try
+        {
+            IEUSignCP.CtxDevelopData(
+                pkContext,
+                encryptedUserInfo.EncryptedUserInfo,
+                null,
+                out var developedUserInfo,
+                out _);
+
+            Stream userInfoStream = new MemoryStream(developedUserInfo);
```
VadymLevkovskyi: (Resource deallocation) Please use `using` here

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+            logger.LogError(ex, "An error occurred while decrypting user info");
+            return null;
+        }
+    }
+
+    /// <summary>
+    /// Пошук зареєстрованого носія ключової інформації
+    /// This code is from IIT Library usage example with minimal code styling changes.
+    /// </summary>
+    private static void GetKeyMedia(
```
VadymLevkovskyi, line 122: (Codestyle) This whole method is written using non-C# code style

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+    {
+        keyMedia = new IEUSignCP.EU_KEY_MEDIA(0, 0, string.Empty);
+
+        try
+        {
+            int typeIndex, deviceIndex;
+            string curType, curDevice;
+
+            typeIndex = 0;
+            while (true)
```
VadymLevkovskyi: That's very brave idea to not limit `typeIndex` and `deviceIndex` in these `while(true)` cycles

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+
+        // Встановлення параметрів OCSP-серверу ЦСК за замовчанням
+        IEUSignCP.SetOCSPSettings(
+            true, true, eUSignConfig.DefaultOCSPServer, "80");
+
+        // Встановлення налаштувань точок доступу до OCSP-серверів
+        // Необхідні при обслуговуванні користувачів з різних ЦСК
+        IEUSignCP.SetOCSPAccessInfoModeSettings(true);
+
+        var fsCAs = new FileStream(eUSignConfig.CA.JsonPath, FileMode.OpenOrCreate);
```
VadymLevkovskyi: (Resources deallocation) `Using`

OutOfSchool/OutOfSchool.Encryption/appsettings.jsonc
```diff
@@ -0,0 +1,96 @@
...
+      "JKSAlias": "",
+      // The name of the private key medium type
+      // (the full list of private key medium types is returned by the EnumKeyMediaTypes function)
+      "MediaType": "криптомод. ІІТ Гряда-301",
+      // The name of the private key medium
+      // (a complete list of private key media for a given type is returned by the EnumKeyMediaDevices function)
+      "MediaDevice": "001:3011 (192.168.0.100)",
+      // Password of the server's private key.
+      // If Гряда-301 is used as the medium, the username and password are set as ##Login##Password
+      "Password": "12345678",
```
VadymLevkovskyi: (Security) Though it looks like a test password it's not a good idea to have it in config

OutOfSchool/OutOfSchool.Encryption/Config/EUSignConfig.cs
VadymLevkovskyi: I believe it's config section name. (Used while taking config by deserializing json/ini/other text files)

OutOfSchool/OutOfSchool.Encryption/Handlers/AppHandlers.cs
VadymLevkovskyi: I doubt - it looks more like internal server error (deployment was broken) than absence of customer's entity.

OutOfSchool/OutOfSchool.Encryption/Config/EUSignConfig.cs
h4wk13: If so, could we please rename it to `SectionName` or `ConfigSeciontName` ?

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+            {
+                Initialize();
+            }
+            catch (Exception ex)
+            {
+                // No logging, as this will lead to server crash and log at that point
+                throw new EUSignOAuth2Exception("An error occurred while initializing the cryptographic library.", ex);
+            }
+
+            ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
```
h4wk13: I believe it is the default setting for current .NET version.
Should we also add 1.3 version ?

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+                // No logging, as this will lead to server crash and log at that point
+                throw new EUSignOAuth2Exception("An error occurred while reading the server's private key.", ex);
+            }
+        }
+    }
+
+    /// <inheritdoc/>
+    public CertificateResponse GetEnvelopCertificateBase64()
+    {
+        if (pkEnvelopCert == null)
```
h4wk13: pkEnvelop**e**Cert

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+        {
+            logger.LogError("No envelop certificate was loaded");
```
h4wk13, line 81: envelop**e**

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+            catch (Exception ex)
+            {
+                // No logging, as this will lead to server crash and log at that point
+                throw new EUSignOAuth2Exception("An error occurred while reading the server's private key.", ex);
+            }
+        }
+    }
+
+    /// <inheritdoc/>
+    public CertificateResponse GetEnvelopCertificateBase64()
```
h4wk13: Do we really need to return the cert instead of encrypt something ?

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+        }
+
+        return new()
+        {
+            CertBase64 = Convert.ToBase64String(pkEnvelopCert),
+        };
+    }
+
+    /// <inheritdoc/>
+    public UserInfoResponse DecryptUserInfo(EnvelopedUserInfoResponse encryptedUserInfo)
```
h4wk13: [NotNull]

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+
+        // Ініціалізація бібліотеки
+        IEUSignCP.Initialize();
+
+        // Вимкнення збереження налаштувань до реєстру
+        IEUSignCP.SetRuntimeParameter(
+            IEUSignCP.EU_SAVE_SETTINGS_PARAMETER, IEUSignCP.EU_SETTINGS_ID_NONE);
+
+        // Вимкнення відображення помилок
+        IEUSignCP.SetUIMode(false);
```
h4wk13, line 191: it was called few lines above

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+        string[] certsFiles,
+        string caIssuerCN)
+    {
+        byte[] pKey = null;
+        List<byte[]> certs = new List<byte[]>();
+        bool fromFile = !string.IsNullOrEmpty(pkFile);
+        IEUSignCP.EU_KEY_MEDIA keyMedia = new IEUSignCP.EU_KEY_MEDIA(0, 0, string.Empty);
+        CASettings ca = null;
+        IEUSignCP.EU_CERT_OWNER_INFO info;
+        IEUSignCP.EU_CERT_INFO_EX infoEx;
```
h4wk13: could you please review these variables and move declarations to the places where they really used ?

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+            if (certsFiles != null && certsFiles.Length > 0)
+            {
+                foreach (var file in certsFiles)
+                {
+                    if (!File.Exists(file))
+                    {
+                        throw new Exception($"The certificate file is missing: {file}");
+                    }
+
+                    certs.Add(File.ReadAllBytes(file));
```
h4wk13: How many of certs we can load into memory on the server without experiencing memory issues ?

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+                if (!File.Exists(pkFile))
+                {
+                    throw new Exception($"The private key file is missing: {pkFile}");
+                }
+
+                pKey = File.ReadAllBytes(pkFile);
+                if (!string.IsNullOrEmpty(jksAlias))
+                {
+                    // Пошук відповідного ос. ключа в контейнері JKS
+                    IEUSignCP.GetJKSPrivateKey(pKey, jksAlias, out pKey, out var certificates);
```
h4wk13, line 350: it seems that we're going to overwrite `pKey` variable initialized in the 340 line. Is it OK ? if yes - why ?

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+    // ReSharper disable once InconsistentNaming
+    [SuppressMessage(
+        "StyleCop.CSharp.NamingRules",
+        "SA1306:Field names should begin with lower-case letter",
+        Justification = "CA is an acronym")]
+    private CASettings[] CAs = null;
+    private IntPtr context = IntPtr.Zero;
+    private IntPtr pkContext = IntPtr.Zero;
+    private byte[] pkSignCert = null;
+    private byte[] pkEnvelopCert = null;
```
h4wk13: could you please add `readonly` modifier too all of these fields that are initialized in scope or constructor ?

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+        }
+
+        // Встановлення параметрів TSP-серверу ЦСК
+        // Параметри використовуються при накладанні підпису для додавання мітки часу
+        // Порядок використання параметрів TSP: 
+        // - з точки доступу до TSP, сертифіката підписувача;
+        // - з параметрів TSP, що встановлені за замовчанням.
+
+        // Встановлення параметрів TSP-серверу ЦСК за замовчанням
+        IEUSignCP.SetTSPSettings(true, eUSignConfig.DefaultTSPServer, "80");
```
h4wk13: can we read that and all other `"80"` from config ?

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+                    }
+
+                    certs.Add(File.ReadAllBytes(file));
+                }
+            }
+
+            if (fromFile)
+            {
+                // Зчитування ос. ключа з файлу до байтового масиву
+                if (!File.Exists(pkFile))
```
h4wk13: I suppose we would like to have unit tests for that service, if so - could you please move all System.IO related calls into dedicated helper or service or... and inject it as the constructor argument ?

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+                {
+                    throw new Exception($"The private key file is missing: {pkFile}");
+                }
+
+                pKey = File.ReadAllBytes(pkFile);
+                if (!string.IsNullOrEmpty(jksAlias))
+                {
+                    // Пошук відповідного ос. ключа в контейнері JKS
+                    IEUSignCP.GetJKSPrivateKey(pKey, jksAlias, out pKey, out var certificates);
+                    foreach (var certificate in certificates)
```
h4wk13: please replace it with certs.AddRange()

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,480 @@
...
+                IEUSignCP.CtxGetOwnCertificate(
+                    pkContext,
+                    IEUSignCP.EU_CERT_KEY_TYPE_DSTU4145,
+                    IEUSignCP.EU_KEY_USAGE_KEY_AGREEMENT,
+                    out infoEx,
+                    out pkEnvelopCert);
+            }
+            catch (Exception)
+            {
+                throw new Exception("The private key does not support PKI encryption described in ДСТУ 4145");
```
h4wk13, line 467: We've lost the original exception here, could we log it with the error ? 
If it is taken as is from the example - it might be the risk to log due to sensitive data exposure...

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
h4wk13: You've probably meant that: if device name from the config will be wrong - we will have to wait for some noticeable time while got the very descriptive error.

OutOfSchool/OutOfSchool.Encryption/Config/Proxy.cs
DmyMi: @VadymLevkovskyi @h4wk13 
Port is `string` because the `C` library methods have ports as `string` also. I can add validation, but is it necessary to do the `string -> int -> string` convertion?

OutOfSchool/OutOfSchool.Encryption/Models/CASettings.cs
DmyMi: @VadymLevkovskyi same as above - `C` library uses it as string.

OutOfSchool/OutOfSchool.Encryption/Models/EnvelopedUserInfoResponse.cs
DmyMi: @VadymLevkovskyi 
model is copied from 3rd party authorization API and they do return `200 OK` & error code in body. Error code is their business logic code, not exactly HTTP code. We can check for error before calling this service, will remove this here and add check later in AuthServer.

OutOfSchool/OutOfSchool.Encryption/Config/EUSignConfig.cs
DmyMi: @h4wk13 the configuration is parsed from `appsettings.json` and configured & injected as `IOptions` and the values are cached for the application's lifetime. Unless we implement a lot of custom logic or configure and use `IOptionsMonitor` - the only way to change these settings is to change appsettigns file or add environment variables and completely restart the application.

OutOfSchool/OutOfSchool.Encryption/Services/DevEUSignOAuth2Service.cs
DmyMi: @h4wk13 dev mock authorization server has also this hardcoded as it does not implement any encryption.

OutOfSchool/OutOfSchool.Encryption/Handlers/AppHandlers.cs
DmyMi: @h4wk13 the meaning i put here is that the certificate was not present on the server or the `C` library failed to parse it. So like Vadym said - something is broken on backend.

maybe wrong wording.

OutOfSchool/OutOfSchool.Encryption/Handlers/AppHandlers.cs
DmyMi: According to the [docs](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullattribute?view=net-8.0) it is for static analysis only. For service i agree we can add it because we control how the method is called, but for "Controller" method we have no control over how the method will be called in runtime.

OutOfSchool/OutOfSchool.Encryption/Models/CASettings.cs
DmyMi: @h4wk13  this is parsed from 3rd party JSON. Can't rename it

OutOfSchool/OutOfSchool.Encryption/Models/EnvelopedUserInfoResponse.cs
DmyMi: @h4wk13 the 3rd party API return an encrypted JSON string as a field value. We need to decrypt it and then deserialize the JSON (we can actualy skip deserialization in this service becuase we just serialize it back, but i decided to have it so we can atleast check if the JSON can bind to properties early)

OutOfSchool/OutOfSchool.Encryption/Program.cs
DmyMi: @h4wk13 though `json` can have comments that are parsed by .Net, Visual Studio and Rider highlight it as errors. The `jsonc` (JSON with comments) is accepted by VS/Rider though needs some manual steps to add to project.

OutOfSchool/OutOfSchool.Encryption/Program.cs
DmyMi: `UseSerilogRequestLogging()` never logs body from requests, only URLs

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
DmyMi: @h4wk13 the certificate is used by Authorization server to send UserInfo request during the OAuth2 flow. So we call this service, get the certificate and send it to 3rd party OAuth provider so it encrypts data.

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
DmyMi: @h4wk13 it is initialized in constructor, but as an `out` property

> Field 'pkEnvelopeCert' is read-only (except in constructor or init-only setter of the type in which the field is defined). 'out' argument must be an assignable variable, field, or an array element

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
DmyMi: @h4wk13 according to their source code the exception text is generated by `C` code. No indication if it is sensitive or not.

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
DmyMi: @h4wk13 it is done in the examples in all languages. Probably some logic between those calls resets this value :)

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
DmyMi: @h4wk13  according to documentation there can be 1-2 certificate files, as we are going to host only 1 organization

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
DmyMi: @h4wk13 according to docs certain certificate authorities have certificates as a container for set of different keys instead of being plain certificate. So we read the certificate and if it is a container (we specify the alias to search) - we use it and the key alias to find the actual certificate in the container

OutOfSchool/OutOfSchool.Encryption/Config/Proxy.cs
VadymLevkovskyi: Types are part of self-documented code. In other words, I don't expect `":8080"` value in `int` variable. Thus I'm ready to pay this "double conversion" price. 
Annotation you've used here is good as well for the mentioned purpose (though I more trust to compiler than some reflection).

OutOfSchool/OutOfSchool.Encryption/Models/UserInfoResponse.cs
DmyMi: fixed

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
VadymLevkovskyi: Ok, seems this refactoring could be postponed

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
VadymLevkovskyi: > By default, integral values (such as int, uint, and long) do not throw an exception when they overflow. Instead, they "wrap" -- and this is probably not a value that we want. But we can change this behavior so that an exception is thrown instead. [Link](
https://jeremybytes.blogspot.com/2023/01/checking-for-overflow-in-c.html#:~:text=January%2017%2C%202023-,Checking%20for%20Overflow%20in%20C%23,an%20exception%20is%20thrown%20instead.)

"Noticable" time could last forever.

OutOfSchool/OutOfSchool.Encryption/Services/DevEUSignOAuth2Service.cs
VadymLevkovskyi: ok, let it be, though it could be at least some constant with meaningful name

OutOfSchool/OutOfSchool.Encryption/Config/EUSignConfig.cs
```diff
@@ -0,0 +1,24 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace OutOfSchool.Encryption.Config;
+
+public class EUSignConfig
+{
+    public static readonly string ConfigSectionName = "EUSign";
+
+    public string DefaultOCSPServer { get; set; }
```
h4wk13, line 10: could/should we add `[Required]` here and for other ref types ?

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,485 @@
...
+        return context != IntPtr.Zero;
+    }
+
+    /// <summary>
+    /// This code is from IIT Library usage example with minimal code styling changes.
+    /// </summary>
+    private void Initialize()
+    {
+        // Увімкнення кидання винятків бібліотекою, замість повертання помилок.
+        IEUSignCP.SetThrowExceptions(true);
```
h4wk13, line 178: should these flags be moved to config ?

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
h4wk13: or it is a mistake in example, which is highly possible I suppose )

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,485 @@
...
+        IEUSignCP.SetUIMode(false);
+
+        // Встановлення параметрів взаємодії з сервісами TSP-, OCSP-, CMP- ЦСК
+        // Використання режиму он-лайн
+        IEUSignCP.SetModeSettings(false);
+
+        // Встановлення налаштувань Сховища сертифікатів та СВС
+        IEUSignCP.SetFileStoreSettings(
+            string.Empty,
+            false,
```
h4wk13: configuration ?
I would ask you to move all these `1,2,3, true, false` and other literals to the config or to the constants, if you don't mind.

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
h4wk13: Sounds promising...

OutOfSchool/OutOfSchool.Encryption/Services/EUSignOAuth2Service.cs
```diff
@@ -0,0 +1,485 @@
...
+                        IEUSignCP.GetKeyInfo(keyMedia, out keyInfo);
+                    }
+
+                    string[] cmpServers = { ca.CmpAddress };
+                    string[] cmpServersPorts = { AppConstants.DefaultPort.ToString() };
+                    IEUSignCP.GetCertificatesByKeyInfo(
+                        keyInfo, cmpServers, cmpServersPorts, out certsCMP);
+
+                    int i = 0;
+                    while (true)
```
h4wk13: should it also be changed to `i < int.MaxValue ?` or to `for`

h4wk13 (approved):

PR #1584, opened 2024-10-24
Kyryk/Implemented healthcheck endpoint 

Tetyana-K:
Implemented the ServiceController for handling health check requests.
Added unit tests to verify the functionality of the health check endpoint.

Work in progress.
Thank you to all reviewers for your time and feedback)


VadymLevkovskyi:
@Tetyana-K could you explain why we can't use [standard approach for health checks from Asp.Net](https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-8.0) ? 
Your code seems to be valid, but it's bit "excessive".

Tetyana-K:
Thank you for the comment, I agree, corrected it.

DmyMi (approved):

PR #1586, opened 2024-10-29
Dolishnii/social group error

### LINKED ISSUE #1585: Dolishnii/SocialGroupError

AndrD1406:
Supposedly have fixed the SocialGroup error (pls make sure that I actually have).

### DIFF
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
@@ -32,6 +32,7 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
     private readonly string includingPropertiesForMappingWorkShopCard = $"{nameof(Workshop.Address)}";
 
     private readonly IWorkshopRepository workshopRepository;
+    private readonly IEntityRepository<long, Tag> tagRepository;
     private readonly IEntityRepositorySoftDeleted<long, DateTimeRange> dateTimeRangeRepository;
     private readonly IEntityRepositorySoftDeleted<Guid, ChatRoomWorkshop> roomRepository;
     private readonly ITeacherService teacherService;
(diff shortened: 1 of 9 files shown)


AndrD1406:
@VadymLevkovskyi requesting a review

OutOfSchool/OutOfSchool.BusinessLogic/Services/ISocialGroupService.cs
```diff
@@ -28,7 +28,7 @@ public interface ISocialGroupService
     /// </summary>
     /// <param name="dto">SocialGroup entity to add.</param>
     /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
-    Task<SocialGroupDto> Create(SocialGroupCreate dto);
+    Task<SocialGroupCreate> Create(SocialGroupCreate dto);
```
DmyMi: `SocialGroupCreate` should not be a response type of a service, as anything that is used to create - is an argument of a method.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -119,17 +123,7 @@ public async Task<WorkshopDto> Create(WorkshopCreateUpdateDto dto)
...
-        foreach (var tagId in dto.TagIds)
-        {
-            var tag = await tagService.GetById(tagId);
-            if (tag != null)
-            {
-                var tagDto = mapper.Map<TagDto>(tag);
-                tags.Add(tagDto);
-            }
-        }
+        workshop.Tags = (await tagRepository.GetAll()).Where(tag => dto.TagIds.Contains(tag.Id)).ToList();
```
DmyMi: i do like this one query approach.

But you use both `ITagService` and `IEntityRepository<long, Tag>`. It will take the same amount of time to create this method in the service and use it instead.

OutOfSchool/OutOfSchool.BusinessLogic/Services/SocialGroupService.cs
```diff
@@ -81,8 +94,7 @@ public async Task<SocialGroupDto> Create(SocialGroupCreate dto)
 
         logger.LogInformation($"SocialGroup with Id = {newSocialGroup?.Id} created successfully.");
 
-        return mapper.Map<SocialGroupDto>(socialGroup, opt =>
-        opt.Items["Localization"] = LocalizationType.Ua);
+        return mapper.Map<SocialGroupCreate>(newSocialGroup);
```
DmyMi: You can use the custom code approach to create `SocialGroupDto` manually like you did before. Please don't return `...Create` types :)

DmyMi (changes requested):

DmyMi (approved):

PR #1596, opened 2024-11-06
Add more generic error handling to communication service

DmyMi:


OutOfSchool/OutOfSchool.WebApi.Tests/Common/CommunicationServiceTest.cs
```diff
@@ -71,10 +73,7 @@ public async Task SendRequest_WithCorrectRequest_ReturnsOkResponse()
         // Act
         var result = await communicationService.SendRequest<TestResponse, ErrorResponse>(request);
 
-        result.DoRight(r =>
-        {
-            Assert.AreEqual("OK", r.Content);
-        });
+        result.DoRight(r => { Assert.AreEqual("OK", r.Content); });
```
VadymLevkovskyi: (Just a suggestion) You could introduce extensions for assertion framework that works with result, especially specific result types. In my project I have something similar to (as I'm using `FluentAssertion` with `Should()` my code abstractions are different, but idea is similar):
`result.AssertRight(predicate)` - which runs `predicate` for result with `right` option, but `Assert.Fails()` for `left` option
`result.AssertOkResponse()` - used for controller actions testing, where result is an `IActionResult` of specific type (`OkResult`) 
`result.AssertOkJsonResponse(predicate)` - used for controller actions testing, where result is an `IActionResult` of specific type (`JsonResult`), additionally checks the json object
...and so on

OutOfSchool/OutOfSchool.WebApi.Tests/Common/CommunicationServiceTest.cs
```diff
@@ -165,4 +210,24 @@ private class TestResponse : IResponse
     {
         public string? Content { get; set; }
     }
+
+    private class TestError : IErrorResponse
```
VadymLevkovskyi: (Code style) Why `class` but not `record`? See the beauty:
`private record TestError(HttpStatusCode HttpStatusCode, string Message, string Content) : IErrorResponse {}`

PR #1604, opened 2024-12-03
Storing workshop to cache during creation

HlibBondarev:
This PR is a replacement for the previous PR (https://github.com/ita-social-projects/OoS-Backend/pull/1551) with corrections to all comments received.

Roles: provider, ministryadmin, areaadmin, regionadmin, techadmin.

Story: As a User with the above roles, I want to be able to save the field values before saving the Workshop (Provider, etc.) in the DB. So that if the  Workshop (Provider, etc.) creation is terminated, I can restore the values of previously filled fields.

Solution: Created DraftStorageService. It captures and exports the state of an object (entity) in the storage. This storage has to support IReadWriteCacheService interface. Redis is used as a storage.

1) Created IReadWriteCacheService interface with methods - ReadAsync(), WriteAsync() and RemoveAsync().
(* If I added these methods in the existing ICacheService interface I would have to implement these methods not only in CacheService class but also in MultiLayerCache. The matter of fact that MultiLayerCache will be rewrite in future. So I decided to create a special interface - IReadWriteCacheService.)
2) Implemented IReadWriteCacheService interface in CacheService class.
3) Refactored  CacheService class: 
- used ReadAsync() and WriteAsync()  methods in GetOrAddAsync() method.
- fixed RemoveAsync() method - instead of cache.RemoveAsync(key) used cache.Remove(key).
4) Created IDraftStorageService interface and implemented it in DraftStorageService class .
5) Created abstract generic controller DraftStorageController with actions StoreDraft, RestoreDraft and RemoveDraft. 
(* This abstract class is used like base class for controllers which are responsible for storing Workshops, Providers, etc. ) 
6) Created WorkshopDraftStorageController inherited from DraftStorageController with T = WorkshopMainRequiredPropertiesDto.
7) Added tests to CacheServiceTests class for methods - ReadAsync(), WriteAsync() and RemoveAsync().
8) Created DraftStorageServiceTests class with NUnit tests for DraftStorageService class methods - RestoreAsync(), CreateAsync() and RemoveAsync().


VadymLevkovskyi (approved):
In normal development I'd insist to split this PR in 3 separate PRs:
1. .Net8 related changes in .csproj files and (as I get) explicit nullable types specified
2. Refactoring of one big Workshop Dto into smaller parts
3. Draft storage introduction

Thanks for tests presence. Thus, I'm approving this PR, though not sure if concurrency logic is correct (but it could be solved later in bugs, in case)

PR #1607, opened 2024-12-09
fix employee check for applications

DmyMi:


OutOfSchool/OutOfSchool.BusinessLogic/Services/CurrentUserService.cs
```diff
@@ -196,7 +196,7 @@ private async Task<bool> ParentHasRights(Guid parentId, Guid childId)
 
     private async Task<bool> ProviderHasRights(Guid providerId)
     {
-        if (!IsInRole(Role.Provider) || this.IsEmployeeOrProvider())
+        if (!IsInRole(Role.Provider) || this.IsEmployee())
```
VadymLevkovskyi, line 199: (Excessive logic) Please check this condition. I believe ou don't need to check if user is an employee

OutOfSchool/OutOfSchool.BusinessLogic/Services/ApplicationService.cs
```diff
@@ -252,7 +252,7 @@ public async Task<SearchResult<ApplicationDto>> GetAllByParent(Guid id, Applicat
     public async Task<int> GetCountByParentId(Guid id)
     {
         logger.LogInformation("Getting Applications count by Parent Id started. Looking Parent Id = {Id}", id);
-        if (!currentUserService.IsInRole(Role.Provider) && !currentUserService.IsEmployeeOrProvider())
+        if (!currentUserService.IsInRole(Role.Provider) && !currentUserService.IsEmployee())
```
VadymLevkovskyi, line 255: (Multi-level logic) You have here 2 levels of abstraction - more basic one `.IsInRole(someRole)` and more complex `.IsEmployee()`. Please use only one/ As for me the better choice is to make `IsInRole` private.

OutOfSchool/OutOfSchool.BusinessLogic/Services/CurrentUserService.cs
```diff
@@ -228,7 +228,7 @@ private async Task<bool> ProviderHasRights(Guid providerId)
 
     private async Task<bool> ProviderAdminHasRights(string providerAdminId)
     {
-        if (!this.IsEmployeeOrProvider() || UserId != providerAdminId)
+        if (!this.IsEmployee() || UserId != providerAdminId)
```
VadymLevkovskyi, line 231: (Excessive logic) Looks like provider only should be ok. So this check of `! IsEmloyee()` is wrong)

VadymLevkovskyi (commented):
Logic

OutOfSchool/OutOfSchool.BusinessLogic/Services/CurrentUserService.cs
DmyMi: @VadymLevkovskyi We are transitioning to a more generic approach and instead of `ProviderAdmin` we are going to have `Employee` and `Position` to check what an employee can do. For now this is the first step - renaming enities and removing excessive logic tied to `ProviderAdmin` & `ProviderDeputy`. This method right here will be changed soon and all the naming will be streamlined.

PR #1608, opened 2024-12-09
remove unnecessary logic that was used only in one place

DmyMi:


OutOfSchool/OutOfSchool.BusinessLogic/Services/CurrentUserService.cs
```diff
@@ -226,9 +224,10 @@ private async Task<bool> ProviderHasRights(Guid providerId)
         return result;
     }
 
+    // TODO: Need to check which method should stay after ProviderAdmin -> Employee refactoring.
     private async Task<bool> ProviderAdminHasRights(string providerAdminId)
     {
-        if (!this.IsEmployee() || UserId != providerAdminId)
+        if (!this.IsInRole(Role.Employee) || UserId != providerAdminId)
```
VadymLevkovskyi, line 230: Shouldn't it be `!InRole(Provider)`

VadymLevkovskyi: Ah, I've read your comment. "... we are going to have Employee and Position to check what an employee can do"
Now resolving this one

VadymLevkovskyi (approved):

PR #1624, opened 2025-01-04
[Data model] Educational subject

### LINKED ISSUE #1533: [Data model] Educational subject
Unified information about the educational subject, discipline, specialized course, workshop and any other educational topic, regarding which the educational process is carried out.

**Is a business entity as in #1524**

Parent.Name | Name | Data Type | Mandatory | Comment
-- | -- | -- | -- | --
Educational subject | Name in the language of teaching | String | FALSE |
Educational subject | Name in Ukrainian | String | FALSE |
Educational subject | if it is the language, is it the main one | boolean | FALSE |
Educational subject | identifier | int | FALSE |

omaiva:
Updated Educational Subject(StudySubject) entity, its service and controller. Updated Language entity, created service and controller for it. Added tests for services and controllers. #1533


OutOfSchool/OutOfSchool.BusinessLogic/Models/LanguageDto.cs
```diff
@@ -0,0 +1,15 @@
...
+
+    /// <summary>
+    /// ISO code of the language
+    /// </summary>
+    public string Code { get; set; }
+
+    /// <summary>
+    /// Title of the language
+    /// </summary>
+    public string Title { get; set; }
```
DmyMi: For naming `Name` is more suitable than `Title` for a language.

OutOfSchool/OutOfSchool.BusinessLogic/Models/StudySubjects/StudySubjectCreateUpdateDto.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// Language of instruction (allows multiple selection)
+    /// </summary>
+    [Required(ErrorMessage = "The language of instruction is required.")]
+    public List<long> LanguageIds { get; set; }
+
+    /// <summary>
+    /// Primary language of the subject
+    /// </summary>
+    [Required(ErrorMessage = "Primary language's id is required.")]
+    public int PrimaryLanguageId { get; set; }
```
DmyMi: `int` when language id is actualy long?

OutOfSchool/OutOfSchool.BusinessLogic/Models/StudySubjects/StudySubjectDto.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace OutOfSchool.BusinessLogic.Models.StudySubjects;
+public class StudySubjectDto
+{
+    public Guid Id { get; set; }
+    public string NameInUkrainian { get; set; }
+    public string NameInInstructionLanguage { get; set; }
+    public bool IsPrimaryLanguageUkrainian { get; set; }
+    public List<long> LanguageIds { get; set; }
```
DmyMi: This is a "display dto", so it should contain information suitable for users to see. So instead of language Ids we need to return a list of `LanguageDto`.

OutOfSchool/OutOfSchool.BusinessLogic/Models/StudySubjects/StudySubjectCreateUpdateDto.cs
```diff
@@ -0,0 +1,64 @@
...
+            yield return new ValidationResult("PrimaryLanguageId cannot be smaller than one.", new[] { nameof(PrimaryLanguageId) });
+        }
+
+        if (LanguageIds == null || !LanguageIds.Any())
+        {
+            yield return new ValidationResult("LanguageIds cannot be null or empty.", new[] { nameof(LanguageIds) });
+        }
+        else
+        {
+            if (!LanguageIds.Contains(PrimaryLanguageId))
```
DmyMi: to reduce the needed checks for "is this id in the list", we can have a list with custom simple objects having two properties `languageId, isPrimary`, where `isPrimary` is boolean. There's going to be only a simple validation for `isPrimary=true` only for single object.

OutOfSchool/OutOfSchool.BusinessLogic/Models/StudySubjects/StudySubjectDto.cs
```diff
@@ -0,0 +1,15 @@
...
+    public int PrimaryLanguageId { get; set; }
+    public Guid WorkshopId { get; set; }
+    public DateTime CreatedAt { get; set; }
```
DmyMi: `CreatedAt` and `UpdatedAt` is usually considered internal information, unless required we should not return it.

OutOfSchool/OutOfSchool.BusinessLogic/Models/StudySubjects/StudySubjectDto.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace OutOfSchool.BusinessLogic.Models.StudySubjects;
+public class StudySubjectDto
+{
+    public Guid Id { get; set; }
+    public string NameInUkrainian { get; set; }
+    public string NameInInstructionLanguage { get; set; }
+    public bool IsPrimaryLanguageUkrainian { get; set; }
+    public List<long> LanguageIds { get; set; }
+    public int PrimaryLanguageId { get; set; }
+    public Guid WorkshopId { get; set; }
```
DmyMi, line 10: Note: `WorkshopId` can stay as id, because the information about workshop will already be available on client side.

OutOfSchool/OutOfSchool.BusinessLogic/Services/LanguageService.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        this.repository = repository ?? throw new ArgumentNullException(nameof(repository));
+        this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        this.mapper = mapper ?? throw new ArgumentNullException(nameof(mapper));
+    }
+
+    /// <inheritdoc/>
+    public async Task<IEnumerable<LanguageDto>> GetAll()
+    {
+        logger.LogInformation("Getting all languages started.");
```
DmyMi: make it `debug` to reduce logging  _noise_

OutOfSchool/OutOfSchool.BusinessLogic/Services/LanguageService.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        var languages = await repository.GetAll().ConfigureAwait(false);
+
+        logger.LogInformation(!languages.Any()
```
DmyMi: though you might find this kind of logging somewhere in our legacy code, we try to not do it any more :) this is the way (text is compile time constant, no period (`.`) in the end:

```csharp
logger.LogDebug("{Count} records were successfully received from the Languages table, languages.Count())
```

OutOfSchool/OutOfSchool.BusinessLogic/Services/LanguageService.cs
```diff
@@ -0,0 +1,41 @@
...
+            ? "Languages table is empty."
+            : $"All {languages.Count()} records were successfully received from the Languages table.");
+
+        return languages.Select(mapper.Map<LanguageDto>).ToList();
```
DmyMi:
```csharp
mapper.Map<List<TDestination>>(listOfSourceObjects)
```

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -0,0 +1,201 @@
...
+            throw new ArgumentException("Dto is null.", nameof(dto));
+        }
+
+        await CheckIfLanguagesIdsAreCorrect(dto);
+
+        var studySubject = mapper.Map<StudySubject>(dto);
+
+        var newStudySubject = await studySubjectRepository.Create(studySubject).ConfigureAwait(false);
+
+        logger.LogDebug($"StudySubject with Id = {newStudySubject?.Id} created successfully.");
```
DmyMi: same comment for compile time constant, period, etc.

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -0,0 +1,201 @@
...
+                whereExpression: predicate
+            ).AsNoTracking()
+            .ToListAsync()
+            .ConfigureAwait(false);
+
+        logger.LogInformation(!subjects.Any()
+            ? "StudySubject table is empty."
+            : $"All {subjects.Count()} records were successfully received from the StudySubject table");
+
+        return subjects.Select(mapper.Map<StudySubjectDto>).ToList();
```
DmyMi:
```csharp
mapper.Map<List<TDestination>>(listOfSourceObjects)
```

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        var studySubject = mapper.Map<StudySubject>(dto);
+
+        var newStudySubject = await studySubjectRepository.Create(studySubject).ConfigureAwait(false);
+
+        logger.LogDebug($"StudySubject with Id = {newStudySubject?.Id} created successfully.");
+
+        logger.LogInformation("StudySubjectLanguages table updated succesfully.");
+
+        return mapper.Map<StudySubjectCreateUpdateDto>(newStudySubject);
```
DmyMi: `StudySubjectCreateUpdateDto` is only for adding into the system, it should not be returned. You need to map to "display dto" here

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -0,0 +1,201 @@
...
+        if (entity is null || entity.IsDeleted)
+        {
+            logger.LogWarning($"StudySubject with Id = {id} was not found.");
+            throw new KeyNotFoundException($"StudySubject with Id = {id} does not exist or it was deleted.");
+        }
+
+        try
+        {
+            await studySubjectRepository.Delete(entity).ConfigureAwait(false);
+            logger.LogInformation($"StudySubject with Id = {id} successfully deleted.");
```
DmyMi: the same logging comment :)

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -0,0 +1,201 @@
...
+    public async Task Delete(Guid id)
+    {
+        logger.LogInformation($"Deleting StudySubject with Id = {id} started.");
+
+        var entity = await studySubjectRepository.GetById(id).ConfigureAwait(false);
+
+        if (entity is null || entity.IsDeleted)
+        {
+            logger.LogWarning($"StudySubject with Id = {id} was not found.");
+            throw new KeyNotFoundException($"StudySubject with Id = {id} does not exist or it was deleted.");
```
DmyMi: when you delete an entity that was already deleted - it is not an error. You can return `204` or `404` status code here. Also don't use exceptions as part of the control flow :)

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -0,0 +1,201 @@
...
+                || s.NameInInstructionLanguage.Contains(filter.SearchString));
+        }
+
+        predicate = predicate.And(s => !s.IsDeleted);
+
+        var subjects = await studySubjectRepository
+            .Get(
+                skip: filter.From,
+                take: filter.Size,
+                whereExpression: predicate
```
DmyMi, line 129: You return a list, that contains a list of language ids, then when you map each study subject - lazy loading will make 100500 extra queries to load languages. Use explicit include to load everything in 1 sql request.

Also check any other get by filter method, e.g., in Anna's PR. Need to have additional logic added to count how many other entities with the same filter exist, etc.

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -0,0 +1,201 @@
...
+            ).AsNoTracking()
+            .ToListAsync()
+            .ConfigureAwait(false);
+
+        logger.LogInformation(!subjects.Any()
```
DmyMi: same logging :)

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -0,0 +1,201 @@
...
+    public async Task<StudySubjectDto> GetById(Guid id)
+    {
+        logger.LogInformation($"Getting StudySubject by Id started. Looking Id = {id}.");
+
+        var studySubject = await studySubjectRepository.GetById(id)
+            .ConfigureAwait(false);
+
+        if (studySubject == null || studySubject.IsDeleted)
+        {
+            throw new KeyNotFoundException($"There are no recors in StudySubjects table with such id - {id}, or such StudySubject was deleted.");
```
DmyMi: don't throw, return null and just return `204` or `404` in controller. Also technically, the `isDeleted` should be handled by softdelete repository (if i remember correctly). So only null check is enough

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -0,0 +1,201 @@
...
+            throw new ArgumentException("Dto is null.", nameof(dto));
+        }
+
+        await CheckIfLanguagesIdsAreCorrect(dto);
+
+        var studySubject = await studySubjectRepository.GetById(dto.Id).ConfigureAwait(false);
+
+        if (studySubject == null || studySubject.IsDeleted)
+        {
+            throw new KeyNotFoundException($"There are no recors in StudySubjects table with such id - {dto.Id}, or such StudySubject was deleted.");
```
DmyMi: same deleted comment as before, also don't throw. This is not an exception :)

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        try
+        {
+            var updatedStudySubject = await studySubjectRepository.Update(studySubject)
+                .ConfigureAwait(false);
+            await studySubjectRepository.SaveChangesAsync().ConfigureAwait(false);
+
+            logger.LogInformation($"StudySubject updated succesfully.");
+
+            return mapper.Map<StudySubjectCreateUpdateDto>(updatedStudySubject);
```
DmyMi: return the "display dto"

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -0,0 +1,201 @@
...
+            throw;
+        }
+    }
+
+    private async Task CheckIfLanguagesIdsAreCorrect(StudySubjectCreateUpdateDto dto)
+    {
+        var existingLanguages = await languageRepository.GetAll().ConfigureAwait(false);
+        var existingLanguageIds = existingLanguages.Select(x => x.Id).ToHashSet();
+
+        if (dto.LanguageIds.Where(id => !existingLanguageIds.Contains(id)).ToList().Any() || !existingLanguageIds.Contains(dto.PrimaryLanguageId))
```
DmyMi: You are doing a foreign key check that is done by the database. Can leave this part to the database

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -0,0 +1,201 @@
...
+        var ukrainianLanguage = existingLanguages.FirstOrDefault(x => x.Code.Equals("uk", StringComparison.OrdinalIgnoreCase));
+        if (ukrainianLanguage == null)
+        {
+            logger.LogDebug("Operation failed, Ukrainian language is not found in the database.");
+            throw new ArgumentException("Ukrainian language is not found in the database.");
+        }
+
+        var ukrainianLanguageId = ukrainianLanguage.Id;
+
+        if (dto.IsPrimaryLanguageUkrainian && ukrainianLanguageId != dto.PrimaryLanguageId)
```
DmyMi: this can be simplified if `IsPrimaryLanguageUkrainian = true` by just getting the Ukrainian language id & setting it ourself on backend. This should not be considered as error.

OutOfSchool/OutOfSchool.BusinessLogic/Util/MappingProfile.cs
```diff
@@ -817,6 +820,33 @@ public MappingProfile()
...
+            .ForMember(dest => dest.Document, opt => opt.Ignore())
+            .ForMember(dest => dest.File, opt => opt.Ignore())
+            .ForMember(dest => dest.ActiveFrom, opt => opt.Ignore())
+            .ForMember(dest => dest.ActiveTo, opt => opt.Ignore())
+            .ForMember(dest => dest.IsBlocked, opt => opt.Ignore())
+            .ForMember(dest => dest.PrimaryLanguage, opt => opt.Ignore())
+            .ForMember(dest => dest.Workshop, opt => opt.Ignore())
+            .ForMember(dest => dest.StudySubjectLanguages,
+                opt => opt.MapFrom(src => src.LanguageIds.Select(id => new StudySubjectLanguage { LanguageId = id })))
+            .ReverseMap();
```
DmyMi: there should not be a reverse map here :)

OutOfSchool/OutOfSchool.BusinessLogic/Util/MappingProfile.cs
```diff
@@ -817,6 +820,33 @@ public MappingProfile()
...
+
+        CreateMap<StudySubject, StudySubjectDto>()
+            .ForMember(dest => dest.LanguageIds,
+                opt => opt.MapFrom(src => src.StudySubjectLanguages.Select(x => x.LanguageId)))
+            .ReverseMap();
```
DmyMi: this is "display dto", there should also not be a reverse map. And here you use a language dto, not id.

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/StudySubjectConfiguration.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+using OutOfSchool.Services.Models.Configurations.Base;
+
+namespace OutOfSchool.Services.Models.Configurations;
+public class StudySubjectConfiguration : BusinessEntityConfiguration<StudySubject>
+{
+    public override void Configure(EntityTypeBuilder<StudySubject> builder)
```
DmyMi, line 8: call base configure method before using your code

OutOfSchool/OutOfSchool.DataAccess/Models/StudySubjectLanguage.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+
+namespace OutOfSchool.Services.Models;
+/// <summary>
+/// Table for many-to-many relationship between StudySubject and Language
+/// </summary>
+public class StudySubjectLanguage : IKeyedEntity<long>
```
DmyMi: this is handled by EF automatically, no need to specify explicitly. There is no business usecase for this entity right now.

OutOfSchool/OutOfSchool.DataAccess/Models/Language.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    /// <summary>
+    /// Title of the language
+    /// </summary>
+    [Required(ErrorMessage = "The title is required.")]
+    [MinLength(1, ErrorMessage = "The title must be at least 1 character.")]
+    [MaxLength(50, ErrorMessage = "The title of instruction can't exceed 50 characters.")]
+    public string Title { get; set; }
+
+    public virtual List<StudySubject> StudySubjects { get; set; }
```
DmyMi: Check [this MS docs](https://learn.microsoft.com/en-us/ef/core/modeling/relationships/many-to-many#unidirectional-many-to-many). This should be a unidirectional many-to-many in such a way that only Subjects know about languages. This is because language can later be re-used by other entities.

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/StudySubjectLanguageConfiguration.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+
+namespace OutOfSchool.Services.Models.Configurations;
+public class StudySubjectLanguageConfiguration : IEntityTypeConfiguration<StudySubjectLanguage>
+{
+    public void Configure(EntityTypeBuilder<StudySubjectLanguage> builder)
+    {
+        builder.HasKey(x => x.Id);
+
+        builder.HasOne(x => x.StudySubject)
```
DmyMi: check language & StudySubjectLanguage comments :)

OutOfSchool/OutOfSchool.DataAccess/Models/StudySubject.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// </summary>
+    [Required(ErrorMessage = "The primary language's ID is required.")]
+    public long PrimaryLanguageId { get; set; }
+    public virtual Language PrimaryLanguage { get; set; }
+
+    /// <summary>
+    /// Id of the related workshop
+    /// </summary>
+    [Required(ErrorMessage = "The workshop's id is required.")]
+    public Guid WorkshopId{ get; set; }
```
DmyMi: do not link it to workshop for now

OutOfSchool/OutOfSchool.DataAccess/Models/Workshop.cs
```diff
@@ -191,5 +191,6 @@ public virtual ICollection<Workshop>
 
     public virtual List<Tag> Tags { get; set; }
 
+    public virtual List<StudySubject> StudySubjects { get; set; }
```
DmyMi: do not link it to workshop for now

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/StudySubjectController.cs
```diff
@@ -0,0 +1,272 @@
...
+    /// Get StudySubject by it's id.
+    /// </summary>
+    /// <param name="id">StudySubject's id.</param>
+    /// <returns>StudySubject.</returns>
+    [AllowAnonymous]
+    [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(StudySubjectDto))]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpGet("{id}")]
+    public async Task<IActionResult> GetById(Guid id)
```
DmyMi: currently, this API is strictly for provider, and they are able to access only their own subjects. So it should be something like `/api/v1/providers/{providerid}/studysubjects/{id}` with all the checks like you've done in create/update methods.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/StudySubjectController.cs
```diff
@@ -0,0 +1,272 @@
...
+    /// <summary>
+    /// Get filtered list of StudySubjects from the database.
+    /// </summary>
+    /// <returns>List of StudySubjects.</returns>
+    [AllowAnonymous]
+    [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(IEnumerable<StudySubjectDto>))]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpGet]
+    public async Task<IActionResult> Get([FromQuery] SearchStringFilter filter = null)
```
DmyMi: currently, this API is strictly for provider, and they are able to access only their own subjects. So it should be something like `/api/v1/providers/{providerid}/studysubjects` with all the checks like you've done in create/update methods.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/StudySubjectController.cs
```diff
@@ -0,0 +1,272 @@
...
+    [Authorize]
+    [HasPermission(Permissions.WorkshopEdit)]
+    [ProducesResponseType(StatusCodes.Status201Created)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpPost]
+    public async Task<IActionResult> Create([FromBody] StudySubjectCreateUpdateDto dto)
```
DmyMi: currently, this API is strictly for provider, and they are able to access only their own subjects. So it should be something like `/api/v1/providers/{providerid}/studysubjects`.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/StudySubjectController.cs
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        var isWorkshopExists = await _workshopService.Exists(dto.WorkshopId).ConfigureAwait(false);
+
+        if(!isWorkshopExists)
+        {
+            return NotFound("There's no such workshop in the database.");
+        }
+
+        var providerId = await _providerService.GetProviderIdForWorkshopById(dto.WorkshopId).ConfigureAwait(false);
```
DmyMi: these provider checks can be moved to service method

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/StudySubjectController.cs
```diff
@@ -0,0 +1,272 @@
...
+        {
+            return StatusCode(403, "It is forbidden to add study subjects to workshops at blocked providers");
+        }
+
+        if (!ModelState.IsValid)
+        {
+            return BadRequest(ModelState);
+        }
+
+        var userHasRights = await this.IsUserProvidersOwnerOrAdmin(dto.WorkshopId).ConfigureAwait(false);
```
DmyMi: As for now we're only checking for provider and not linking to workshop - can skip this

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/StudySubjectController.cs
```diff
@@ -0,0 +1,272 @@
...
+    /// <returns>StudySubject.</returns>
+    [Authorize]
+    [HasPermission(Permissions.WorkshopEdit)]
+    [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(StudySubjectDto))]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpPut]
+    public async Task<IActionResult> Update([FromBody] StudySubjectCreateUpdateDto dto)
```
DmyMi: same as for create

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/StudySubjectController.cs
```diff
@@ -0,0 +1,272 @@
...
+    /// <param name="id">StudySubject's id.</param>
+    /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
+    [Authorize]
+    [HasPermission(Permissions.WorkshopEdit)]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpDelete("{id}")]
+    public async Task<IActionResult> Delete(Guid id)
```
DmyMi: Same as for the get by id method.

OutOfSchool/OutOfSchool.BusinessLogic/Models/StudySubjects/StudySubjectCreateUpdateDto.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// Primary language of the subject
+    /// </summary>
+    [Required(ErrorMessage = "Primary language's id is required.")]
+    public int PrimaryLanguageId { get; set; }
+
+    /// <summary>
+    /// Id of the related workshop
+    /// </summary>
+    [Required(ErrorMessage = "The workshop's id is required.")]
+    public Guid WorkshopId { get; set; }
```
DmyMi: do not link to workshop for now.

DmyMi (changes requested):

OutOfSchool/OutOfSchool.BusinessLogic/Util/MappingProfile.cs
```diff
@@ -865,6 +866,33 @@ public MappingProfile()
...
+            .ForMember(dest => dest.PrimaryLanguage, opt => opt.Ignore())
+            .ForMember(dest => dest.Languages,
+                opt => opt.Ignore())
+            .ForMember(dest => dest.PrimaryLanguageId,
+                opt => opt.MapFrom(src => src.LanguagesSelection.FirstOrDefault(l => l.IsPrimary).Id));
+
+        CreateMap<StudySubject, StudySubjectDto>()
+            .ForMember(dest => dest.WorkshopId, opt => opt.Ignore())
+            .ForMember(dest => dest.Languages,
+                opt => opt.MapFrom(src => src.Languages.Select(l => new Language { Id = l.Id, Code = l.Code, Name = l.Name } )));
```
DmyMi: you are mapping from entity to dto, are you sure `new Language` is the correct object? Also, you already have `CreateMap<Language, LanguageDto>().ReverseMap();` so this field should be mapped automatically, no need to specify anything in this case.

DmyMi (approved):

PR #1629, opened 2025-01-13
Fixed Provider Upload the list of employees

HlibBondarev:
1) Added check - 'is Provider exists?' to UploadEmployeesForProvider() method of ProviderService class.
2) Added check - 'does Provider has rights to upload employees?' to UploadEmployeesForProvider() method of ProviderService class.
3) Added attribute - [ProducesResponseType(StatusCodes.Status403Forbidden)] to Upload action of ProviderController class.
4) Fixed tests for UploadEmployeesForProvider() method of ProviderService class.
5) Added tests for 2 cases ('When Provider does not exist' and 'When User does not have rights') to ProviderServiceTests class.

### DIFF
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Services/ProviderServices/ProviderService.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Services/ProviderServices/ProviderService.cs
@@ -542,7 +542,8 @@ public class ProviderService : IProviderService, ISensitiveProviderService
 
         if (!isProviderExists)
         {
-            throw new UnauthorizedAccessException($"User has no rights to perform operation. Provider with Id = {id} doesn't exist.");
+            logger.LogError("User has no rights to perform operation. Provider with Id = {id} doesn't exist.", id);
+            throw new UnauthorizedAccessException($"User has no rights to perform operation.");
         }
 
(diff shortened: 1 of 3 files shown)


OutOfSchool/OutOfSchool.BusinessLogic/Services/ProviderServices/ProviderService.cs
```diff
@@ -538,6 +538,15 @@ public Task<bool> Exists(Guid id)
 
     public async Task<UploadEmployeeResponse> UploadEmployeesForProvider(Guid id, UploadEmployeeRequestDto[] data)
     {
+        var isProviderExists = await Exists(id).ConfigureAwait(false);
+
+        if (!isProviderExists)
+        {
+            throw new UnauthorizedAccessException($"User has no rights to perform operation. Provider with Id = {id} doesn't exist.");
```
DmyMi: leave only this text "User has no rights to perform operation."

"Provider with Id = {id} doesn't exist" can be correctly formatted and used as log message.

HlibBondarev: Ok, fixed

DmyMi (approved):

PR #1632, opened 2025-01-15
Add additional export data

DmyMi:
Add additional export data for external system to consume.

Possible (re)implementation of logic inside the export service as later this might get extracted into separate module.

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi.Tests/Services/ExternalExportSubDirectionsTests.cs
+++ b/OutOfSchool/OutOfSchool.WebApi.Tests/Services/ExternalExportSubDirectionsTests.cs
@@ -19,6 +19,7 @@ using OutOfSchool.Services.Repository;
 using OutOfSchool.Services.Repository.Api;
 using OutOfSchool.Services.Repository.Base;
 using OutOfSchool.Tests.Common;
+using OutOfSchool.Tests.Common.TestDataGenerators;
 
 namespace OutOfSchool.WebApi.Tests.Services;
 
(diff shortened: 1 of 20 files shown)


OutOfSchool/OutOfSchool.WebApi.Tests/Services/ExternalExportSubDirectionsTests.cs
```diff
@@ -0,0 +1,186 @@
...
+        Assert.IsNotNull(result);
+        Assert.AreEqual(2, result.TotalAmount);
+        Assert.AreEqual(2, result.Entities.Count);
+        Assert.AreEqual(1, result.Entities.First(s => s.Id == Guid.Parse("b7e1322e-7575-48c1-a444-4effb8f4d083")).DirectionIds.Count);
+        Assert.AreEqual(2, result.Entities.First(s => s.Id == Guid.Parse("a042661d-9be8-4bfb-adcd-06cbe91388a0")).DirectionIds.Count);
+    }
+
+    private void SeedSubDirections(IInstitutionHierarchyRepository repository)
+    {
+        var fakeInstitutions = new List<Institution>
```
VadymLevkovskyi: Bogus could be used

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ExternalExportController.cs
```diff
@@ -65,4 +67,52 @@ public async Task<IActionResult> GetWorkshopsByFilter([FromQuery] DateTime updat
...
+    /// <param name="offsetFilter">Filter to get a part of all directions that were found.</param>
+    /// <returns><see cref="SearchResult{DirectionInfoDto}"/>, or no content.</returns>
+    [HttpGet]
+    [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(SearchResult<DirectionInfoDto>))]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [Route("directions")]
+    public async Task<IActionResult> GetDirectionsByFilter([FromQuery] OffsetFilter offsetFilter)
+    {
+        try
```
VadymLevkovskyi: Just an idea. This 
```try { var result = await DoSomething(); return mapped(result); } catch { return 500 code; }```
seems like a pattern that could be refactored via `Task<T>` extension method `ProtectAndMap` that awaits passed taks, and either catches and transforms errors to StatusCode or transforms result using passed map function, like this
```= > await DoSomething().ProtectAndMap(mapped);```

VadymLevkovskyi (approved):

PR #1635, opened 2025-01-17
TKyryk/ add endpoint to fetch competitive event view cards by organizer ID

Tetyana-K:
сreated DTO for Competitive Event View Card
added GetByProviderId method in the service to fetch competitive event view cards by organizer ID (provider's Id)
added endpoint in the controller
added unit tests for the service and controller

### DIFF
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Services/CompetitiveEventService.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Services/CompetitiveEventService.cs
@@ -124,7 +124,7 @@ public class CompetitiveEventService : ICompetitiveEventService
     /// <inheritdoc/>
     public async Task<SearchResult<CompetitiveEventViewCardDto>> GetByProviderId(Guid id, ExcludeIdFilter filter)
     {
-        logger.LogDebug($"Getting Competitive events by organization started. Looking ProviderId = {id}.");
+        logger.LogDebug("Getting Competitive events by organization started. Looking ProviderId = {id}.", id);
 
         filter ??= new ExcludeIdFilter();
         ValidateExcludedIdFilter(filter);
@@ -149,9 +149,14 @@ public class CompetitiveEventService : ICompetitiveEventService
 
         var competitiveEventViewCards = mapper.Map<List<CompetitiveEventViewCardDto>>(competitiveEvents);
 
-        logger.LogDebug(!competitiveEventViewCards.Any()
-          ? $"There aren't CompetitiveEvents for Provider with Id = {id}."
-          : $"From CompetitiveEvents table were successfully received {competitiveEventViewCards.Count()} records.");
+        if (competitiveEventViewCards.Any())
+        {
+            logger.LogDebug("From CompetitiveEvents table were successfully received {count} records.", competitiveEventViewCards.Count);
+        }
+        else
+        {
+            logger.LogDebug("There aren't CompetitiveEvents for Provider with Id = {id}.", id);
+        }
 
         var result = new SearchResult<CompetitiveEventViewCardDto>()
         {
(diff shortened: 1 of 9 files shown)


OutOfSchool/OutOfSchool.BusinessLogic/Services/CompetitiveEventService.cs
```diff
@@ -121,6 +121,55 @@ public async Task Delete(Guid id)
...
+        {
+            predicate = predicate.And(x => x.Id != filter.ExcludedId);
+        }
+        var competitiveEventCardsCount = await competitiveEventRepository.Count(
+            whereExpression: predicate).ConfigureAwait(false);
+
+        var competitiveEvents = await competitiveEventRepository.Get(
+            skip: filter.From,
+            take: filter.Size,
+            includeProperties: null,
```
DmyMi: extract `includeProperties` as a string readonly property, that can be empty for now. Add a `TODO` to fill it when we have final list of properties. Will be easier to modify just that string.

OutOfSchool/OutOfSchool.BusinessLogic/Services/CompetitiveEventService.cs
```diff
@@ -121,6 +121,55 @@ public async Task Delete(Guid id)
...
+            whereExpression: predicate)
+            .ToListAsync()
+            .ConfigureAwait(false);
+
+        var competitiveEventViewCards = mapper.Map<List<CompetitiveEventViewCardDto>>(competitiveEvents);
+
+        if (competitiveEventViewCards.Any())
```
DmyMi: `logger.LogDebug("{Count} CompetitiveEvents were successfully received", competitiveEventViewCards.Count);` is more then enough, no need for if else just for logging. If `count = 0` anyone would understand that the list is empty :)

OutOfSchool/OutOfSchool.BusinessLogic/Services/CompetitiveEventService.cs
```diff
@@ -121,6 +121,55 @@ public async Task Delete(Guid id)
         }
     }
 
+    /// <inheritdoc/>
+    public async Task<SearchResult<CompetitiveEventViewCardDto>> GetByProviderId(Guid id, ExcludeIdFilter filter)
+    {
+        logger.LogDebug("Getting Competitive events by organization started. Looking ProviderId = {id}.", id);
+
```
DmyMi, line 143: Use `ICurrentUserService.UserHasRights` method to check if user can actually get info about this provider's competitive events as this is provider only method.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/CompetitiveEventController.cs
```diff
@@ -144,6 +145,39 @@ public async Task<IActionResult> Delete(Guid id)
...
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpGet("/api/v{version:apiVersion}/provider/{id}/competitiveevents")]
+    public async Task<IActionResult> GetCompetitiveEventViewCardByProviderId(Guid id, [FromQuery] ExcludeIdFilter filter)
+    {
+        if (id == Guid.Empty)
+        {
+            return BadRequest("Provider id is empty.");
+        }
+        if (filter?.ExcludedId == Guid.Empty)
```
DmyMi: don't do this please, this is not a bad request. You already have a check in service, just extend it, e.g., `filter.ExcludedId is not null && filter.ExcludedId != Guid.Empty`

DmyMi (approved):

PR #1637, opened 2025-01-20
Added the Authorize attribute to the methods whereby BusinessEntityIn…

### LINKED ISSUE #1600: BusinessEntityInterceptor can't retrieve User Id from HttpContext
Investigate why `BusinessEntityInterceptor` can't retrieve User Id when a logged in user tries to save or update a business entity.

AndrD1406:
Adding the Authorize attribute although HasPermission is derived from Authorize seems to be the only way to make the BusinessEntityInterceptor able to retrieve the UserId. Also added it to the Delete method because an unauthenticated user would never be able to delete an existing Workshop anyway (This method had the HasPermission as well so it seems to be right).

### DIFF
--- a/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopController.cs
+++ b/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopController.cs
@@ -263,6 +263,7 @@ public class WorkshopController : ControllerBase
     /// <response code="403">If the user has no rights to use this method, or sets some properties that are forbidden.</response>
     /// <response code="500">If any server error occures.</response>
     [HasPermission(Permissions.WorkshopAddNew)]
+    [Authorize]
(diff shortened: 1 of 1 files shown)


OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopController.cs
```diff
@@ -263,6 +263,7 @@ public async Task<IActionResult> GetByFilter([FromQuery] WorkshopFilter filter,
     /// <response code="403">If the user has no rights to use this method, or sets some properties that are forbidden.</response>
     /// <response code="500">If any server error occures.</response>
     [HasPermission(Permissions.WorkshopAddNew)]
+    [Authorize]
```
DmyMi, line 266: can you double check that `HasPermission` does not work as authorize as it extends authorize? Maybe it is better to fix HasPermission instead of duplicating attributes

PR #1711, opened 2025-03-11
Price Tracker Enhancement for Workshops

### LINKED ISSUE #1692: Price Tracker Enhancement for Workshops
## Overview
As part of the feature request [Frontend story #2675](https://github.com/ita-social-projects/OoS-Frontend/issues/2675), we need to dynamically update the price tracker to display the minimum and maximum prices found among workshops that match all selected filter parameters (except the price itself). 

## Task description

1. Create a new endpoint (e.g., `GET /api/v1/workshops/price-range`) which accepts a modified workshop filter.  
   - The filter should include all existing non-price-related parameters and also a new field (e.g., `PayRateType`) that indicates whether the price should be interpreted per class, course, year, etc..
   - The endpoint should return:
     - Minimum price among all matching workshops.
     - Maximum price among all matching workshops.

2. Based on our existing strategy pattern for workshop searches, create separate logic for fetching the min/max price:
   - Extend the existing Elasticsearch strategy to include an aggregator query that computes the minimum and maximum prices for the current filter (not including any price range parameters).
   - Enhance the existing database query logic to compute the min and max prices using EF Core queries.
   - Depending on the availability/status of Elasticsearch, dynamically decide which strategy to invoke. 
     
3. Extend the current filter models:
     - Add a new property (`PayRateType`) to the `WorkshopFilter` model.
     - For `WorkshopFilterES`, add a similar property, so that the mapping between the business logic and Elasticsearch layers is consistent.
   - Add relevant queries to database and Elasticsearch filter methods to get the workshops list.

4. Ensure that the response structure is consistent with the rest of the API responses (including proper error handling and HTTP status codes).

5. Investigate and, if beneficial, implement caching strategies (such as the use of HTTP Cache) to store results of frequent min/max queries, especially if the underlying data does not change very often.

omaiva:
- Modified WorkshopFilter and WorkshopFilterES
- Added aggregator query to ESWorkshopProvider
- Extended ElasticSearch\Workshop strategy pattern
- Modified workshop services
- Created new endpoint for WorkshopController

## Summary by CodeRabbit

- **New Features**
	- Introduced a new pricing feature that enables users to view the minimum and maximum workshop prices.
	- Enhanced workshop filtering by incorporating pay rate criteria for more refined searches.
	- Added an API endpoint to retrieve price range information using advanced query methods.

- **Tests**
	- Expanded test coverage to confirm accurate price calculations and robust behavior under various filtering scenarios.


OutOfSchool/Directory.Packages.props
```diff
@@ -23,6 +23,7 @@
     <PackageVersion Include="Microsoft.Extensions.Configuration.UserSecrets" Version="8.0.1" />
     <PackageVersion Include="Microsoft.Extensions.Hosting" Version="8.0.1" />
     <PackageVersion Include="Microsoft.Extensions.Http.Resilience" Version="8.10.0" />
+    <PackageVersion Include="NEST" Version="7.17.5" />
```
DmyMi: We are using the new version of the client libarary `<PackageVersion Include="Elastic.Clients.Elasticsearch" Version="8.15.6" />`, please remove this old version, do a `dotnet clean` and make sure your feature works with new library API.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -781,6 +781,35 @@ public async Task<SearchResult<WorkshopDto>> FetchByFilterForAdmins(WorkshopFilt
...
     }
 
+    /// <inheritdoc/>
+    public async Task<PriceRange> GetPriceRange(WorkshopFilter filter = null)
+    {
+        logger.LogDebug("Getting Price Range for workshops by filter started.");
+
+        filter ??= new WorkshopFilter();
+
+        var filterPredicate = PredicateBuild(filter);
```
DmyMi: you can't reuse the same predicate without modification, as it already adds price as part of `where` clause. you probably can modify the `PredicateBuild` method to accept some parameter to ignore adding price to predicate for this usecase.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -781,6 +781,35 @@ public async Task<SearchResult<WorkshopDto>> FetchByFilterForAdmins(WorkshopFilt
...
+
+        var query = await workshopRepository.GetByFilter(
```
DmyMi: `GetByFilter` return an enumerable, so technically this request will return 1000s workshops for you to process and receive OOM exception :). You can use `Get` to return queryable with the given `where` predicate, then do some SQL/Linq magic, e.g.:
```csharp
queriable
    .GroupBy(_ => 1)
    .Select(g => new PriceRange
    {
        MinPrice = g.Min(p => p.Price),
        MaxPrice = g.Max(p => p.Price)
    })
```

Though need to investigate if `GroupBy(_ => 1)` is still a thing to make 1 single query, cause it was a solution from [2018](https://github.com/dotnet/efcore/issues/13515)

OutOfSchool/OutOfSchool.BusinessLogic/Util/Mapping/ElasticProfile.cs
```diff
@@ -213,5 +213,7 @@ public ElasticProfile()
                 opt =>
                     opt.MapFrom(src =>
                         src.Coverage.Title));
+
+        CreateMap<PriceRange, PriceRangeES>();
```
DmyMi: do you really need this map? shouldn't it be the other way around?

OutOfSchool/OutOfSchool.ElasticsearchData/ESCompetitiveEventProvider.cs
```diff
@@ -31,6 +33,43 @@ public override async Task<SearchResultES<CompetitiveEventES>> Search(Competitiv
         };
     }
 
+    public override async Task<PriceRangeES> GetPriceRangeAsync(CompetitiveEventFilterES filter = null)
+    {
+        filter ??= new CompetitiveEventFilterES();
+
+        var query = CreateQueryFromFilter(filter);
```
DmyMi: here might be the same case as with database, price may be included in query filter so need a way (parameter) to not add it so you can reuse this query builder.

OutOfSchool/OutOfSchool.ElasticsearchData/ESWorkshopProvider.cs
```diff
@@ -39,6 +40,43 @@ public override async Task<SearchResultES<WorkshopES>> Search(WorkshopFilterES f
         };
     }
 
+    public override async Task<PriceRangeES> GetPriceRangeAsync(WorkshopFilterES filter = null)
+    {
+        filter ??= new WorkshopFilterES();
+
+        var query = CreateQueryFromFilter(filter);
```
DmyMi: here might be the same case as with database, price may be included in query filter so need a way (parameter) to not add it so you can reuse this query builder.

OutOfSchool/OutOfSchool.ElasticsearchData/ESCompetitiveEventProvider.cs
```diff
@@ -31,6 +33,43 @@ public override async Task<SearchResultES<CompetitiveEventES>> Search(Competitiv
...
+
+        var request = new SearchRequest<CompetitiveEventES>
+        {
+            Query = query,
+            Aggregations = new Dictionary<string, Aggregation>
+            {
+                {
+                    "min_price", Aggregation.Min(new MinAggregation
```
DmyMi: use constants for `min_price` / `max_price` strings

OutOfSchool/OutOfSchool.ElasticsearchData/ESWorkshopProvider.cs
```diff
@@ -39,6 +40,43 @@ public override async Task<SearchResultES<WorkshopES>> Search(WorkshopFilterES f
...
+
+        var request = new SearchRequest<WorkshopES>
+        {
+            Query = query,
+            Aggregations = new Dictionary<string, Aggregation>
+            {
+                {
+                    "min_price", Aggregation.Min(new MinAggregation
```
DmyMi: use constants for `min_price` / `max_price` strings

OutOfSchool/OutOfSchool.WebApi/Startup.cs
```diff
@@ -111,6 +111,8 @@ public static void Configure(this WebApplication app)
         // Enable extracting token from QueryString for Hub-connection authorization
         app.UseMiddleware<AuthorizationTokenMiddleware>();
 
+        app.UseResponseCaching();
```
DmyMi: double check if we have no extensions that set this up, as i think we've added some caching already in the project.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopController.cs
```diff
@@ -521,6 +521,29 @@ public async Task<IActionResult> Delete(Guid id)
...
+    /// <summary>
+    /// Gets price range of filtered workshops.
+    /// </summary>
+    /// <param name="filter">Filter for workshop entities.</param>
+    /// <returns><see cref="PriceRange"/></returns>
+    /// <response code="200">If the price range was returned.</response>
+    /// <response code="401">If the user is not authorized.</response>
+    /// <response code="403">If the user has no rights to use this method, or sets some properties that are forbidden to change.</response>
+    /// <response code="500">If any server error occures.</response>
+    [ResponseCache(Duration = 300, Location = ResponseCacheLocation.Client, VaryByQueryKeys = new[] { "*" })]
```
DmyMi: `Client` means `Cache-Control: private` (it is stored in the browser of each user separately), we need to have it as `Cache-Control: public` - it is cached in the internet and all users can hit this cache. There should be an option for this.

DmyMi (changes requested):

DmyMi (approved):

PR #1721, opened 2025-03-18
Features hiding

### LINKED ISSUE #1715: Hiding Features via Feature Flags
## Overview
Implement feature toggling to hide legacy functionalities or functionalities that are going to get complete overhaul from end users. The backend is configured with .NET Feature Management, and the feature flags must be exposed via an API endpoint (`/api/v1/FeatureManagement/Get`). The frontend should then consume these flags to disable or hide the corresponding UI elements.

## Task Description

- Create individual feature flags for each functionality to be hidden. These include:
  - Login & Authorization
    - Sending email confirmations to new users
    - Password change and recovery via email

  - Administration Panel
    - In the **Напрямки** section: Management (addition, editing, deletion) of workshop directions for end users (напрямки для користувача).
    Image

    - The **Користувачі** section entirely with all operations (view, block, remove).
  - Achievements:
    - Addition of Workshop achievements across the system
    Image

  - Other Functionalities:
    - Messaging (chat between service providers and parents)
    - English version of the site and localization change

### Expose Flags Endpoint
- Update an API endpoint that returns the current state of all feature flags.
  - Ensure the response is well-structured so the frontend can easily parse and use the flag values.

### Frontend Flag Consumption:
- UI adjustments based on the flag values:
  - Disable or hide the UI elements corresponding to the legacy functionalities.
  - Ensure that when a flag indicates the feature is off, the respective components or buttons are either not rendered or are disabled.

annygumeniuk:
## Summary by CodeRabbit

- **New Features**
  - Introduced dynamic feature toggles for key functionalities, including email confirmation, password registration, achievement creation, messaging, and administrative controls. These improvements allow endpoints to be conditionally available based on enabled features.

- **Chores**
  - Expanded configuration settings and system enumerations with additional feature flags (disabled by default) to support granular control and smoother rollout of new functionality.

### DIFF
--- a/OutOfSchool/OutOfSchool.AuthCommon/Controllers/AccountController.cs
+++ b/OutOfSchool/OutOfSchool.AuthCommon/Controllers/AccountController.cs
@@ -465,6 +465,7 @@ public class AccountController : Controller
         return await userService.LogOutUserById(userId);
     }
 
+    [FeatureGate(AuthServerConstants.FeatureManagement.EmailConfirmation)]
     [Obsolete("Confirm email API is no longer supported. Exists only for testing purposes.")]
     private async Task<IActionResult> SendConfirmEmailProcess(string action, User user, string razorTemplate, object passedData)
     {
--- a/OutOfSchool/OutOfSchool.AuthCommon/Controllers/AccountController.cs
+++ b/OutOfSchool/OutOfSchool.AuthCommon/Controllers/AccountController.cs
@@ -464,8 +464,7 @@ public class AccountController : Controller
 
         return await userService.LogOutUserById(userId);
     }
-
-    [FeatureGate(AuthServerConstants.FeatureManagement.EmailConfirmation)]
+    
     [Obsolete("Confirm email API is no longer supported. Exists only for testing purposes.")]
     private async Task<IActionResult> SendConfirmEmailProcess(string action, User user, string razorTemplate, object passedData)
     {
--- a/OutOfSchool/OutOfSchool.AuthCommon/Controllers/AuthController.cs
+++ b/OutOfSchool/OutOfSchool.AuthCommon/Controllers/AuthController.cs
@@ -462,8 +462,7 @@ public class AuthController : Controller
             return View("Register", model);
         }
     }
-
-    [FeatureGate(AuthServerConstants.FeatureManagement.PasswordRegistration)]
+    
     [Obsolete("Registration API is no longer supported. Exists only for testing purposes.")]
     private bool GetProviderRegistrationFromUri(string returnUrl)
     {
--- a/OutOfSchool/OutOfSchool.AuthCommon/Controllers/AuthController.cs
+++ b/OutOfSchool/OutOfSchool.AuthCommon/Controllers/AuthController.cs
@@ -463,6 +463,7 @@ public class AuthController : Controller
         }
     }
 
+    [FeatureGate(AuthServerConstants.FeatureManagement.PasswordRegistration)]
     [Obsolete("Registration API is no longer supported. Exists only for testing purposes.")]
     private bool GetProviderRegistrationFromUri(string returnUrl)
     {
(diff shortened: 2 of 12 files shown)


OutOfSchool/OutOfSchool.AuthCommon/Controllers/AccountController.cs
```diff
@@ -465,6 +465,7 @@ public async Task<ResponseDto> LogOutUser(string userId)
         return await userService.LogOutUserById(userId);
     }
 
+    [FeatureGate(AuthServerConstants.FeatureManagement.EmailConfirmation)]
```
DmyMi: technically private methods should not be disabled if the public method that uses them is disabled.

OutOfSchool/OutOfSchool.AuthCommon/Controllers/AuthController.cs
```diff
@@ -463,6 +463,7 @@ public async Task<IActionResult> Register(RegisterViewModel model)
         }
     }
 
+    [FeatureGate(AuthServerConstants.FeatureManagement.PasswordRegistration)]
```
DmyMi: same private methods comment

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/AchievementController.cs
```diff
@@ -94,6 +96,7 @@ public async Task<IActionResult> GetByWorkshopId([FromQuery] AchievementsFilter
     [ProducesResponseType(StatusCodes.Status401Unauthorized)]
     [ProducesResponseType(StatusCodes.Status403Forbidden)]
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [FeatureGate(nameof(Feature.CreateAchievement))]
```
DmyMi: you can also add it for update and delete, can leave get methods without feature gate.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChildController.cs
```diff
@@ -70,6 +73,7 @@ public async Task<IActionResult> GetAllForAdmin([FromQuery] ChildSearchFilter fi
     [ProducesResponseType(StatusCodes.Status403Forbidden)]
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     [HttpGet("/api/v{version:apiVersion}/parents/{id}/children")]
+    [FeatureGate(nameof(Feature.AdminsChildernParentsManagement))]
```
DmyMi, line 76: are you sure this method is not used to display children for parent? double check with frontend

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChildController.cs
```diff
@@ -277,6 +281,7 @@ public async Task<IActionResult> Update([FromBody] ChildUpdateDto dto, Guid id)
     [ProducesResponseType(StatusCodes.Status403Forbidden)]
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     [HttpDelete("{id}")]
+    [FeatureGate(nameof(Feature.AdminsChildernParentsManagement))]
```
DmyMi: delete also can be used by parents i think, need to check with frontend.

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ParentController.cs
```diff
@@ -100,6 +103,7 @@ public async Task<ActionResult<ParentDTO>> GetProfile()
     [ProducesResponseType(StatusCodes.Status400BadRequest)]
     [ProducesResponseType(StatusCodes.Status403Forbidden)]
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [FeatureGate(nameof(Feature.AdminsChildernParentsManagement))]
```
DmyMi, line 106: double check if this method is used by admins only, as we can block parent by provider (so parent can't apply to workshops)

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ChildController.cs
annygumeniuk: To display children for parent we use GetUsersChildren

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/ParentController.cs
annygumeniuk: In this case we use BlockedProviderParent/Block method

DmyMi (approved):

PR #1731, opened 2025-03-21
Add functionality to attach and detach StudySubject to Workshops

max-khakimzhanov:
This PR introduces **a complete set of features to attach and detach StudySubjects from Workshops.**
### **🔹 Database Updates:**  
- Added a relationship between StudySubjects and Workshops.  
- Added a direct relationship between StudySubjects and Providers — now each StudySubject is explicitly linked to its owning provider.
- Created and applied a new migration to reflect these changes in the database schema.  

### **🔹 Services Updates**

**UpdateWorkshopsForStudySubject:**
- Implements logic to attach or detach workshops based on `IsAttached` flag.
- Validates **provider ownership** before applying changes.
- Logs cases when a passed workshop is not found.
- Optimized in-memory checks and LINQ calls for performance.

**DetachAllWorkshops:**
- Detaches all workshops belonging to a provider from a given study subject.
- Includes early exit checks to avoid unnecessary DB writes.
- Uses filtered queries to reduce loaded data.

**GetAttachedWorkshops:**
- Retrieves **paginated** list of provider’s workshops with their attachment status to a specific study subject.
- Applies efficient filtering and projection in a single query using LINQ.
- Prevents potential N+1 issues by ensuring eager loading where needed.

### **🔹 DTOs & Utilities**
- Introduced `WorkshopAttachmentStatusDto` for transmitting workshop ID, title, and attachment flag.
- Introduced `PaginatedResult<T>` and extension method `ToPaginatedResultAsync` for reusable pagination.

### **🔹 Controller & Routing**
 **RESTful Routes**:
- `GET /providers/{providerId}/studysubjects/{studySubjectId}/workshops/attached` — get paginated workshops with attachment status.
- `PUT /api/v1/providers/{providerId}/studysubjects/{studySubjectId}/workshops` — update (attach/detach) specific workshops.
- `DELETE /api/v1/providers/{providerId}/studysubjects/{studySubjectId}/workshops` — detach all workshops for a provider.

### **🔹Testing**
 **Unit tests added/updated** for:
- `GetAttachedWorkshops`: including pagination, filtering, and boundary checks.
- `UpdateWorkshopsForStudySubject`: checks ownership, attachment toggling, and invalid workshop handling.
- `DetachAllWorkshops`: verifies correct filtering and removal.
- Mocked `IQueryable` queries now support `ToPaginatedResultAsync` via `TestAsyncEnumerableQuery`.

## Summary by CodeRabbit

- **New Features**
  - Expanded workshop management by allowing study subjects to be linked with multiple workshops via provider associations.
  - Introduced endpoints to update, detach, and retrieve workshop attachment statuses with built-in pagination and sorting.
  - Enhanced API responses with improved validations and error handling for workshop management.

- **Tests**
  - Expanded test coverage to verify the new workshop management operations and pagination behavior, ensuring reliable functionality.

### DIFF
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
@@ -53,7 +53,6 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
     private readonly ITagService tagService;
     private readonly IContactsService<Workshop, IHasContactsDto<Workshop>> contactsService;
     private readonly IApplicationRepository applicationRepository;
-    private readonly IEntityRepositorySoftDeleted<Guid, StudySubject> studySubjectRepository;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="WorkshopService"/> class.
@@ -75,7 +74,6 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
     /// <param name="codeficatorService">Srvice for CATOTTG.</param>
     /// <param name="searchStringService">Service for handling the search string.</param>
     /// <param name="tagService">Service for Tag entity.</param>
-    /// <param name="studySubjectRepository">Repository for StudySubject entity.</param>
     public WorkshopService(
         IWorkshopRepository workshopRepository,
         IEntityRepository<long, Tag> tagRepository,
@@ -94,8 +92,7 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
         ITagService tagService,
         ISearchStringService searchStringService,
         IContactsService<Workshop, IHasContactsDto<Workshop>> contactsService,
-        IApplicationRepository applicationRepository,
-        IEntityRepositorySoftDeleted<Guid, StudySubject> studySubjectRepository)
+        IApplicationRepository applicationRepository)
     {
         this.workshopRepository = workshopRepository;
         this.tagRepository = tagRepository;
@@ -115,7 +112,6 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
         this.tagService = tagService;
         this.contactsService = contactsService;
         this.applicationRepository = applicationRepository;
-        this.studySubjectRepository = studySubjectRepository;
     }
 
     /// <inheritdoc/>
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
@@ -12,6 +12,7 @@ using OutOfSchool.BusinessLogic.Services.AverageRatings;
 using OutOfSchool.BusinessLogic.Services.SearchString;
 using OutOfSchool.BusinessLogic.Services.Workshops;
 using OutOfSchool.Common.Enums;
+using OutOfSchool.Common.Models;
 using OutOfSchool.Services.Enums;
 using OutOfSchool.Services.Models.Images;
 using OutOfSchool.Services.Repository.Api;
@@ -52,7 +53,6 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
     private readonly ITagService tagService;
     private readonly IContactsService<Workshop, IHasContactsDto<Workshop>> contactsService;
     private readonly IApplicationRepository applicationRepository;
-    private readonly IEntityRepositorySoftDeleted<Guid, StudySubject> studySubjectRepository;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="WorkshopService"/> class.
@@ -74,7 +74,6 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
     /// <param name="codeficatorService">Srvice for CATOTTG.</param>
     /// <param name="searchStringService">Service for handling the search string.</param>
     /// <param name="tagService">Service for Tag entity.</param>
-    /// <param name="studySubjectRepository">Repository for StudySubject entity.</param>
     public WorkshopService(
         IWorkshopRepository workshopRepository,
         IEntityRepository<long, Tag> tagRepository,
@@ -93,8 +92,7 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
         ITagService tagService,
         ISearchStringService searchStringService,
         IContactsService<Workshop, IHasContactsDto<Workshop>> contactsService,
-        IApplicationRepository applicationRepository,
-        IEntityRepositorySoftDeleted<Guid, StudySubject> studySubjectRepository)
+        IApplicationRepository applicationRepository)
     {
         this.workshopRepository = workshopRepository;
         this.tagRepository = tagRepository;
@@ -114,7 +112,6 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
         this.tagService = tagService;
         this.contactsService = contactsService;
         this.applicationRepository = applicationRepository;
-        this.studySubjectRepository = studySubjectRepository;
     }
 
     /// <inheritdoc/>
@@ -340,45 +337,27 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
     }
 
     /// <inheritdoc/>
-    public async Task<Result<IEnumerable<WorkshopAttachmentStatusDto>>> GetWorkshopsWithAttachmentStatusByProviderId(
+    public Task<PaginatedResult<WorkshopAttachmentStatusDto>> GetAttachedWorkshops(
            Guid studySubjectId,
-           Guid providerId)
+           Guid providerId,
+           int page,
+           int pageSize)
     {
-        // Get all workshops for the provider
-        var allWorkshops = await workshopRepository
-            .GetByFilter(whereExpression: x => x.ProviderId == providerId)
-            .ConfigureAwait(false);
+        logger.LogInformation("Getting workshops with attachment status. ProviderId = {ProviderId}, " +
+                              "StudySubjectId = {StudySubjectId}, Page = {Page}, PageSize = {PageSize}",
+                               providerId, studySubjectId, page, pageSize);
 
-        // Get the study subject with the specified Id and include the workshops
-        var studySubject = await studySubjectRepository
-            .GetByIdWithDetails(studySubjectId, includeProperties: "Workshops")
-            .ConfigureAwait(false);
-
-        if (studySubject == null)
-        {
-            logger.LogWarning("StudySubject with Id = {StudySubjectId} was not found", studySubjectId);
-            return Result<IEnumerable<WorkshopAttachmentStatusDto>>.Failed(new OperationError
+        var query = workshopRepository
+            .GetByFilterNoTracking(whereExpression: w => w.ProviderId == providerId)
+            .Select(w => new WorkshopAttachmentStatusDto
             {
-                Code = "404",
-                Description = $"StudySubject with Id = {studySubjectId} was not found"
-            });
-        }
-
-        // Get the IDs of the already attached workshops
-        var attachedWorkshopIds = studySubject.Workshops.Select(w => w.Id).ToList();
-
-        // Map the workshops to the DTOs and set the IsAttached property
-        var result = allWorkshops
-            .Select(w =>
-            {
-                var dto = mapper.Map<WorkshopAttachmentStatusDto>(w);
-                dto.IsAttached = attachedWorkshopIds.Contains(w.Id);
-                return dto;
+                Id = w.Id,
+                Title = w.Title,
+                IsAttached = w.StudySubjects.Any(ss => ss.Id == studySubjectId)
             })
-            .OrderBy(w => w.Title)
-            .ToList();
+            .OrderBy(w => w.Title);
 
-        return Result<IEnumerable<WorkshopAttachmentStatusDto>>.Success(result);
+        return query.ToPaginatedResultAsync(page, pageSize);
     }
 
     /// <inheritdoc/>
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
@@ -52,6 +52,7 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
     private readonly ITagService tagService;
     private readonly IContactsService<Workshop, IHasContactsDto<Workshop>> contactsService;
     private readonly IApplicationRepository applicationRepository;
+    private readonly IEntityRepositorySoftDeleted<Guid, StudySubject> studySubjectRepository;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="WorkshopService"/> class.
@@ -73,6 +74,7 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
     /// <param name="codeficatorService">Srvice for CATOTTG.</param>
     /// <param name="searchStringService">Service for handling the search string.</param>
     /// <param name="tagService">Service for Tag entity.</param>
+    /// <param name="studySubjectRepository">Repository for StudySubject entity.</param>
     public WorkshopService(
         IWorkshopRepository workshopRepository,
         IEntityRepository<long, Tag> tagRepository,
@@ -91,7 +93,8 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
         ITagService tagService,
         ISearchStringService searchStringService,
         IContactsService<Workshop, IHasContactsDto<Workshop>> contactsService,
-        IApplicationRepository applicationRepository)
+        IApplicationRepository applicationRepository,
+        IEntityRepositorySoftDeleted<Guid, StudySubject> studySubjectRepository)
     {
         this.workshopRepository = workshopRepository;
         this.tagRepository = tagRepository;
@@ -111,6 +114,7 @@ public class WorkshopService : IWorkshopService, ISensitiveWorkshopsService
         this.tagService = tagService;
         this.contactsService = contactsService;
         this.applicationRepository = applicationRepository;
+        this.studySubjectRepository = studySubjectRepository;
     }
 
     /// <inheritdoc/>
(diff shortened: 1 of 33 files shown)


OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -94,7 +96,8 @@ public WorkshopService(
         ITagService tagService,
         ISearchStringService searchStringService,
         IContactsService<Workshop, IHasContactsDto<Workshop>> contactsService,
-        IApplicationRepository applicationRepository)
+        IApplicationRepository applicationRepository,
+        IEntityRepositorySoftDeleted<Guid, StudySubject> studySubjectRepository)
```
DmyMi: We have special case for `GUID` keys - `ISensitiveEntityRepositorySoftDeleted`

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -356,6 +360,48 @@ public async Task<SearchResult<WorkshopProviderViewCard>> GetByProviderId(Guid i
...
+           Guid studySubjectId,
+           Guid providerId)
+    {
+        // Get all workshops for the provider
+        var allWorkshops = await workshopRepository
+            .GetByFilter(whereExpression: x => x.ProviderId == providerId)
+            .ConfigureAwait(false);
+
+        // Get the study subject with the specified Id and include the workshops
+        var studySubject = await studySubjectRepository
```
DmyMi: even with pagination, maybe a `left join` would be better? maybe as a little optional subtask try to test what runs faster.

Even if a bit slower - still will make the entire method fit into like 5 lines instead of making multiple array enumerations

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -263,6 +272,150 @@ private async Task CheckIfLanguageIdIsCorrect(StudySubjectCreateUpdateDto dto)
...
+            return NotFoundResult<StudySubjectDto>(providerId);
+        }
+
+        // Check if all workshops to attach were found
+        var existingWorkshopIds = studySubject.Workshops.Select(w => w.Id).ToList();
+
+        foreach (var workshopDto in workshopsWithStatus)
+        {
+            var workshop = allWorkshops.FirstOrDefault(w => w.Id == workshopDto.Id);
+            if (workshop == null) continue; // Skip if workshop was not found
```
DmyMi: maybe log a warning or info that user passed non existing workshop into attachment.

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -263,6 +272,150 @@ private async Task CheckIfLanguageIdIsCorrect(StudySubjectCreateUpdateDto dto)
...
+        }
+    }
+
+    /// <summary>
+    /// Returns a standardized "Not Found" result for an entity.
+    /// </summary>
+    /// <typeparam name="T">The type of the entity that was not found.</typeparam>
+    /// <param name="id">The unique identifier of the missing entity.</param>
+    /// <returns>A failed <see cref="Result{T}"/> indicating that the entity was not found.</returns>
+    private Result<T> NotFoundResult<T>(Guid id)
```
DmyMi: it might be useful not only here, you can extract this helper method to Util

OutOfSchool/OutOfSchool.BusinessLogic/Util/MappingProfile.cs
```diff
@@ -863,10 +872,14 @@ public MappingProfile()
...
-                opt => opt.MapFrom(src => src.Language.Id));
+                opt => opt.MapFrom(src => src.Language.Id))
+            .ForMember(dest => dest.ProviderId, opt => opt.Ignore())
+            .ForMember(dest => dest.Provider, opt => opt.Ignore())
+            .ForMember(dest => dest.Workshops, opt => opt.Ignore());
+
 
         CreateMap<StudySubject, StudySubjectDto>()
-            .ForMember(dest => dest.WorkshopId, opt => opt.Ignore());
+            .ForMember(dest => dest.Workshops, opt => opt.MapFrom(src => src.Workshops));
```
DmyMi, line 885: Haven't seen get method changes, but double check that the `src.Workshops` is populated by eager loading and not using N+1 query during mapping.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -356,6 +360,48 @@ public async Task<SearchResult<WorkshopProviderViewCard>> GetByProviderId(Guid i
         return result;
     }
 
+    /// <inheritdoc/>
+    public async Task<Result<IEnumerable<WorkshopAttachmentStatusDto>>> GetWorkshopsWithAttachmentStatusByProviderId(
+           Guid studySubjectId,
+           Guid providerId)
+    {
+        // Get all workshops for the provider
+        var allWorkshops = await workshopRepository
```
DmyMi: There may be 100s of workshops per provider and workshops can have many subjects according to configuration. Maybe introduce pagination that will allow to fetch batches of data.

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/ProviderConfiguration.cs
```diff
@@ -89,5 +89,10 @@ public void Configure(EntityTypeBuilder<Provider> builder)
 
         builder.Property(x => x.UpdatedAt)
                 .ValueGeneratedOnAddOrUpdate();
+
+        builder.HasMany(p => p.StudySubjects)
```
DmyMi: is the duplicate configuration needed as the same is in `StudySubjectConfig`?

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/StudySubjectConfiguration.cs
```diff
@@ -24,5 +24,13 @@ public override void Configure(EntityTypeBuilder<StudySubject> builder)
             .WithMany()
             .HasForeignKey(x => x.LanguageId)
             .OnDelete(DeleteBehavior.Restrict);
+
+        builder.HasMany(x => x.Workshops)
+            .WithMany(x => x.StudySubjects);
+
+        builder.HasOne(x => x.Provider)
```
DmyMi, line 34: check duplicate comments.

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/WorkshopConfiguration.cs
```diff
@@ -40,5 +40,8 @@ public override void Configure(EntityTypeBuilder<Workshop> builder)
 
         builder.HasMany(x => x.Tags)
             .WithMany(x => x.Workshops);
+
+        builder.HasMany(x => x.StudySubjects)
```
DmyMi: is the duplicate configuration needed as the same is in `StudySubjectConfig`?

OutOfSchool/OutOfSchool.DataAccess/Models/Provider.cs
```diff
@@ -159,4 +159,5 @@ public class Provider : IKeyedEntity<Guid>, IImageDependentEntity<Provider>, ISo
     public DateTime UpdatedAt { get; set; }
 
     public virtual ICollection<Position> Positions { get; set; }
+    public virtual List<StudySubject> StudySubjects { get; set; }
```
DmyMi: I haven't seen any logic using provider as main point of interaction with subjects (i may be blind though 😅 ). So for now this navigation property is not needed

OutOfSchool/OutOfSchool.DataAccess/Models/Workshop.cs
```diff
@@ -152,8 +152,10 @@ public virtual ICollection<Workshop>
     public virtual List<Image<Workshop>> Images { get; set; }
 
     public virtual List<Tag> Tags { get; set; }
+
+    public virtual List<StudySubject> StudySubjects { get; set; }
```
DmyMi, line 148: Same as with providers, i may be blind 😅 but most interaction goes from study subjects to workshops. So this may not be required

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopController.cs
```diff
@@ -109,6 +109,44 @@ public async Task<IActionResult> GetWorkshopListByProviderId(Guid providerId)
...
+    /// A task that represents the asynchronous operation. The task result contains a list of <see cref="ShortEntityDto"/> 
+    /// representing the workshops associated with the provider.
+    /// </returns>
+    [Authorize]
+    [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(IEnumerable<WorkshopAttachmentStatusDto>))]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpGet("{studySubjectId}/{providerId}")]
```
DmyMi: need to add something with path. As it should look like the following:
`/providers/{providerId}/studysubject/{studySubjectId}/workshops` (or `/workshops/attached` in the end).

Please think about the RESTful naming

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopController.cs
```diff
@@ -109,6 +109,44 @@ public async Task<IActionResult> GetWorkshopListByProviderId(Guid providerId)
...
+    public async Task<IActionResult> GetWorkshopsWithAttachmentStatusByProviderId(Guid studySubjectId, Guid providerId)
```
DmyMi: Think about adding pagination

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/StudySubjectController.cs
```diff
@@ -210,6 +210,71 @@ public async Task<IActionResult> Delete(Guid id, Guid providerId)
...
+    /// the success or failure of the operation, along with the updated study subject data if successful.
+    /// </returns>
+    [Authorize]
+    [HasPermission(Permissions.WorkshopEdit)]
+    [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(StudySubjectDto))]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpPut("{studySubjectId}")]
```
DmyMi: lets make new endpoints REST friendly :) think about RESTful naming for this and another new endpoint

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
```diff
@@ -263,6 +272,150 @@ private async Task CheckIfLanguageIdIsCorrect(StudySubjectCreateUpdateDto dto)
         }
     }
 
+    /// <inheritdoc/>
+    public async Task<Result<StudySubjectDto>> UpdateWorkshopsForStudySubject(
+        Guid studySubjectId,
+        Guid providerId,
+        IEnumerable<WorkshopAttachmentStatusDto> workshopsWithStatus)
+    {
+        await providerService.HasProviderRights(providerId).ConfigureAwait(false);
```
DmyMi: Also, do not use `providerService.HasProviderRights`, this method will be removed. Use `ICurrentUserService` directly

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
max-khakimzhanov: The `studySubjectRepository` is no longer used in this service — it was removed along with related logic during recent refactoring. I've already deleted it from the constructor.

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
max-khakimzhanov: I've refactored the method to build WorkshopAttachmentStatusDto directly in the LINQ query.
This combines filtering, sorting, and checking attachment status in one step and avoids extra in-memory processing.

I also tested both versions — the updated one is faster.
The previous approach loaded unnecessary data and did more work in memory, while the new one runs as a single efficient SQL query

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
max-khakimzhanov: ✅

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
max-khakimzhanov: ✅

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
max-khakimzhanov: Pagination has been added to the method and API endpoint as suggested.
The method now supports page and pageSize parameters, and returns a PaginatedResult<T> to avoid loading large datasets at once

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/ProviderConfiguration.cs
max-khakimzhanov: Thanks for pointing that out! I've already removed the duplicate configuration

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/StudySubjectConfiguration.cs
max-khakimzhanov: ✅

OutOfSchool/OutOfSchool.DataAccess/Models/Configurations/WorkshopConfiguration.cs
max-khakimzhanov: This has already been deleted

OutOfSchool/OutOfSchool.DataAccess/Models/Provider.cs
max-khakimzhanov: You're right — there was no actual usage of that : ) 
I've removed it

OutOfSchool/OutOfSchool.DataAccess/Models/Workshop.cs
max-khakimzhanov: Initially, it wasn’t used, but now this navigation property is needed. It’s used in WorkshopService.GetAttachedWorkshops to determine whether each workshop is attached to the given study subject

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopController.cs
max-khakimzhanov: ✅

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/WorkshopController.cs
max-khakimzhanov: Pagination has been added to this endpoint.
It returns a PaginatedResult<WorkshopAttachmentStatusDto> and accepts page and pageSize as query parameters.
Also added the [ValidatePagination] attribute to validate the query parameters

OutOfSchool/OutOfSchool.WebApi/Controllers/V1/StudySubjectController.cs
max-khakimzhanov: ✅

OutOfSchool/OutOfSchool.BusinessLogic/Services/StudySubjectService.cs
max-khakimzhanov: ✅

OutOfSchool/OutOfSchool.BusinessLogic/Util/MappingProfile.cs
max-khakimzhanov: I've checked this — `StudySubject.Workshops` is eager-loaded using `.GetByIdWithDetails(..., includeProperties: "Workshops")` before mapping.
So AutoMapper uses fully loaded data and no additional queries are made

OutOfSchool/OutOfSchool.WebApi/Util/ValidatePaginationAttribute.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            context.Result = new BadRequestObjectResult("Page number must be greater than 0.");
+            return;
+        }
+
+        if (args.TryGetValue("pageSize", out var pageSizeObj) &&
+            pageSizeObj is int pageSize && pageSize < 1)
+        {
+            context.Result = new BadRequestObjectResult("Page size must be greater than 0.");
+            return;
```
max-khakimzhanov: this redundant return will be deleted

OutOfSchool/OutOfSchool.Migrations/Data/Migrations/OutOfSchoolMigrations/20250409143340_AddStudySubjectRelations.cs
```diff
@@ -0,0 +1,83 @@
...
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<Guid>(
+                name: "ProviderId",
+                table: "StudySubjects",
+                type: "UUID(16)",
+                nullable: false,
+                defaultValue: new Guid("00000000-0000-0000-0000-000000000000"));
```
DmyMi: it is a `NOT NULL` field that should never have a default property. I have not seen any migration before creating a default value for a foreign key. Yes it would fail a foreign key constraint most of the time, but there might be an error and some one adds a provider with empty Guid and this default value will bind to that provider

OutOfSchool/OutOfSchool.BusinessLogic/Services/Database/WorkshopService.cs
```diff
@@ -335,6 +336,30 @@ public async Task<SearchResult<WorkshopProviderViewCard>> GetByProviderId(Guid i
...
     }
 
+    /// <inheritdoc/>
+    public Task<PaginatedResult<WorkshopAttachmentStatusDto>> GetAttachedWorkshops(
+           Guid studySubjectId,
+           Guid providerId,
+           int page,
+           int pageSize)
+    {
+        logger.LogInformation("Getting workshops with attachment status. ProviderId = {ProviderId}, " +
```
DmyMi: This log should be `Debug`

OutOfSchool/OutOfSchool.DataAccess/Models/Workshop.cs
DmyMi: As it is a join anyway (we need to take workshops and inner join to studysubjects) under the hood you can "reverse" it by taking the studysubjects and inner joining workshops.

And business logic wise this is more related to study subjects, so can be moved to studysubjects service (even the method args do not have workshops).

DmyMi (approved):

PR #1766, opened 2025-04-02
feat(workshop): add NoAgeRestrictions property and update age validation

### LINKED ISSUE #1741: Add "no limit" to workshop age
For "Вік учасників" add a check box "Гурток не має вікових обмежень". If it is checked - disable the age input fields.

Backend should correctly handle this use case. Possible approaches can include, but not limited to:
- Make fields not required and/or send default values (0, 120) from frontend
- Add this flag to DTO so we can correctly set default values (0, 120) into entity and know this was not a mistake
- Any better solution you can think of :)

Image

synapsetron:
Introduce NoAgeRestrictions boolean to WorkshopMainRequiredPropertiesDto and WorkshopBaseDto. Make MinAge and MaxAge nullable and conditionally required. Update Validate method to handle NoAgeRestrictions logic. Ignore NoAgeRestrictions in WorkshopDraftMappingProfile and MappingProfile. Add unit tests for new validation logic.
Minor changes in ApplicationControllerTests and ChildControllerTests.

## Summary by CodeRabbit

- **New Features**
  - Workshops now offer flexible age restriction settings, including a new option to indicate no age restrictions, with automatic default boundaries applied when applicable.

- **Tests**
  - A new test class has been introduced to validate the behavior of the workshop model concerning age restrictions, ensuring robust validation of workshop configurations.


OutOfSchool/OutOfSchool.BusinessLogic/Util/MappingProfile.cs
```diff
@@ -187,7 +188,8 @@ public MappingProfile()
             .ForMember(dest => dest.Facebook,
                 opt => opt.MapFrom(src => src.Contacts.FirstOrDefault(c => c.IsDefault).SocialNetworks.FirstOrDefault(s => s.Type == SocialNetworkContactType.Facebook).Url))
             .ForMember(dest => dest.Instagram,
-                opt => opt.MapFrom(src => src.Contacts.FirstOrDefault(c => c.IsDefault).SocialNetworks.FirstOrDefault(s => s.Type == SocialNetworkContactType.Instagram).Url));
+                opt => opt.MapFrom(src => src.Contacts.FirstOrDefault(c => c.IsDefault).SocialNetworks.FirstOrDefault(s => s.Type == SocialNetworkContactType.Instagram).Url))
+            .ForMember(dest => dest.NoAgeRestrictions, opt => opt.Ignore());
```
DmyMi: In hindsight, we probably should map it :) with logic like `if minage = 0 and maxage = 120 then true else false`

OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/TempSave/WorkshopMainRequiredPropertiesDto.cs
```diff
@@ -28,13 +29,16 @@ public class WorkshopMainRequiredPropertiesDto : IValidatableObject
     [MaxLength(Constants.MaxWorkshopShortTitleLength)]
     public string ShortTitle { get; set; } = string.Empty;
 
-    [Required(ErrorMessage = "Children's min age is required")]
-    [Range(0, 120, ErrorMessage = "Min age should be a number from 0 to 120")]
-    public int MinAge { get; set; }
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
```
DmyMi: In hindsight, we probably should return it 😅

OutOfSchool/OutOfSchool.BusinessLogic/Util/MappingProfile.cs
synapsetron: Resolved

OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/TempSave/WorkshopMainRequiredPropertiesDto.cs
synapsetron: done

PR #1849, opened 2025-05-27
Update workshop draft filter validation logic and Moderator/TechAdmin authorization

max-khakimzhanov:
This PR enhances the workshop draft filtering system by updating validation logic and strengthening moderation checks to improve content quality and data integrity. Additionally, it introduces comprehensive authorization checks for Moderator and TechAdmin roles to ensure users exist in the corresponding database tables before granting access.

### Enhanced Authorization System
**Updated CurrentUserService with robust role-based security:**
- **Added new dependencies**: `ISensitiveEntityRepositorySoftDeleted<Moderator>` and `ISensitiveEntityRepositorySoftDeleted<TechAdmin>` repositories
- **Updated private methods**: `ModeratorHasRights()` and `TechAdminHasRights()` for database validation
- **Updated UserHasRights logic**: Added support for `ModeratorRights` and `TechAdminRights` in the authorization pipeline
- **Enhanced security**: Users are now automatically logged out if removed from Moderator/TechAdmin tables, even before token expiration

**New Authorization Flow:**
1. **Role validation**: Check if user has the required role claim
2. **Database validation**: Verify user exists in the corresponding table via `Individual.UserId` lookup
3. **Audit logging**: Log unauthorized access attempts when `AccessLogEnabled` is true
4. **Graceful handling**: Return false for missing users without exposing system details

### Workshop Draft Filter Validation
- **Updated validation rules** for workshop draft filtering

### Moderation System Improvements
- **Enhanced content moderation checks** for workshop drafts before publication

## Summary by CodeRabbit

- **New Features**
  - Moderation actions on workshop drafts now support multiple statuses, allowing for broader moderation scenarios.
- **Refactor**
  - Moderator and tech admin permissions are now determined by user context rather than requiring explicit user IDs.
  - Workshop draft moderation endpoints and related services have simplified method signatures, no longer requiring moderator IDs.
- **Bug Fixes**
  - Enhanced validation for draft statuses, ensuring only allowed statuses are processed during moderation.
- **Tests**
  - Updated and added tests to reflect changes in permission checks and method signatures for moderation actions.

### DIFF
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Models/WorkshopDraft/WorkshopDraftFilterAdministration.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Models/WorkshopDraft/WorkshopDraftFilterAdministration.cs
@@ -1,12 +1,25 @@
 ﻿using OutOfSchool.BusinessLogic.Models.Workshops;
 using OutOfSchool.BusinessLogic.Util.CustomValidation;
 using OutOfSchool.Services.Enums.WorkshopStatus;
-using System.ComponentModel.DataAnnotations;
 
 namespace OutOfSchool.BusinessLogic.Models.WorkshopDraft;
 public class WorkshopDraftFilterAdministration : WorkshopFilterAdministration
 {
-    [EnumDataType(typeof(WorkshopDraftStatus), ErrorMessage = Constants.EnumErrorMessage)]
+    private HashSet<WorkshopDraftStatus> workshopDraftStatuses;
+
     [WorkshopDraftStatusValidation]
-    public WorkshopDraftStatus WorkshopDraftStatus { get; set; } = WorkshopDraftStatus.PendingModeration;
+    public HashSet<WorkshopDraftStatus> WorkshopDraftStatuses
+    {
+        get => workshopDraftStatuses ?? GetDefaultStatuses();
+        set => workshopDraftStatuses = value;
+    }
+
+    private static HashSet<WorkshopDraftStatus> GetDefaultStatuses()
+    {
+        return
+        [
+            WorkshopDraftStatus.PendingModeration,
+            WorkshopDraftStatus.EditedByModerator
+        ];
+    }
 }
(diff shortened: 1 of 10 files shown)


OutOfSchool/OutOfSchool.BusinessLogic/Services/CurrentUserService.cs
```diff
@@ -295,43 +297,38 @@ private async Task<bool> EmployeeWorkshopRights(Guid workshopId)
...
+    private async Task<bool> ModeratorHasRights()
     {
         if (!IsModerator())
         {
-            return Task.FromResult(false);
+            return false;
         }
 
-        var result = UserId == moderatorId.ToString();
+        var result = await moderatorRepository.Any(m => m.Individual.UserId == UserId);
```
DmyMi, line 307: Technically, moderator/techadmin `Id` is both a primary key to moderator/techadmin as well as foreign key to individual.
As this is not the issue now, but we can add individual `Id` to token during auth and just do the `Any(m => m.Id = <claim_from_toekn>)`

OutOfSchool/OutOfSchool.BusinessLogic/Models/WorkshopDraft/WorkshopDraftFilterAdministration.cs
```diff
@@ -1,12 +1,25 @@
 ﻿using OutOfSchool.BusinessLogic.Models.Workshops;
 using OutOfSchool.BusinessLogic.Util.CustomValidation;
 using OutOfSchool.Services.Enums.WorkshopStatus;
-using System.ComponentModel.DataAnnotations;
 
 namespace OutOfSchool.BusinessLogic.Models.WorkshopDraft;
 public class WorkshopDraftFilterAdministration : WorkshopFilterAdministration
 {
-    [EnumDataType(typeof(WorkshopDraftStatus), ErrorMessage = Constants.EnumErrorMessage)]
```
DmyMi, line 9: need to think about how to do this check for enum datatype, because it is possible in some cases to pass an array of random integers and this will deserialize into broken enums.

DmyMi (approved):

PR #1859, opened 2025-06-14
fix(filters): add default value to LanguageOfEducationId

### LINKED ISSUE #1858: Bug(api): 400 Bad Request when LanguageOfEducationId is missing in Workshop/GetByFilter request
### **Title:**
**400 Bad Request when `LanguageOfEducationId` parameter is missing in `Workshop/GetByFilter` request**

---

### **Environment:**  
**OS:** Windows 11  
**Browser:** Google Chrome 125.0.6422.113 (Official Build) (64-bit)  
**Device:** PC  

---

### **Reproducible:**  
**Always**

---

### **Preconditions:**  
- Backend project is running locally on `localhost:5000`  
- Swagger UI is accessible  
- Endpoint `GET /api/v1/Workshop/GetByFilter` is available

---

### **Steps to reproduce:**  
1. Open Swagger UI: `http://localhost:5000/swagger/index.html`  
2. Navigate to the endpoint: `GET /api/v1/Workshop/GetByFilter`  
3. Leave the `LanguageOfEducationId` parameter empty (do not provide a value)  
4. Set `isAdmins = false`  
5. Click the **Execute** button  

---

### **Actual result:**  
The server returns a **400 Bad Request** with the following response body:
```json
{
  "type": "https://tools.ietf.org/html/rfc9110#section-15.5.1",
  "title": "One or more validation errors occurred.",
  "status": 400,
  "errors": {
    "LanguageOfEducationId": [
      "LanguageId value must be a positive number and greater than 0"
    ]
  },
  "traceId": "00-e00af493f320025bbfdf1d38ccf56c5a-2698549852fd7128-00"
}

synapsetron:
### **Purpose**
This pull request fixes a bug where the `LanguageOfEducationId` query parameter in the `Workshop/GetByFilter` endpoint caused a 400 Bad Request response when it was not provided.

---

### **Changes**
- Preserved the `[Range(1, long.MaxValue)]` validation to ensure that when the value is provided, it must be greater than 0
- Adjusted model binding logic to handle the absence of this parameter gracefully

---

### **Reason**
The original implementation used a non-nullable `long`, which defaulted to `0` when the parameter was missing. Since `0` does not meet the `[Range]` validation constraint, the request would fail with a 400 error, even when the parameter was not required.

---

### **Сlosses: #1858**

---

### **How to test**
1. Run the backend
2. Use Swagger or Postman to send a GET request to `/api/v1/Workshop/GetByFilter` **without** the `LanguageOfEducationId` parameter
3. Ensure the response is **200 OK** (or the correct filtered data based on other parameters)
---

## Summary by CodeRabbit

- **Refactor**
  - Updated default selection for language of education in workshop filters to ensure a value is always set.
- **Bug Fixes**
  - Adjusted workshop filtering to correctly return no results for certain learning form queries.

### DIFF
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/WorkshopFilter.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Models/Workshops/WorkshopFilter.cs
@@ -57,8 +57,8 @@ public class WorkshopFilter : OffsetFilter
 
     public bool IsStrictWorkdays { get; set; } = false;
 
-    [Range(1,long.MaxValue, ErrorMessage = "LanguageId value must be a possitive number and greater than 0")]
-    public long LanguageOfEducationId { get; set; }
+    [Range(1, long.MaxValue, ErrorMessage = "LanguageId value must be a possitive number and greater than 0")]
+    public long LanguageOfEducationId { get; set; } = 1;
 
     public long CATOTTGId { get; set; } = default;
 
--- a/OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/WorkshopServiceDBTests.cs
+++ b/OutOfSchool/OutOfSchool.WebApi.Tests/Services/Database/WorkshopServiceDBTests.cs
@@ -136,7 +136,7 @@ public class WorkshopServiceDBTests
         var result = await workshopService.GetByFilter(filter).ConfigureAwait(false);
 
         // Assert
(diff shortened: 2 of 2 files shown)


DmyMi:
@synapsetron I will close your PR and merge a similar fix :) as in BL layer we do check if `languageid > 0` everywhere - we can just remove the hard requirement for this field to be present in the query filter.

PR #1950, opened 2025-09-04
Edrpou for branches

### LINKED ISSUE #1945: Enable Provider Branch Organizations with Shared EDRPOU
## Context
We currently enforce EDRPOU uniqueness across all providers, but need to support organizational branches that share the same EDRPOU as their main organization while maintaining uniqueness for main organizations only.

**Current State:**
- `Provider.Edrpou` has unique index via EF configuration
- `Provider.IsStructuralUnit` exists but is unused for this purpose
- Using EF Core with Pomelo.EntityFrameworkCore.MySql for MariaDB

**Target State:**
- One main organization per EDRPOU (maintains uniqueness)
- Multiple branches per main organization (share same EDRPOU)
- Branches must link to their main provider

## Technical Approach

Use MariaDB's generated column feature to create a composite uniqueness key that allows EDRPOU sharing for branches.

**Pros:** Database-level guarantee, preserves EDRPOU on branches, minimal API changes
**Cons:** Requires MariaDB 10.2+ (virtual columns) or 10.3+ (persistent columns)

### Data Model Changes

**File:** `OutOfSchool.DataAccess/Models/Provider.cs`

Add self-referencing navigation properties:
```csharp
public Guid? ParentProviderId { get; set; }
public virtual Provider ParentProvider { get; set; }
public virtual ICollection<Provider> Branches { get; set; }
```

Keep `Edrpou` as required string (no nullability changes).

### EF Core Configuration Updates

**File:** `OutOfSchool.DataAccess/Models/Configurations/ProviderConfiguration.cs`

- Remove existing unique index: `builder.HasIndex(x => x.Edrpou).IsUnique();`
- Add self-reference configuration:
```csharp
builder.HasOne(p => p.ParentProvider)
    .WithMany(p => p.Branches)
    .HasForeignKey(p => p.ParentProviderId)
    .OnDelete(DeleteBehavior.Restrict);
```
- Add generated column with unique index
   - Investigate MariaDB `Virtual` vs `Persistent` generated columns
   - Try using `HasComputedColumnSql()` (if we choose `virtual`), but probably for our case we will need to use persistent
   - Create unique index on this column
   - If `HasComputedColumnSql()` won't work - fall back to custom SQL (maybe even in migration) like:
        ```
        ALTER TABLE `Providers` 
        ADD COLUMN `EdrpouUniqKey` VARCHAR(100) 
        GENERATED ALWAYS AS (
            CASE 
                WHEN `IsStructuralUnit` = 1 
                THEN CONCAT(`Edrpou`, '-', REPLACE(LOWER(`Id`), '-', ''))
                ELSE `Edrpou`
            END
        ) STORED
        ```

> [!NOTE]
> The generated column ensures main organizations (`IsStructuralUnit = false`) use plain EDRPOU for uniqueness, while branches (`IsStructuralUnit = true`) get EDRPOU concatenated with their unique ID.

### Repository Layer Updates

**File:** `OutOfSchool.DataAccess/Repository/ProviderRepository.cs`

Update uniqueness check for main organizations only:
```csharp
public bool SameExists(Provider entity);
```

Add method to validate parent-branch relationships:
```csharp
public async Task<bool> IsValidParentProvider(Guid parentId);
```

### 5. Business Logic Updates

**File:** `OutOfSchool.BusinessLogic/Services/ProviderServices/ProviderService.cs`

**For Main Organizations (`IsStructuralUnit = false`):**
- Continue existing EDRPOU uniqueness validation
- Ensure `ParentProviderId` is null
- Use existing `SameExists()` method

**For Branch Organizations (`IsStructuralUnit = true`):**
- Require valid `ParentProviderId`
- Automatically inherit parent's EDRPOU (ignore client input)
- Validate parent exists and is not a branch itself

### DTO Updates

**Files:** 
- `OutOfSchool.BusinessLogic/Models/Providers/ProviderBaseDto.cs`
- `OutOfSchool.BusinessLogic/Models/Providers/ProviderCreateDto.cs`
- `OutOfSchool.BusinessLogic/Models/Providers/ProviderUpdateDto.cs`

Add to base DTO:
```csharp
public Guid? ParentProviderId { get; set; }

[JsonIgnore] // Don't expose in API - server managed
public string ParentProviderName { get; set; } // For display only
```

**Validation Attributes:**
```csharp
// Custom validation attribute
public class ValidProviderHierarchyAttribute : ValidationAttribute
{
    public override bool IsValid(object value)
    {
        if (value is ProviderBaseDto dto)
        {
            // IsStructuralUnit = true requires ParentProviderId
            // IsStructuralUnit = false requires ParentProviderId = null
            return dto.IsStructuralUnit ? dto.ParentProviderId.HasValue : !dto.ParentProviderId.HasValue;
        }
        return true;
    }
}
```

### API Impact

**EDRPOU Handling:**
- Remove EDRPOU from create/update DTOs
- Server-side assignment based on hierarchy (main: user input, branch: inherit from parent)
- Prevent EDRPOU changes via public APIs

**New Endpoints (Optional):**
```csharp
[HttpGet("{providerId}/branches")]
public async Task<ActionResult<IEnumerable<ProviderDto>>> GetBranches(Guid providerId)

[HttpGet("{providerId}/parent")]
public async Task<ActionResult<ProviderDto>> GetParentProvider(Guid providerId)
```

### Query Impact

**Minimal Changes Required:**
- Existing EDRPOU searches (`x.Edrpou.Contains(...)`) work unchanged
- Branch providers appear in searches with same EDRPOU as main
- Provider lists may need hierarchy indicators in UI

**Search Enhancements:**
```csharp
// Option to search main organizations only
.Where(p => !p.IsStructuralUnit)

// Include hierarchy information in results
.Include(p => p.ParentProvider)
.Include(p => p.Branches)
```

### Error Handling

**New Exception Scenarios:**
- Attempting to create main with duplicate EDRPOU
- Creating branch without valid parent
- Circular reference in provider hierarchy
- Database constraint violations

annygumeniuk:
## Summary by CodeRabbit

- New Features
  - Provider hierarchy support: main organizations can have branches; branches reference a parent.
  - New endpoints to fetch a provider’s branches and its parent.
  - Provider create/update and responses include optional parent provider info for display.

- Validation
  - New hierarchy validation rules (branches must have a parent; main organizations cannot).

- Chores
  - Uniqueness handling for identifiers adjusted to support hierarchy.
  - Authorization checks refined for updates involving parent providers.

- Tests
  - New tests covering branch/parent retrieval and controller/service behaviors.

### DIFF
--- a/OutOfSchool/OutOfSchool.BusinessLogic/Services/ProviderServices/ProviderService.cs
+++ b/OutOfSchool/OutOfSchool.BusinessLogic/Services/ProviderServices/ProviderService.cs
@@ -233,6 +233,33 @@ public class ProviderService(
     }
 
     /// <inheritdoc/>
+    public async Task<IEnumerable<ProviderDto>> GetBranchesAsync(Guid providerId)
+    {
+        var branches = await providerRepository.GetByFilter(
+            p => p.ParentProviderId == providerId,
+            includeExpression: q => q.Include(p => p.ParentProvider)
+        );
+
+        return branches.ToDto();
+    }
+    
+    /// <inheritdoc/>
+    public async Task<ProviderDto> GetParentProviderAsync(Guid providerId)
+    {
+        var providerList = await providerRepository.GetByFilter(
+            p => p.Id == providerId,
+            includeExpression: q => q.Include(p => p.ParentProvider)
+        );
+
+        var provider = providerList.FirstOrDefault();
+
+        if (provider?.ParentProvider == null)
+            return null;
+
+        return provider.ParentProvider.ToDto();
+    }
+
+    /// <inheritdoc/>
     public async Task<ProviderDto> Update(ProviderUpdateDto providerUpdateDto, string userId)
         => await UpdateProviderWithActionBeforeSavingChanges(providerUpdateDto, userId).ConfigureAwait(false);
 
@@ -543,18 +570,13 @@ public class ProviderService(
         var providerDomainModel = providerDto.ToModel();
 
         contactsService.PrepareNewContacts(providerDomainModel, providerDto);
-
-        // BUG: concurrency issue:
-        //      while first repository with this particular user id is not saved to DB - we can create any number of repositories for this user.
-        if (providerRepository.SameExists(providerDomainModel))
-        {
-            throw new InvalidOperationException(localizer["There is already a provider with such a data"]);
-        }
-
+     
         providerDomainModel.Status = ProviderStatus.Pending;
         providerDomainModel.LicenseStatus = providerDomainModel.License == null
             ? ProviderLicenseStatus.NotProvided
             : ProviderLicenseStatus.Pending;
+
+        await ApplyHierarchyRulesAsync(providerDomainModel);
 
         var newProvider = await providerRepository.Create(providerDomainModel).ConfigureAwait(false);
 
@@ -587,14 +609,26 @@ public class ProviderService(
         logger.LogDebug("Updating Provider with Id = {Id} was started", providerUpdateDto.Id);
 
         try
-        {
-            await currentUserService.UserHasRights(new ProviderRights(providerUpdateDto.Id), new DeputyDirectorRights(providerUpdateDto.Id)).ConfigureAwait(false);
-
+        {   if (providerUpdateDto.ParentProviderId != null)
+            {
+                await currentUserService.UserHasRights(
+                    new ProviderRights(providerUpdateDto.ParentProviderId.Value)).ConfigureAwait(false);
+            }
+            else
+            {
+                await currentUserService.UserHasRights(
+                    new ProviderRights(providerUpdateDto.Id), new DeputyDirectorRights(providerUpdateDto.Id)).ConfigureAwait(false);
+            }            
+            
             var checkProvider = await providerRepository.GetWithNavigations(providerUpdateDto.Id).ConfigureAwait(false);
-
+           
+            await providerRepository.IsValidParentProvider(providerUpdateDto.Id);
+            
             ChangeProviderStatusIfNeeded(providerUpdateDto, checkProvider, out var statusChanged, out var licenseChanged);
 
             contactsService.PrepareUpdatedContacts(checkProvider, providerUpdateDto);
+
+            await ApplyHierarchyRulesAsync(checkProvider, checkProvider.Id);
 
             if (IsNeedInRelatedWorkshopsUpdating(providerUpdateDto, checkProvider))
             {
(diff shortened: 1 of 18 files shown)


OutOfSchool/OutOfSchool.BusinessLogic/Services/ProviderServices/ProviderService.cs
```diff
@@ -587,15 +609,27 @@ private protected async Task<ProviderDto> UpdateProviderWithActionBeforeSavingCh
...
+            else
+            {
+                await currentUserService.UserHasRights(
+                    new ProviderRights(providerUpdateDto.Id), new DeputyDirectorRights(providerUpdateDto.Id)).ConfigureAwait(false);
+            }            
+            
             var checkProvider = await providerRepository.GetWithNavigations(providerUpdateDto.Id).ConfigureAwait(false);
-
+           
+            await providerRepository.IsValidParentProvider(providerUpdateDto.Id);
```
DmyMi: i agree with rabbit, in repo method expects parent id, and also the boolean result is not used anywhere. Either this check is redundant or should be used elsewhere.

OutOfSchool/OutOfSchool.BusinessLogic/Services/ProviderServices/ProviderService.cs
```diff
@@ -652,6 +686,42 @@ await UpdateWorkshopsProviderStatus(providerUpdateDto.Id, providerUpdateDto.Stat
...
+                throw new InvalidOperationException("Invalid parent provider.");
+            }
+
+            provider.Edrpou = parentProvider.Edrpou; // Automatically inherit parent's EDRPOU            
+        }
+        else // main organization
+        {
+            // BUG: concurrency issue:
+            //      while first repository with this particular user id is not saved to DB - we can create any number of repositories for this user.
+            if (providerRepository.SameExists(provider))
```
DmyMi: double check rabbit suggestion, as the original `SameExists` was made to not create another provider with same edrpou. Which means if you find the same object in repo - it would return `true` and logic might fail somewhere. I'm not sure we need to refactor the repository method like rabbit suggests, as it may break something.

Maybe add a new method for same check in this case?

OutOfSchool/OutOfSchool.DataAccess/Models/Provider.cs
```diff
@@ -142,4 +142,12 @@ currently is a placeholder property.
     public List<Contacts> Contacts { get; set; } = [];
 
     #endregion
+
+    #region Self-referencing navigation properties
+    
+    public Guid? ParentProviderId { get; set; }
+    public virtual Provider ParentProvider { get; set; }  
+    public virtual ICollection<Provider> Branches { get; set; }
```
DmyMi: i agree with empty list suggestion by rabbit like with contacts

OutOfSchool/OutOfSchool.DataAccess/Repository/ProviderRepository.cs
```diff
@@ -21,7 +21,7 @@ public ProviderRepository(OutOfSchoolDbContext dbContext)
     /// </summary>
     /// <param name="entity">Entity.</param>
     /// <returns>Bool.</returns>
-    public bool SameExists(Provider entity) => dbSet.Any(x => !x.IsDeleted && x.Edrpou == entity.Edrpou);
+    public bool SameExists(Provider entity) => dbSet.Any(x => !x.IsDeleted && x.Edrpou == entity.Edrpou && !x.IsStructuralUnit);
```
DmyMi: need to check this, according to service comment. I think another method may be needed, unless this method is usd only in like 1 place. And in general for most checks to create a new provider we should not ignore "same entity", as it would allow to execute the creation of duplicate record in database, which will be blocked by unique index hopefully, but still it is an extra exception we might need to handle :)

DmyMi (changes requested):

DmyMi (approved):

DmyMi (approved):

PR #1968, opened 2025-09-22
Implement System User for Background Jobs and Non-Web Context Operations

### LINKED ISSUE #1954: Implement System User for Background Jobs and Non-Web Context Operations
## Overview
Create a system user that exists in the database with a guaranteed fixed ID and can be used by background jobs and other processes that don't have a web context to track entity changes. This system user must be immutable and never change or be recreated with a different ID.

## Background
Currently, the application uses `ICurrentUser` interface to track who made changes to entities through the `BusinessEntityInterceptor` and `TrackableEntityInterceptor`. Background jobs (Quartz jobs) and other non-web processes don't have access to a web context, so they can't provide user information for audit trails. This results in empty `CreatedBy`, `ModifiedBy`, and `DeletedBy` fields.

### Create System User Model
- Add a new role `System` to the `Role` enum in `OutOfSchool.DataAccess/Enums/Role.cs`
- Add "system" to the default role names in `IdentityRolesInitializerHostedService.GetDefaultRoleNames()`
  - This role should not have any permissions

### Create System User Constants
- Add system user constants in `OutOfSchool.Common/Constants.cs`:
```csharp
public static class SystemUserConstants
{
    // or any other GUID that we would not create otherwise
    public const string SystemUserId = "00000000-0000-0000-0000-000000000001";
    public const string SystemUserRole = "system";
    public const string SystemUserName = "system";
    public const string SystemUserEmail = "system@outofschool.local";
}
```

### Create System User Initialization Service
- Create `ISystemUserInitializer` interface and `SystemUserInitializer` implementation
- The service should:
  - Check if system user exists in database using the fixed ID
  - Create system user **only if it doesn't exist** (never recreate or update)
  - Ensure system user has the "system" role
  - Use the predefined system user ID from constants
  - Be idempotent and safe to run multiple times
  - **Never modify existing system user** - only create if missing

### Create Context-Aware User Service
- Create `IContextAwareUserService` interface that can determine the appropriate user context
- The service should:
  - Return web user context when available (from `IHttpContextAccessor`)
  - Return system user context when no web context is available
  - Be used by entity interceptors instead of direct `ICurrentUser` injection

### Update Entity Interceptors
- Modify `BusinessEntityInterceptor` and `TrackableEntityInterceptor` to use `IContextAwareUserService`
- The interceptors should:
  - Use `IContextAwareUserService` instead of `ICurrentUser`
  - Automatically get the correct user context (web or system)
  - Handle both web context and system user context seamlessly

### Background Jobs
- Background jobs don't need to be modified - they will automatically use system user context
- The entity interceptors will handle the user context resolution
- No changes needed to job constructors or DI configuration

### Add System User Protection
- Add `IsSystemProtected` flag to User model (if not already present)
- Ensure system user is marked as system protected
- Add validation to prevent system user deletion or modification
- Update user management services to respect system user protection

omaiva:
- Added new role "System" for users
- Created SystemUserInitializer service that initializes system user
- Created ContextAwareUserService service that can determine the appropriate user context for web and non-web processes

## Summary by CodeRabbit

- New Features
  - Automatically initializes a protected "system" user and a new "system" role; adds a context-aware current user service.
- Bug Fixes
  - Prevents modification or deletion of system-protected users.
  - Returns a clearer error when updating a non-existent user.
- Refactor
  - Replaces legacy current-user dependency with a context-aware current-user across services and interceptors.
- Tests
  - Adds comprehensive tests for system user initialization, hosted service behavior, and context-aware user logic; updates related tests.


OutOfSchool/OutOfSchool.AuthorizationServer/Startup.cs
```diff
@@ -292,8 +292,10 @@ public static void AddApplicationServices(this WebApplicationBuilder builder)
         services.AddTransient<IProfileService, ProfileService>();
         services.AddScoped<IUserService, UserService>();
         services.AddSingleton<ISendGridAccessibilityService, SendGridAccessibilityService>();
+        services.AddScoped<ISystemUserInitializer, SystemUserInitializer>();
 
         services.AddHostedService<IdentityRolesInitializerHostedService>();
+        services.AddHostedService<SystemUserInitializerHostedService>();
```
DmyMi, line 304: we do have some jobs running on authorization server also, even if they do not currently have interceptor - add the `IContextAwareUserService` here also. Maybe move implementation to common project?

OutOfSchool/OutOfSchool.BusinessLogic/Services/ContextAwareUserService.cs
```diff
@@ -0,0 +1,37 @@
...
+public class ContextAwareUserService : IContextAwareUserService
+{
+    private readonly IHttpContextAccessor httpContextAccessor;
+
+    public ContextAwareUserService(IHttpContextAccessor httpContextAccessor)
+    {
+        this.httpContextAccessor = httpContextAccessor ?? throw new ArgumentNullException(nameof(httpContextAccessor));
+    }
+
+    public bool IsSystemContext => !(Principal?.Identity?.IsAuthenticated ?? false);
```
DmyMi: Technically, if a User is not authenticated for a request - it would consider the user being a `system`.
Though in general the idea of this class is correct - maybe think about additional check?

OutOfSchool/OutOfSchool.Common/Models/IContextAwareUserService.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Security.Claims;
+
+namespace OutOfSchool.Common.Models;
+public interface IContextAwareUserService
```
DmyMi: this interface can extend the `ICurrentUser` interface and override UserId or other properties if needed and also add the system protection. But delegate other methods to CurrentUser implementation.
This way we can use both interchangeably anywhere

